PR #9, opened 2015-04-12
k8petstore example unified into the main repo

jayunit100:
In case a good example is needed, this just merges my work into atomicapp-run.

This will make maintaining things natural for both the atomicapp-run creators and me as well in case we want to curate k8petstore as an example/.


jayunit100:
hi @vpavlin . shall we close this ? I assume a lot has changed since this was created.

vpavlin:
Yeah, I think we would have take a look again and rework the example to be compliant with latest stuff

PR #11, opened 2015-04-21
I cleand that up a bit

goern:


vpavlin:
This commit was part of another PR

PR #29, opened 2015-05-07
mostly working runnable dockerfile

whitel:
Will do some more testing now that the latest upstream changes are in


Dockerfile.fedora
```diff
@@ -2,22 +2,19 @@ FROM fedora
...
-ADD requirements.txt /opt/atomicapp/
-ADD setup.py /opt/atomicapp/
-ADD atomicapp/ /opt/atomicapp/atomicapp/
-
-WORKDIR /opt/atomicapp
-RUN python setup.py install
+RUN yum -y update
+RUN yum -y install python python-pip docker-io kubernetes && yum clean all
 
+COPY . /opt/atomicapp/
```
vpavlin: I don't want to use `COPY .` because my workdir often contains a lot of random files and dirs created during development and I don't want to add them in the image.

Dockerfile.fedora
```diff
@@ -2,22 +2,19 @@ FROM fedora
...
+WORKDIR /opt/atomicapp/
+RUN chmod u+x /opt/atomicapp/setup.py
```
vpavlin: This is useless layer - I don't think it makes any difference to be able to call ./setup.py instead of python setup.py

Dockerfile.fedora
```diff
@@ -2,22 +2,19 @@ FROM fedora
...
 
 WORKDIR /application-entity
-VOLUME /application-entity/answers.conf
-
-LABEL RUN docker run -it --rm --privileged --net=host -v ${PWD}:/atomicapp -v /run:/run -v /:/host -v ${PWD}/answers.conf:/application-entity/answers.conf --name NAME -e NAME=NAME -e IMAGE=IMAGE IMAGE atomicapp -v run /atomicapp
-LABEL INSTALL docker run --rm -it --privileged -v /run:/run -v ${PWD}:/atomicapp -v /:/host -v ${PWD}/answers.conf:/application-entity/answers.conf -e IMAGE=IMAGE -e NAME=NAME --name NAME IMAGE atomicapp -v install --path /atomicapp /application-entity
+VOLUME /application-entity
 
-CMD  atomicapp -h
+LABEL RUN docker run --rm -it --privileged --net=host -v ${PWD}:/atomicapp -v /run:/run -v /:/host -v ${PWD}:/application-entity --name NAME -e NAME=NAME -e IMAGE=IMAGE IMAGE atomicapp -v run /atomicapp
```
vpavlin: This is broken - with `-v ${PWD}:/application-entity` you would replace the app content with the content of `${PWD}`

Dockerfile.fedora
```diff
@@ -2,22 +2,19 @@ FROM fedora
...
+LABEL INSTALL docker run --rm -it --privileged -v /run:/run -v ${PWD}:/atomicapp -v /:/host -v ${PWD}:/application-entity -e IMAGE=IMAGE -e NAME=NAME --name NAME IMAGE atomicapp -v install --path /atomicapp /application-entity
```
vpavlin: same as line 17

build_run.sh
```diff
@@ -10,10 +11,26 @@ if [ -z "$USERNAME" ]; then
...
+popd
+
+COMMAND="\"docker run --rm -it --privileged -v /run:/run -v /:/host -v $PWD:/application-entity $USERNAME/atomicapp-run\" "
+
+echo The application is now built as a command line executable docker container. 
+echo
+echo You can run it using:
+echo $COMMAND
+echo or, you can create an alias:
+echo alias nulecule=$COMMAND
```
vpavlin: I'd suggest to use `atomicapp` instead of `nulecule` here

build_run.sh
```diff
@@ -10,10 +11,26 @@ if [ -z "$USERNAME" ]; then
...
+echo
+echo If you aren\'t sure what to do next, try \(after creating the alias\):
+echo nulecule --help
```
vpavlin: and here too `s/nulecule/atomicapp/`

build_run.sh
```diff
@@ -10,10 +11,26 @@ if [ -z "$USERNAME" ]; then
...
+echo
 
 rm -f Dockerfile
```
vpavlin: `rm -f Dockerfile` needs to happen before popd

Dockerfile.fedora
whitel: I copied that from upstream :)

build_run.sh
whitel: ok..

PR #31, opened 2015-05-08
Move Dockerfile.centos to Dockerfile

aweiteka:
Automated builds require a file named "Dockerfile". We're standardizing
on Centos. Dockerfiles for Fedora and RHEL are provided for reference
and can be used from the base directory of the project:

    [sudo] docker build -t projectatomic/atomicapp -f Dockerfile.fedora .


whitel:
Ok, the minor problem is, I think, that the name of the dockerfiles are not, entirely, reflective of what they are. The df.fedora is really the one that can be run as an "executable." I believe the rhel and centos ones are meant for the user to run interactively. Feel free to correct me. As a result, I think we need to standardize on that too (either exec or inside the container).

aweiteka:
I'm not sure what you mean. The docker hub automated build system will only work with "/base/path/to/Dockerfile/" and in the dir the file must be named `Dockerfile`. They run a command such as `docker build -t projectatomic/atomicapp .` which expects a file named `Dockerfile`.

whitel:
I just mean, in order to have a Dockerfile that is "correct" for automated builds we need to decide: 1) is the docker container meant to be used as an executable? or as an interactive shell? or both? 2) what FROM image it should use.

I believe, the problem is 1) as the reason for the different "FROM image" versions (cent, rhel, fedora) is not because of 2) but because of different beliefs on 1) and that was as convenient a way to do it as any.

vpavlin:
I'd go with centos now..We cannot use labels and stuff in Fedora Atomic anyway so I'd make Centos upstream for now (that's for 2)

For 1)..I guess it should be executable through labels and interactive through CMD..

vpavlin:
Ah..gotcha..you are talking about whether the image is a base for atomicapps or a cmd tool delivered through a docker container (like rhel-tools)

We might need/want both with different names - atomicapp vs. atomicapp-base ?

whitel:
Well, technically, i think we have three 1) executable 2) interactive shell 3) base for atomicapps. 1 & 2 can be delivered from here based on centos (ill make the mods and a pull request). I definitely think we need a separate repo for 3 or at least a separate directory. 

I propose we close this PR, let me do the mods to base the exec on centos. However, it needs to be "entrypoint" to be used as an exec. If someone wants an interactive shell, they can always --entrypoint="/bin/bash". We need to open an issue for the base-app and figure out where that lives.

vpavlin:
Can you elaborate on the difference between executable and interactive shell (and how would I use each)? I am probably still a bit confused..

whitel:
@vpavlin "docker run atomicapp install wp-app" vs "docker run atomicapp /bin/bash; atomicapp install wp-app". Like does running the docker container immediately expose the application or do you need to "enter" the container and then run atomicapp.

vpavlin:
ok, makes sense to have really just 2 images containing atomicapp - one exec and one base:)  Interactive is more for debugging and can be achieved with `--entrypoint` as you mentioned. I think `projectatomic/atomicapp` and `projectatomic/atomicapp-base` should be ok

aweiteka:
First, I'm not trying to change behavior; just trying to get automated builds working.b But we might as well have this larger discussion...

Docker doesn't require entrypoint but I agree it's a better model and I've been looking at ways to optimize ENTRYPOINT + CMD (defaults) + atomic CLI to make the user experience ideal.

Can we just merge this then tweak the Dockerfile in another PR?

vpavlin:
Agreed, let's not get blocked by details (/me thinks about making issues we had at the beginning)

Issue #27, opened 2015-05-07
Simplify CLI


aweiteka:
We're running into some issues when running as a container such as how to pass in an optional conf file and how to pass in params like interactive `--ask`.

The `install` behavior is not intuitive. It should be optional in a container context. My understanding is it pulls in remote sources (if there are any) and moves the files to a new directory. I suggest we default to current working directory `${PWD}` and leave things there, adding an `external` directory during install for remote sources. Options would be `--source <src_dir>` and `--destination <dst_dir>`. No strong opinion on option names; whatever makes sense.

In a container context bindmount options are simplified as `-v ${PWD}:/atomicapp`. The `answerfile.conf` will appear there (if present). If `/atomicapp` doesn't have a `Nulecule` file then copy contents of `cp /application-entity/* /atomicapp/ -r`. We won't want to do this in a non-container context.

vpavlin:
Current behvaiour

```
[vpavlin@dhcp-30-155 ~/devel/upstream/atomicapp-run/test-app(master)]
$ ls
[vpavlin@dhcp-30-155 ~/devel/upstream/atomicapp-run/test-app(master)]
$ atomicapp install vpavlin/helloapache
2015-05-08 12:21:39,267 - atomicapp.install - INFO - App name is vpavlin/helloapache, will be populated to None
2015-05-08 12:21:39,267 - atomicapp.params - WARNING - No answers file found.
Trying to pull repository docker.io/vpavlin/helloapache ...
0f330701741d: Download complete 
6941bfcbbfca: Download complete 
41459f052977: Download complete 
fd44297e2ddb: Download complete 
f286a6c39e45: Download complete 
73846768f6ac: Download complete 
d7b655f6e4dc: Download complete 
b2d4efd88915: Download complete 
5bce9d791f66: Download complete 
d0cfb7fc6a83: Download complete 
922a34ba98de: Download complete 
de84a0b71273: Download complete 
3e7e3dcc0576: Download complete 
d27de2504144: Download complete 
064327e7515d: Download complete 
209139e515c1: Download complete 
1971e8e9c2d9: Download complete 
6fa124794b79: Download complete 
d73fcf4d4871: Download complete 
Status: Downloaded newer image for docker.io/vpavlin/helloapache:latest
973c06b1321b2c69d05ee99237cf38a6600cea48cd62f6dfc388529f115eaa76
[...]

PR #35, opened 2015-05-09
Install on run

vpavlin:
Try to install the app if the run is called with non-existing path (i.e. with the image name)

This is the sequence of commands you had to do to run the app
```
cd /some/path
atomicapp install vpavlin/myapp
atomicapp run .
```

with this PR you can skip the install step

```
cd /some/path
atomicapp run vpavlin/myapp
```


vpavlin:
This is part of #27

goern:
your PR seem to be mixed up, what about using feature branches? ;)

aweiteka:
Other than the messy commit history, LGTM. Great turnaround on these features!

PR #39, opened 2015-05-11
Readme cleanups, fix for #37

### LINKED ISSUE #37: Quick question about Python version and OrderedDict
What version of Python does atomicapp work with? I just notice that `OrderedDict` is used here https://github.com/projectatomic/atomicapp/blob/master/atomicapp/providers/kubernetes.py#L33 . `OrderedDict` was introduced since Python 2.7.

goern:
renamed the README and converted it to Asciidoc, added a comment on the dependencies, this closes #37

### DIFF
--- a/README.asciidoc
+++ b/README.asciidoc
@@ -20,7 +20,7 @@ If you want to do some changes to the code, I suggest to do:
 
 ```
 cd atomicapp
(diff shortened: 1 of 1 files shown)


vpavlin:
LGTM

PR #44, opened 2015-05-11
fix dependencies in centos and fedora images

tkdchen:
Without these dependencies, PyYAML cannot be installed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ MAINTAINER Vaclav Pavlin <vpavlin@redhat.com>
 
 RUN echo -e "[epel]\nname=epel\nenabled=1\nbaseurl=https://dl.fedoraproject.org/pub/epel/7/x86_64/\ngpgcheck=0" > /etc/yum.repos.d/epel.repo
 
-RUN yum install -y --setopt=tsflags=nodocs python-pip docker && \
+RUN yum install -y --setopt=tsflags=nodocs python-pip python-devel docker gcc libyaml-devel && \
     yum clean all
 
 ADD atomicapp/ /opt/atomicapp/atomicapp/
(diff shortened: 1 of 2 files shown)


aweiteka:
LGTM

vpavlin:
@tkdchen Images work perfectly for me - what is your problem with them? Can you provide some error messages or something?

aweiteka:
@tkdchen we're not seeing the errors on the automated build.

[Dockerfile](https://registry.hub.docker.com/u/projectatomic/atomicapp/dockerfile/)
[Build log](https://registry.hub.docker.com/u/projectatomic/atomicapp/builds_history/198334/)

tkdchen:
@aweiteka partial logs here. Error happens while installing PyYAML. This is generated from the step of

    RUN python setup.py install

based on commit 2b27f6c96f9b5f

```
Searching for PyYAML
Reading https://pypi.python.org/simple/PyYAML/
Best match: PyYAML 3.11
Downloading https://pypi.python.org/packages/source/P/PyYAML/PyYAML-3.11.zip#md5=89cbc92cda979042533b640b76e6e055
Processing PyYAML-3.11.zip
Writing /tmp/easy_install-6qwemq/PyYAML-3.11/setup.cfg
Running PyYAML-3.11/setup.py -q bdist_egg --dist-dir /tmp/easy_install-6qwemq/PyYAML-3.11/egg-dist-tmp-bVicMp
unable to execute 'gcc': No such file or directory

libyaml is not found or a compiler error: forcing --without-libyaml
(if libyaml is installed correctly, you may need to
 specify the option --include-dirs or uncomment and
 modify the parameter include_dirs in setup.cfg)
zip_safe flag not set; analyzing archive contents...
Adding PyYAML 3.11 to easy-install.pth file
```

vpavlin:
Well, PyYAML falls back to python parser instead of libyaml if it can't be installed properly. For me it is probably more important not to install unnecessary packages rather then using C implementation over Python implementation...Any strong opinions anybody?

aweiteka:
I think unless we really see a need for bringing in additional dependencies we should leave them out. If you disagree for some reason please re-open and provide additional detail.

PR #47, opened 2015-05-11
added a check if we build for the default target... centos

goern:
this fixes what we have broken by moving from Dockerfile.centos to Dockerfile, that was due to some CI requirements...


goern:
No, we should stop hacking on that and hand the task over to hub.docker for building...

vpavlin:
I don't really care..I see this only as a development script for rebuilding an image instead of writing `docke build .....` - so IMO let's either drop it or make it work like `./build.sh ... done`

PR #50, opened 2015-05-11
Use 'pwd' backticks syntax as opposed to '\${PWD}' for current workin…

aweiteka:
…g directory


goern:
do they behave differently? what was the reason for the change?

aweiteka:
FYI, bug filed: https://bugzilla.redhat.com/show_bug.cgi?id=1219124

PR #53, opened 2015-05-12
add break after title so first line will render

jasonbrooks:


aweiteka:
Good catch! LGTM

vpavlin:
LGTM

PR #58, opened 2015-05-13
Feature/inherit

vpavlin:
We have `inherit` in spec version 0.0.1-alpha but it was not implemented until now

I also updated examples.


aweiteka:
LGTM

PR #65, opened 2015-05-14
proper dynamic import of plugins

TomasTomecek:
This should fix warnings like:

```
RuntimeWarning: Parent module 'atomicapp.providers' not found while handling absolute import...
```

The problem is the module name passed to `load_source`. We don't need to
put absolute path there, because we are loading the plugin directly from
sources. And that's where the warning comes from: it can't find
'atomicapp.providers' on sys.path so hence the warning.


vpavlin:
LGTM

aweiteka:
LGTM

PR #74, opened 2015-05-15
avoid duplicate isTrue

tkdchen:
`isTrue` should be a general function that is applicable globally. Move it out of class `Params` and put into `utils.py`.


vpavlin:
LGTM

aweiteka:
LGTM

PR #82, opened 2015-05-15
Add action 'stop' to atomicapp

vpavlin:
This adds an 'undeploy' method to Provider interface and an
implementation to kubes provider which removes all services/rc/pods
deployed by 'atomicapp run'

It's currently blocked by https://github.com/projectatomic/atomic/pull/54 to be used as

```
atomic stop my/awesome-atomicapp
```

PR #93, opened 2015-05-19
Add action 'stop' to atomicapp

goern:
this superseeds #82


goern:
@vpavlin can you have a look it this covers your original PR.

vpavlin:
Yeah, it seems to be ok

PR #97, opened 2015-05-20
Cleanup based on landscape.io warnings

vpavlin:


goern:
sorry if landscape is to noisy ;)

Issue #89, opened 2015-05-19
Implement JSON r/o Web API for atomicapp


goern:
As a computer program,
I want to query the current state of atomicapp run via HTTP/JSON
so that I can reuse that information.

While atomicapp is running, a web api shall be exposed via port 80/tcp, there shall be no authentication or encryption.
This api shall provide
 1. the nulecule file itself
 2. a global state information

after the execution of atomicapp has finished, the web api shall accept one final request and than terminat atomicapp (so that the container will stop).

The global state information provided by the web api will be feed from a global structure inside atomicapp, this structure will be updated by any method that would like to 'publish' a state information (like 'we are pulling docker images', or 'installation failed').

vpavlin:
Let's not use port 80, so that we don't have to run with privileges:-) 8080 should be fine for now, we'll have to figure out some "standard port" later

Global state information should be a list IMO - it could also easily serve as a log for the computer program as well to review what happend

vpavlin:
Duplicate of https://github.com/projectatomic/atomicapp/issues/30 ?

PR #99, opened 2015-05-20
Implement JSON r/o Web API for atomicapp

sub-mod:
PoC for https://github.com/projectatomic/atomicapp/issues/89
@vpavlin @goern 

daemon option:
atomicapp -d install aweiteka/guestbook-app

rest endpoints
curl -X GET http://localhost:5000/atomicapp-run/api/v1.0/status
curl -X POST http://localhost:5000/atomicapp-run/api/v1.0/quit

### DIFF
--- a/atomicapp/cli/main.py
+++ b/atomicapp/cli/main.py
@@ -6,6 +6,9 @@ import os
 from argparse import ArgumentParser
 from argparse import RawDescriptionHelpFormatter
 import logging
+from atomicapp.rest import Rest
+from atomicapp import context
+import threading
 
 from atomicapp import set_logging
 from atomicapp.constants import ANSWERS_FILE, __ATOMICAPPVERSION__, __NULECULESPECVERSION__
@@ -15,14 +18,22 @@ logger = logging.getLogger(__name__)
 def cli_install(args):
     install = Install(**vars(args))
     install.install()
+    context.globalCtx.addStatus("Install Successful." ,status_type="COMPLETED")
 
 def cli_run(args):
     ae = Run(**vars(args))
     ae.run()
+    context.globalCtx.addStatus("Run Successful." ,status_type="COMPLETED")
 
 def cli_stop(args):
     stop = Run(stop = True, **vars(args))
     stop.run()
+
+def format_answers_path(cargs):
+    if cargs.APP and os.path.exists(cargs.APP):
+        cargs.answers = os.path.join(cargs.APP, ANSWERS_FILE)
+
+    return cargs
 
 # see https://docs.python.org/release/2.2.3/whatsnew/sect-rellinks.html
 class CLI(object):
@@ -34,6 +45,7 @@ class CLI(object):
         self.parser.add_argument("-V", "--version", action='version', version='atomicapp %s, Nulecule Specification %s' % (__ATOMICAPPVERSION__, __NULECULESPECVERSION__), help="show the version and exit.") # TODO refactor program name and version to some globals
         self.parser.add_argument("-v", "--verbose", dest="verbose", default=False, action="store_true", help="Verbose output mode.")
         self.parser.add_argument("-q", "--quiet", dest="quiet", default=False, action="store_true", help="Quiet output mode.")
+        self.parser.add_argument("-d", "--daemon", dest="daemon", default=False, action="store_true",  help="Run the CLI as a REST server daemon.")
 
         self.parser.add_argument("--dry-run", dest="dryrun", default=False, action="store_true", help="Don't actually call provider. The commands that should be run will be sent to stdout but not run.")
         self.parser.add_argument("-a", "--answers", dest="answers", default=os.path.join(os.getcwd(), ANSWERS_FILE), help="Path to %s" % ANSWERS_FILE)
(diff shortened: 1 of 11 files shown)


vpavlin:
Can you please rebase on top of current master? There has been some cleanup done and this commit brings a lot of the mess back:)

vpavlin:
Add something like 

```
atomicapp.config.globalCtx.addStatus(repr(ex))
```

At the line `109` (under `logger.error`) - to log error on failure

vpavlin:
You have tabs and spaces mixed, please use spaces  (4 spaces = 1 tab) so that the code in project is consistent

atomicapp/utils.py
```diff
@@ -20,6 +20,20 @@
...
     return str(val).lower() in true_values
 
+def addShelveKV(dbname, key, val):
+    with closing(shelve.open(dbname, writeback = True)) as s:
+        s[key] = val
+
+def appendShelveKV(dbname, key, val):
+    with closing(shelve.open(dbname, writeback = True)) as s:
+        temp = s[key]
+        temp.append(val)
```
vpavlin: No need to use `temp`, simply call

```
s[key].append(val)
```

atomicapp/cli/main.py
```diff
@@ -106,8 +109,20 @@ def run(self):
...
+    start_restapp(debug=False, port=5000)
+
+
+def start_cli():
     cli = CLI()
     cli.run()
 
+def start_restapp(debug, port):
+    rest = atomicapp.rest.Rest()
+    rest.addResource(atomicapp.rest.StatusList, '/atomicapp-run/api/v1.0/status')
```
vpavlin: This should be probably part `rest.py`

atomicapp/utils.py
sub-mod: in shelve you have to do this. you need to fetch the entire value and  append and save the entire object back again.

vpavlin: interesting..I tried that and it worked..

sub-mod: ah ok i have  writeback = True ,so you dont need it. writeback = True ; will cache the data and rewrite it.

vpavlin:
Can you please just commit changes rather than doing a rebase everytime? This is really hard to review - I can't see the differences between your commit if you do rebase..

sub-mod:
ok , sure.

goern:
@sub-mod could you add `-p 5000:5000` to `atomicapp.sh` so that it supports your REST server. And change the API version to `__ATOMICAPPVERSION__` it is defined in https://github.com/projectatomic/atomicapp/blob/master/atomicapp/constants.py

goern:
please also put at least `id` from the nulecule file in each response, so that clients know which atomicapp instance is responding, maybe the `metadata` object too

goern:
beside of that comments it looks good! please what out for landscape comments and keep up the good work! thx.

vpavlin:
Please file namespace commit as separate PR

sub-mod:
@vpavlin  https://github.com/projectatomic/atomicapp/pull/123

vpavlin:
This patch: https://vpavlin.fedorapeople.org/0001-Fix-rest-API-feature.patch should fix all issues and make this PR acceptable, please apply it  and push to you branch

```
curl -O https://vpavlin.fedorapeople.org/0001-Fix-rest-API-feature.patch
git am 0001-Fix-rest-API-feature.patch
```

sub-mod:
@goern kindly review

vpavlin:
LGTM 

Note: Neither I, nor @sub-mod think this is a final perfect solution, but it's good enough to have a working example of integration with Cockpit

sub-mod:
+1 @vpavlin 's comment
the way we do addStatus is a duplication of logging efforts. IMHO i think we should a refactor of the code flow , add the some method descriptors to avoid duplication.But It's ok now, I think. We need to  figure the way out later.

goern:
LOKTM, please test accessing the REST API while `atomic install projectatomic/wordpress-centos7-atomicapp` (or zookeeper or whatever atomicapp) is running, please update the atomicapp base image. and I vote for keeping this in :dev up until 0.5.0 aka https://github.com/projectatomic/atomicapp/milestones/Red%20Hat%20Summit%202015

sub-mod:
@goern is port 5000 value acceptable ?

stefwalter:
@goern @sub-mod I'm not sure I understand how this rest API is meant to be located. Will it always listen on port 5000 and conflict with any service listening on that port on the host or other container on that port? Is it by design that only one such atomicapp installer can be active at a time?

vpavlin:
@stefwalter Good question. I don't think we should force a single instance of atomicapp at a time.

@sub-mod Is it possible to remove port binding to the host port and use <container_ip>:<port> instead of localhost:<port> for you in your cockpit work?

goern:
@vpavlin @sub-mod as we use `--net=host` for atomicapp ( #165 ) container, I think the restserver cant listen on any container_ip

sub-mod:
since atomicapp will be invoked within the atomicapp container via atomic command there will be no port conflict because each instance of the atomic can have a different exposed port number.

stefwalter:
> since atomicapp will be invoked within the atomicapp container via atomic command there will be no port conflict because each instance of the atomic can have a different exposed port number.

How would you know which container corresponds to the command you just launched?

>  curl -X POST http://localhost:5000/atomicapp-run/api/v1.0/quit

This is unauthenticated. Anyone on the machine can quit any active atomicapp installer. As I've mentioned before, I think the REST API is needlessly complex way to get status information. It's possible to make it work, but it's going to be a lot of pain finishing it up.

PR #100, opened 2015-05-20
show logs when in dry-run way

### LINKED ISSUE #84: Inconsistent dry-run behavior
Currently, there are several different behaviors of dry-run. When dry-run is enabled, some of them print a message to stdout by `print` function, some of them output a message by logger, and some of them do not output anything.

When dry-run is enabled, each time encounter dry-run option, a message should be outputted.

IMO, `logger.info` is a good way to output such a dry-run message, which is prefixed with "DRY-RUN: " that is useful to separate dry-run message from normal log messages.

tkdchen:
This is to fix #84

### DIFF
--- a/atomicapp/create.py
+++ b/atomicapp/create.py
@@ -57,7 +57,7 @@ class Create(object):
 
         cmd = ["docker", "build", "-t", tag, "."]
         if self.dryrun:
-            logger.info("DRY-RUN: %s" % " ".join(cmd))
+            logger.info("DRY-RUN: %s", " ".join(cmd))
         else:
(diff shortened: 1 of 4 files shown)


vpavlin:
LGTM

vpavlin:
I haven't noticed that at first - can you please update your patch so that landscape.io is happy about it?:) (i.e. replace `%` with `,` in logger calls)

tkdchen:
Yes. I'll fix those 4 problems in another commit.

vpavlin:
Cool, thanks

LGTM

PR #111, opened 2015-05-21
Move SCHEMA_URL into constants module

### LINKED ISSUE #91: hardcoded SCHEMA_URL
In `create.py`, `SCHEMA_URL` is hardcoded to https://raw.githubusercontent.com/projectatomic/nulecule/master/spec/0.0.1-alpha/schema.json . When user does not provide a schema via `--schema` from command line, a `schema.json` file will be downloaded from that URL. Is this an expected behavior?

There is already a new version 0.0.2 of `schema.json` in `spec/`, should this URL be changed to that new one?

vpavlin:
Fixes #91


aweiteka:
@vpavlin need to remove "fix rebase openshift provider". Otherwise LGTM

PR #113, opened 2015-05-21
Use pip install instead of direct call to setup.py

### LINKED ISSUE #40: pip should be used over plain distutils, even for local installations
"python setup.py install" doesn't create any installation metadata, so it should almost never be used.

"pip install ." is the equivalent that creates the appropriate installation metadata that allows uninstallation later.

vpavlin:
Fixes #40

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ ADD requirements.txt /opt/atomicapp/
 
 WORKDIR /opt/atomicapp
 
-RUN python setup.py install
+RUN pip install .
 
 WORKDIR /atomicapp
 VOLUME /atomicapp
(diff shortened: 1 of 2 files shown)


aweiteka:
@vpavlin should we use pip in the Dockerfile as well?

atomicapp/providers/openshift.py
```diff
@@ -1,13 +1,9 @@
 from atomicapp.plugin import Provider, ProviderFailedException
 
-<<<<<<< HEAD
```
goern: What is this? Fixing a merge conflict?

vpavlin: Fixed..

PR #115, opened 2015-05-21
Add namespaced labels to Dockerfile

### LINKED ISSUE #33: Proposal: usage of LABELs in Application's Dockerfile
To give a little more meaning to the dockerfile and make the meta data available to atomic tool (and other tools like devassistant) I would like to propose the following LABELs

`LABEL io.projectatomic.nulecule.specversion 0.0.2` - this indicates, that the Dockerfile belongs to a Atomicapp, compatible with Nulecule Spec 0.0.2, inferred from that, we know that this docker container is an atomicapp.

vpavlin:
Fixes #33

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,5 +22,8 @@ LABEL RUN docker run -it --rm --privileged --net=host -v `pwd`:/atomicapp -v /ru
 LABEL STOP docker run -it --rm --privileged --net=host -v `pwd`:/atomicapp -v /run:/run -v /:/host --name NAME -e NAME=NAME -e IMAGE=IMAGE IMAGE -v stop /atomicapp
 LABEL INSTALL docker run --rm -it --privileged -v /run:/run -v `pwd`:/atomicapp -e IMAGE=IMAGE -e NAME=NAME --name NAME IMAGE -v install --destination /atomicapp /application-entity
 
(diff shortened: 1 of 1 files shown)


aweiteka:
We [lost the reference](https://github.com/projectatomic/ContainerApplicationGenericLabels/issues/7) to these namespaced labels. Some more discussion is warranted, I think. I recommend holding off for a bit so we don't go down a path that might change.

goern:
So what I read from the Dockerfile after the merge: This atomicapp will support provider kube, dockah, and openshift. As this is the base image, it has no artifacts at all, and layered images need to overwrite the Label anyways... so putting the list of supported providers in the base image may lead to errors: the layered image may not have overwritten the label, someone asks the layered image about the supported providers and a list that has too many providers will be returned.

vpavlin:
Hmm..I understood this as a list of providers supported by atomicapp (i.e. which providers are implemented).

goern:
So Atomicapp Base Image may support all these, but if the ISV's app does not provide deployment configuration files for it?!

vpavlin:
Yes..I don't know..the difference is in the point of view (and incorrect naming from my side). This definitely should not go in..closing

My point was more to provide a list of implemented providers..but that should be rather an argument in atomicapp cli

Issue #75, opened 2015-05-15
Support provider context


aweiteka:
Application context and namespacing is a key feature of providers such as kubernetes. Here's how I think it should work.

### Definition

Default namespace is defined in the nulecule file in the metadata section.

```
metadata:
  kubernetes:
    context: mycluster
    namespace: mycoolapp
...
```

At deployment the default context and namespace maybe overridden using a general parameter.

```
[general]
provider=kubernetes
context=aweiteka-cluster
namespace=mycoolerapp
...
```

### Provider implementation

If provided the provider plugin should create resources with the context and/or namespace options.

```
kubectl create -f mycoolapp-pod.json --context=aweiteka-cluster --namespace=mycoolapp
```

The context and namespace must exist for the above create command to succeed. It appears to be safe to do this even if values exist.
```
kubectl config set-context aweiteka-cluster --namespace=mycoolapp
```

#### Configuration file

It appears `~/.kube/.kubeconfig` file does not exist by default and causes `set-context` calls to fail. Merely touching this file resolves the failure.

```
$ kubectl config set-context myapp --namespace=mycoolapp
open /home/vagrant/.kube/.kubeconfig: no such file or directory
$ sudo kubectl config set-context myapp --namespace=mycoolapp
open /root/.kube/.kubeconfig: no such file or directory
$ mkdir /home/vagrant/.kube/
$ touch /home/vagrant/.kube/.kubeconfig
```

#### Viewing and set the context namespace 
```
$ kubectl config view
apiVersion: v1
clusters: []
contexts: []
current-context: ""
kind: Config
preferences: {}
users: []
$ kubectl config set-context myapp --namespace=mycoolapp
$ kubectl config view
apiVersion: v1
clusters: []
contexts:
- context:
    cluster: ""
    namespace: mycoolapp
    user: ""
  name: myapp
current-context: ""
kind: Config
preferences: {}
users: []
```

aweiteka:
@goern is this a dupe of #60  ?

goern:
@aweiteka didnt we do that via namespaces?

aweiteka:
@vpavlin please reference your provider config work.

vpavlin:
Closed by #267

PR #123, opened 2015-05-25
support namespace for kubernetes provider

sub-mod:
support namespace for kubernetes provider 
so that when create artifacts via kubectl command --namespace argument can be used to create artifacts within a namespace.

notes:
config is constructed as content of answers[general] + params defaults + answers[component]/interactive answers
provider_class(self.params.getValues(component), dst_dir, self.dryrun) -> first argument contains all params from the "component" and "general" merged..so it will already contain "namespace"..you can access it as "self.config["namespace"]

### DIFF
--- a/atomicapp/constants.py
+++ b/atomicapp/constants.py
@@ -14,7 +14,8 @@ ANSWERS_FILE_SAMPLE="answers.conf.sample"
 WORKDIR=".workdir"
 
 DEFAULT_PROVIDER="kubernetes"
-DEFAULT_ANSWERS={"general":{"provider":DEFAULT_PROVIDER}}
+DEFAULT_NAMESPACE="default"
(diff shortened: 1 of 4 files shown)


vpavlin:
Please change your code as suggested by following patch: http://pastebin.centos.org/27866/
Also add some `logger.info("Using namespace %s", self.namespace)` to the `init()` function.

Also note that PR #124 has one conflicting change so you will need to fix your patch, sorry

Otherwise looks good, thanks

sub-mod:
@goern kindly review

vpavlin:
LGTM

aweiteka:
Per https://github.com/projectatomic/atomicapp/issues/75 this relies on the namespace already being configured. As a first pass at support I think this is fine. Full support for kube provider context may take a bit more design consideration.

LGTM

PR #128, opened 2015-05-26
Feature/comply to nulecule 0.0.2

vpavlin:
This PR contains 3 changes:
* atomicapp should no comply to spec 0.0.2 (please verify by running it against: https://github.com/projectatomic/nulecule/pull/82)
* Refactorization of params and utils - describer below
* Bump of spec and app version

I came to the point where I either had to copy code between utils and
params or have cross dependency between params and utils.

I decided to move all nulecule-related methods (things like
getComponent, get*Whatever*FromMainfile..) to params and rename params
to nulecule_base.

This makes dependencies in the project cleaner - nulecule_base will cary
all nulecule specific methods, install/run will carry the logic and
utils will provide (mostly) static helper methods.

This commit also removes generator/ dir as there was nothing useful atm
and create.py which was broken and unusable anyway. It will come back
once we figure out what it should do and how it should work.


atomicapp/nulecule_base.py
```diff
@@ -265,3 +238,114 @@ def _update(self, old_dict, new_dict):
...
+
+    def checkSpecVersion(self):
+        if not self.mainfile_data:
+            raise ValueError("Could not access %s data" % MAIN_FILE)
+
+        if "specversion" not in self.mainfile_data:
+            raise ValueError("Data corrupted: couldn't find specversion in %s" % MAIN_FILE)
+
+        if self.mainfile_data["specversion"] == __NULECULESPECVERSION__:
+            logger.info("Version check successful: specversion == %s", __NULECULESPECVERSION__)
```
goern: I wouldnt log a successful check... no news is good news

atomicapp/nulecule_base.py
```diff
@@ -265,3 +238,114 @@ def _update(self, old_dict, new_dict):
...
+            raise Exception("Couldn't pull %s" % image)
+
+
+    def fromListToDict(self, llist):
+        result = {}
+        for item in llist:
+            if "name" in item:
+                result[item.get("name")] = item
+            else:
+                logger.warning("Atribute 'name' missing in %s", item)
```
goern: there is a T missing in A*t*ribute

atomicapp/constants.py
```diff
@@ -1,5 +1,5 @@
-__ATOMICAPPVERSION__ = '0.0.1-alpha'
-__NULECULESPECVERSION__ = '0.0.1-alpha'
+__ATOMICAPPVERSION__ = '0.1.0'
+__NULECULESPECVERSION__ = '0.0.2'
```
aweiteka: You'll need to update SCHEMA_URL also

vpavlin: Hmm..the question is if the SCHEMA_URL needs to be there as it's not used anymore (was in create.py)

goern: lets remove all that is not used...

PR #138, opened 2015-06-01
create namespace in kubernetes provider deploy call

sub-mod:
related to https://github.com/projectatomic/atomicapp/pull/123
@vpavlin @goern

### DIFF
--- a/atomicapp/providers/kubernetes.py
+++ b/atomicapp/providers/kubernetes.py
@@ -56,11 +56,14 @@ class KubernetesProvider(Provider):
 
     def tmpfile(self):
         temp = tempfile.NamedTemporaryFile(suffix='.json',prefix='k8-namespace_',dir='/tmp')
-            logger.info("Using temporary file %s", temp)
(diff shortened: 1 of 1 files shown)


vpavlin:
Why do we need this? Namespace gets created automatically if it does not exist, why do we need a code to create it explicitly?

aweiteka:
I understood kube namespacing to be part of context. See findings in #75. So the end-goal was to enable the end-user to pass in context and optionally namespace during deployment via answerfile. Regardless, I wouldn't want to hard-code the API version.

vpavlin:
What I got from my experiments, it's ok to just add `--namespace=foo` and the namespace is created and pods/services are assigned to it...  That's why I don't understand the purpose of this patch.. Being able to define context and namespace in Nulecule with capability to override it in answers.conf is a different story I think.

sub-mod:
@vpavlin sorry for the delayed response, Yes this PR is not needed.
I was not aware that --namespace created a namespace since I was using an older version of kubectl.

Issue #136, opened 2015-05-28
Running second app in home directory attempts to run wrong app


nzwulfin:
Platform: Fedora 22 Atomic 22.17

Testing `atomic run projectatomic/helloapache` from Docker Hub, I had issues running a second app after a successful run.  I removed the helloapache pod using kubectl. 

A subsequent run of `atomic run projectatomic/guestbookgo-app` scheduled the 'helloapache' app instead of the 'guestbookgo-app'.

I was running in the 'fedora' home dir, and the artifacts from the 'helloapache' app aren't getting overwritten.  

Going over the `atomic run` output with @vpavlin in #atomic, the relevant output is:

`2015-05-28 17:25:36,331 - atomicapp.install - INFO - App name is projectatomic/guestbookgo-app, will be populated to /atomicapp
2015-05-28 17:25:36,332 - atomicapp.install - INFO - Component data exist in /atomicapp, skipping population...`

fpaste of complete `atomic run` output here http://ur1.ca/mnvxl

Workaround (from @jasonbrooks ): create a new working directory for each new app, and run the `atomic run` command from that new directory.

dustymabe:
@nzwulfin Can you verify that #207 fixed this issue for you and close out this issue if so?

dustymabe:
@nzwulfin nevermind. Looks like your issue really had nothing to do with concurrently running atomic but rather running two different atomicapps back to back without switching your working dir. I think #142 is the real answer here.

goern:
@vpavlin ping?!

dustymabe:
I think this will be taken care of by PR #278. It should be merge shortly and we can confirm before beta-3.

PR #140, opened 2015-06-03
Fail when user tries to rewrite existing app

vpavlin:
@nzwulfin Does this fix your issue #136 ? 

```
  File "/usr/bin/atomicapp", line 9, in <module>
    load_entry_point('atomicapp==0.1', 'console_scripts', 'atomicapp')()
  File "/usr/lib/python2.7/site-packages/atomicapp/cli/main.py", line 89, in main
    cli.run()
  File "/usr/lib/python2.7/site-packages/atomicapp/cli/main.py", line 70, in run
    args.func(args)
  File "/usr/lib/python2.7/site-packages/atomicapp/cli/main.py", line 17, in cli_install
    install.install()
  File "/usr/lib/python2.7/site-packages/atomicapp/install.py", line 104, in install
    raise Exception("You are trying to overwrite existing app %s with app %s - clear or change current directory." % (current_app_id, self.nulecule_base.app_id))
Exception: You are trying to overwrite existing app skydns-atomicapp with app helloapache-app - clear or change current directory.
```

### DIFF
--- a/atomicapp/install.py
+++ b/atomicapp/install.py
@@ -81,29 +81,40 @@ class Install(object):
             dst = self.nulecule_base.target_path
         distutils.dir_util.copy_tree(src, dst, update=(not self.nulecule_base.update))
 
+    def _fromImage(self):
+        return not self.nulecule_base.app_path or self.nulecule_base.target_path == self.nulecule_base.app_path
+
     def install(self):
         self.nulecule_base.loadAnswers(self.answers_file)
 
-        if self.nulecule_base.app_path and not self.nulecule_base.target_path == self.nulecule_base.app_path:
-            logger.info("Copying content of directory %s to %s", self.nulecule_base.app_path, self.nulecule_base.target_path)
-            self._populateApp(src=self.nulecule_base.app_path)
-
-        mainfile_path = os.path.join(self.nulecule_base.target_path, MAIN_FILE)
-
-        if not self.nulecule_base.app_path and (self.nulecule_base.update or not os.path.exists(mainfile_path)):
+        mainfile_dir = self.nulecule_base.app_path
+        if self._fromImage():
             self.nulecule_base.pullApp()
             self._copyFromContainer(self.nulecule_base.app)
-            mainfile_path = os.path.join(self.utils.getTmpAppDir(), MAIN_FILE)
-            logger.debug("%s path for pulled image: %s", MAIN_FILE, mainfile_path)
-            self.nulecule_base.loadMainfile(mainfile_path)
-            logger.debug("App ID: %s", self.nulecule_base.app_id)
+            mainfile_dir = self.utils.getTmpAppDir()
 
-            self._populateApp()
-        else:
-            logger.info("Component data exist in %s, skipping population...", self.nulecule_base.target_path)
+        current_app_id = None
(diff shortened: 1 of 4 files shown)


nzwulfin:
OK managed to test and deleted the old comments, I was doing silly things locally.

@vpavlin It depends on what you mean by fix?  It raises a clear error now when not using a working directory.  So that's good.

But, would it be better to create a working directory based on app_id if it doesn't already exist?  This would make the `atomic run <nulecule/container>` more self-contained an experience.  

I looked through the code and couldn't really tell how many different path checks would be impacted.  I'll leave it to you for estimating if it's worth the effort.

vpavlin:
Currently when using `atomicapp` cli (not `atomic`) you can do

```
atomicapp install <image>
```

and then

```
atomicapp run <image>
```

or 

```
atomicapp run .
```

If I do what you are proposing it would be

```
atomicapp run <app-name>/
```

correct? I like the current behaviour better cause it seems a bit more predictable to me - name of the image does not directly imply the name of the app thus you cannot be sure what subdir was created.

Hmm..no thinking about it - what do you mean by "working directory" in your comment?

nzwulfin:
@vpavlin Ah yes, good question, so I'm running these containers with `atomic run <app-name>`, which should push the app container into an atomicapp container for scheduling (if I'm paraphrasing your IRC explanation correctly) from an Atomic host.

Re: "working directory" To stop the behavior reported, I make a new directory for each app and run `atomic run <app-name>` from in that directory.  The artifacts created in whatever directory the user happens to be in at the time.
```
[fedora@atomic-master ~]$ ll
total 20
-rw-r--r--. 1 root   root    104 Jun  3 14:17 answers.conf.sample
drwxr-xr-x. 3 root   root     16 Jun  3 14:17 artifacts
drwxrwxr-x. 3 fedora fedora   24 Jun  3 13:20 atomicapp
-rw-rw-r--. 1 root   root    177 Jun  3 13:08 Dockerfile
-rw-rw-r--. 1 fedora fedora   92 Jun  3 11:16 flanneld-conf.json
drwxrwxr-x. 3 fedora fedora   33 Jun  3 14:19 maria
-rw-rw-r--. 1 root   root    502 Jun  3 09:36 Nulecule
-rw-rw-r--. 1 root   root   1809 Jun  3 09:36 README.md
drwxrwxr-x. 4 fedora fedora   78 Jun  3 14:59 skydns-app
```
I've not tried to directly use `atomicapp run <app-name>`

How does that change the expectation of behavior and should this be raised as bad `atomic` behavior?

nzwulfin:
Ok epiphany whike driving.

I think this is a side effect from the dockerfile of the nulecule examples.

IIRC they may be mapping pwd to the containers, causing the inital behavior
thats now corrected or at least caught.

We may need to look at how nulecule dockerfilea are built.

vpavlin:
Yes, correct they do that - and it's because we cannot pass any arguments or env vars through `atomic` - thus the only way to have some predictive behaviour is to volume mount `pwd`. I don't see any other reliable solution to this right now.

vpavlin:
The commits I added are due to the CI tests @imcleod added - as this PR changes the order of `install` steps to be able to check the app ids, we need to skip this check (and app population) when `--dry-run` is given

PR #141, opened 2015-06-03
Initial drop of tests for Travis CI

imcleod:
These two tests take advantage of the --dry-run option as well as the caching of
previously extracted Docker image content.  If any element of this caching
behavior is changed, these test will likely need to be recreated.

The two tests Nulecules are borrowed from the nulecule spec examples here:

https://github.com/projectatomic/nulecule/tree/master/examples

I created this test content by doing a dry run on a system actually running
Docker, then extracting the "external" directory.

### DIFF
--- a/atomicapp/cli/main.py
+++ b/atomicapp/cli/main.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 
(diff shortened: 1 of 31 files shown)


vpavlin:
LGTM

Thank you!

PR #144, opened 2015-06-03
remove artifact of original debug testing

imcleod:
Apologies.  This is a capture of the output from my online run of this test originally.  It should not be in the repo.  Removing.


aweiteka:
Yeah, I saw this too. Thanks for the cleanup. LGTM

PR #146, opened 2015-06-10
Answer kubectl

markllama:
This change has two effects:

1) ````python setup.py bdist_rpm```` works, to create an RPM of atomicapp
2) (and more significantly) allow the caller to provide [general] provider_cli to override the default CLI path for a provider (kubernetes in this case)

This should allow for running an atomicapp on a Kubernetes Dev/Test environment in which the ````kubectl```` command is inside ````kubernetes/cluster/kubectl.sh``

### DIFF
--- a/atomicapp/providers/kubernetes.py
+++ b/atomicapp/providers/kubernetes.py
@@ -21,17 +21,45 @@ class KubernetesProvider(Provider):
 
         logger.info("Using namespace %s", self.namespace)
         if self.container:
-            self.kubectl = "/host/usr/bin/kubectl"
+            self.kubectl = self._findKubectl("/host")
             if not os.path.exists("/etc/kubernetes"):
                 if self.dryrun:
                     logger.info("DRY-RUN: link /etc/kubernetes from /host/etc/kubernetes")
                 else:
                     os.symlink("/host/etc/kubernetes", "/etc/kubernetes")
+        else:
+            self.kubectl = self._findKubectl()
 
         if not self.dryrun:
             if not os.access(self.kubectl, os.X_OK):
                 raise ProviderFailedException("Command: "+self.kubectl+" not found")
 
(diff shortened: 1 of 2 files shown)


vpavlin:
@markllama Cool:) Could you please fix the issue found by Travis?

markllama:
It looks like there are two problems:

- failed to find kubectl in either /usr/bin/kubectl or /usr/local/bin/kubectl
- raised incorrect exception.

#2 is easy.  #1 makes me wonder where kubectl is for Travis?

vpavlin:
Ah, correct..we run atomicapp with `--dry-run` option so that there are no real docker, kubectl, osc... calls. So the solution (at least for now) is probably to skip the search for `--dry-run`. I haven't looked at Travis infrastructure yet, but I guess we are not able to get kubernetes running there atm - @imcleod ?:)

markllama:
Ahh! *grin*.  OK.  I can set it to bypass and set a default for dry-run.
- Mark

On Wed, Jun 10, 2015 at 10:46 AM, Vaclav Pavlin <notifications@github.com>
wrote:

> Ah, correct..we run atomicapp with --dry-run option so that there are no
> real docker, kubectl, osc... calls. So the solution (at least for now) is
> probably to skip the search for --dry-run. I haven't looked at Travis
> infrastructure yet, but I guess we are not able to get kubernetes running
> there atm - @imcleod <https://github.com/imcleod> ?:)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/projectatomic/atomicapp/pull/146#issuecomment-110780483>
> .
>



-- 
----
Mark Lamourine <markllama@gmail.com>
Dad, Hubbie, Software Developer, System Administrator, Road Cyclist

markllama:
@vpavlin pushed a patch, anything needed to cause re-test?

aweiteka:
LGTM

PR #152, opened 2015-06-16
Add OPT1, OPT2, OPT3  in LABELs

sub-mod:
OPT1, OPT2 LABELs are needed for 
https://github.com/projectatomic/atomicapp/pull/99
https://github.com/projectatomic/atomic/pull/65

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,25 +2,33 @@ FROM centos:centos7
 
 MAINTAINER Vaclav Pavlin <vpavlin@redhat.com>
 
-RUN echo -e "[epel]\nname=epel\nenabled=1\nbaseurl=https://dl.fedoraproject.org/pub/epel/7/x86_64/\ngpgcheck=0" > /etc/yum.repos.d/epel.repo
+WORKDIR /opt/atomicapp
 
-RUN yum install -y --setopt=tsflags=nodocs python-pip docker && \
-    yum clean all
-
+# add all of Atomic App's files to the container image
 ADD atomicapp/ /opt/atomicapp/atomicapp/
 ADD setup.py /opt/atomicapp/
 ADD requirements.txt /opt/atomicapp/
 
-WORKDIR /opt/atomicapp
+# add EPEL repo for pip
+RUN echo -e "[epel]\nname=epel\nenabled=1\nbaseurl=https://dl.fedoraproject.org/pub/epel/7/x86_64/\ngpgcheck=0" > /etc/yum.repos.d/epel.repo
 
-RUN pip install .
+# lets install pip, and gcc for the native extensions
+# and remove all after use
+RUN yum install -y --setopt=tsflags=nodocs python-pip docker gcc && \
+    python setup.py install && \
+    pip install -r ./requirements.txt && \
+    yum remove -y gcc cpp glibc-devel glibc-headers kernel-headers libmpc mpfr python-pip && \
+    yum clean all
 
 WORKDIR /atomicapp
 VOLUME /atomicapp
 
-LABEL RUN docker run -it --rm --privileged --net=host -v `pwd`:/atomicapp -v /run:/run -v /:/host --name NAME -e NAME=NAME -e IMAGE=IMAGE IMAGE -v run /atomicapp
-LABEL STOP docker run -it --rm --privileged --net=host -v `pwd`:/atomicapp -v /run:/run -v /:/host --name NAME -e NAME=NAME -e IMAGE=IMAGE IMAGE -v stop /atomicapp
-LABEL INSTALL docker run --rm -it --privileged -v /run:/run -v `pwd`:/atomicapp -e IMAGE=IMAGE -e NAME=NAME --name NAME IMAGE -v install --destination /atomicapp /application-entity
+LABEL RUN  docker run -it --rm \${OPT1} --privileged -v `pwd`:/atomicapp -v /run:/run -v /:/host --net=host --name \${NAME} -e NAME=\${NAME} -e IMAGE=\${IMAGE} \${IMAGE} -v \${OPT2} run \${OPT3} /atomicapp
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -12,15 +12,14 @@ ADD setup.py /opt/atomicapp/
 ADD requirements.txt /opt/atomicapp/
 
 WORKDIR /opt/atomicapp
-
-RUN pip install .
+RUN python setup.py install
```
goern: why change back from `pip install`?

Dockerfile
```diff
@@ -12,15 +12,14 @@ ADD setup.py /opt/atomicapp/
...
 
 WORKDIR /atomicapp
 VOLUME /atomicapp
-
-LABEL RUN docker run -it --rm --privileged --net=host -v `pwd`:/atomicapp -v /run:/run -v /:/host --name NAME -e NAME=NAME -e IMAGE=IMAGE IMAGE -v run /atomicapp
+ 
+LABEL RUN docker run -it --rm --privileged --net=host -v `pwd`:/atomicapp -v /run:/run -v /:/host --name NAME -e NAME=NAME -e IMAGE=IMAGE IMAGE -v RUNARGS run /atomicapp
```
goern: do you have a pointer/doc what RUNARGS does?

sub-mod: No doc yet, But we discussed this 
https://github.com/projectatomic/atomic/pull/65 Needs RUNARGS in the Dockerfile
i need it for putting -d option into install and run commands

Dockerfile
sub-mod: There is lot a of confusion on which Dockerfile to use.
After trying out all possible options , I found that pip install doesnt seem to install dependecies in requirement.txt and also we need /usr/bin/atomicapp which we can get via python setup.py install
IS there a reason to revert this ?

Dockerfile
sub-mod: atomic --runargs="-d" install <image>
atomic --runargs="-d" run .

goern: if `RUNARGS` is in both cases just `-d` we can move it into the Dockerfile itself, as it is static input

sub-mod: we can have  -q / -d / --dry-run / 
we need to decide if daemon way the default way to hardcode in Dockerfile.

aweiteka:
This seems to be a fine way to use RUNARGS. What is the behavior without support in atomic CLI? I suggest we wait until it's supported in atomic CLI.

vpavlin:
I agree with @aweiteka - the patch makes sense but is dependent on `atomic` patch being accepted

Dockerfile
tkdchen: `pip install` finally reads and installs dependencies from the list passed to `install_requires` of method `setup`, and other dependency related arguments. It does not install dependencies from [Requirement Files](https://pip.pypa.io/en/stable/user_guide.html#requirements-files) automatically.

sub-mod:
@aweiteka currently without this support there no way to inject a args in between atomicapp command
.Only append is supported via atomic command

sub-mod:
@goern kindly review.
This now works for me.

sub-mod:
@aweiteka @goern The atomic patch has been accepted.
I have made the changes in Dockerfile
Kindly review

Dockerfile
```diff
@@ -13,14 +13,16 @@ ADD requirements.txt /opt/atomicapp/
 
 WORKDIR /opt/atomicapp
 
-RUN pip install .
+RUN python setup.py install
+RUN pip install -r ./requirements.txt
```
goern: pls put a yum install gcc && pip install && yum erase gcc+deps somewhere here, so that it is in this layer and doesnt pollute the image ;) rm gcc on line 7

Dockerfile
```diff
@@ -13,14 +13,16 @@ ADD requirements.txt /opt/atomicapp/
...
 
 WORKDIR /atomicapp
 VOLUME /atomicapp
 
-LABEL RUN docker run -it --rm --privileged --net=host -v `pwd`:/atomicapp -v /run:/run -v /:/host --name NAME -e NAME=NAME -e IMAGE=IMAGE IMAGE -v run /atomicapp
-LABEL STOP docker run -it --rm --privileged --net=host -v `pwd`:/atomicapp -v /run:/run -v /:/host --name NAME -e NAME=NAME -e IMAGE=IMAGE IMAGE -v stop /atomicapp
-LABEL INSTALL docker run --rm -it --privileged -v /run:/run -v `pwd`:/atomicapp -e IMAGE=IMAGE -e NAME=NAME --name NAME IMAGE -v install --destination /atomicapp /application-entity
+LABEL RUN  docker run -it --rm \${OPT1} --privileged -v `pwd`:/atomicapp -v /run:/run -v /:/host --net=host --name \${NAME} -e NAME=\${NAME} -e IMAGE=\${IMAGE} \${IMAGE} -v \${OPT2} run \${OPT3} /atomicapp
```
goern: why do we need --net=host?

Dockerfile.fedora
```diff
@@ -1,23 +1,28 @@
 FROM fedora
+
 MAINTAINER langdon <langdon@fedoraproject.org>
 #Derived from Vaclav Pavlin <vpavlin@redhat.com>; https://github.com/vpavlin/atomicapp-run
 
-RUN yum -y install python python-pip docker && \
-    yum update -y && \
+RUN echo -e "[epel]\nname=epel\nenabled=1\nbaseurl=https://dl.fedoraproject.org/pub/epel/7/x86_64/\ngpgcheck=0" > /etc/yum.repos.d/epel.repo
+
+RUN yum install -y --setopt=tsflags=nodocs python-pip docker gcc && \
```
goern: pls implement same gcc pattern as in Dockerfile

Dockerfile
sub-mod: it was already there in the Dockerfile, I didn't add it.

Dockerfile
sub-mod: Does this look good ...RUN yum install -y gcc && python setup.py install && pip install -r ./requirements.txt && yum clean all ?

Dockerfile
goern: see #165

goern:
as blocking https://github.com/projectatomic/atomic/pull/65 has been merged and a little back and forward (thanks for the patience sub-mod) ... LGTM

PR #153, opened 2015-06-16
Init artifacts - artifact list for kubernetes was not re-initialized for each component

markllama:
The kubernetes namespace, kube_order and kubctl values were created as class variables.

The kube_order was later used as the "artifiacts" set for creating components., but was not reset when the KubernetesProvider.init() function executed.

This change places those values as instance variables and initializes them in init()


aweiteka:
LGTM

PR #160, opened 2015-06-19
replaces GNU AGPL by GNU LGPL

### LINKED ISSUE #137: AGPL?
Looking over the repo, I noticed we have gone with the AGPLv3 - which seems an odd, odd choice for something we'd want others to adopt. Can we have a discussion about a more liberal license for this? At least GPL rather than AGPL. Would anyone even be offering Atomic App as a service?

goern:
This PR will relicense Atomic App under the GNU LGPL. This move was motivated by [OSAS](https://community.redhat.com/) in #137. There seem to be no objections to this move:

1. all contributors are current Red Hat employees, and
2. Red Hat is responsible for this repo, and
3. no one in our group of people is opposed to the license change

I assume 3. is true, if anyone is AGAINST the change please use this PR to indicate it. 

This PR will close #137


vpavlin:
LGTM

PR #190, opened 2015-07-27
Fix a logic bug

veillard:
if the id is not found the exception won't be raised by the expected
code for it but by the access to the dict in the message before it

Also testing my workflow with github I'm not familiar with TBH


vpavlin:
LGTM

aweiteka:
LGTM

PR #199, opened 2015-08-07
handle ose config when running from container

### LINKED ISSUE #198: OpenShift provider cannot find config
When running atomicapp from a cloud VM as non-root my OpenShift config is in `/home/cloud-user/.kube/config`. The openshift provider fails with cannot find openshift config. I can provide the openshift config file path in the answers.conf but it must be prepended by `/host/` since we're running in a container. The end-user doesn't know this. Ideally the openshift provider "just works" but if it's not config'd correctly the user should not have to prepend the configfile path with `/host`.

Example answers.conf:

```
[helloapache-app]
image = centos/httpd
hostport = 80
[general]
namespace = default
provider = openshift
openshiftconfig = /host/home/cloud-user/.kube/config
```

aweiteka:
resolves #198

### DIFF
--- a/atomicapp/providers/openshift.py
+++ b/atomicapp/providers/openshift.py
@@ -43,8 +43,7 @@ class OpenShiftProvider(Provider):
             logger.warning("Configuration option 'openshiftconfig' not found")
 
         if not self.config_file or not os.access(self.config_file, os.R_OK):
-            raise ProviderFailedException(
-                "Cannot access configuration file %s" % self.config_file)
+            raise ProviderFailedException("Cannot access configuration file %s. Try adding 'openshiftconfig = /path/to/your/.kube/config' in the [general] section of the answers.conf file." % self.config_file)
 
     def _callCli(self, path):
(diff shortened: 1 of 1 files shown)


goern:
LGTM

atomicapp/providers/openshift.py
```diff
@@ -34,11 +34,13 @@ def init(self):
...
         if "openshiftconfig" in self.config:
             self.config_file = self.config["openshiftconfig"]
+            if self.container:
+                self.config_file = "/host" + self.config_file
         else:
             logger.warning("Configuration option 'openshiftconfig' not found")
 
         if not self.config_file or not os.access(self.config_file, os.R_OK):
-            raise ProviderFailedException("Cannot access configuration file %s" % self.config_file)
+            raise ProviderFailedException("Cannot access configuration file %s. Try adding 'openshiftconfig = /path/to/your/.kube/config' in the [general] section of the answers.conf file." % self.config_file)
```
goern: @vpavlin could you take this one line into https://github.com/projectatomic/atomicapp/pull/213 so we can close this PR due to dup?

PR #200, opened 2015-08-07
Implement stop for OpenShift provider

### LINKED ISSUE #135: implement STOP for OpenShift provider
see https://trello.com/c/k14cmfdj

aweiteka:
There are several ways to stop resources on
OpenShift. The simplest and most analagous to
stop is "stop". This will stop and delete but
it is not surprising since that's what the oc
command does.

resolves #135 

Note: we do not support stop from template yet since this is a bit more complex.

### DIFF
--- a/atomicapp/providers/openshift.py
+++ b/atomicapp/providers/openshift.py
@@ -40,8 +40,8 @@ class OpenShiftProvider(Provider):
         if not self.config_file or not os.access(self.config_file, os.R_OK):
             raise ProviderFailedException("Cannot access configuration file %s" % self.config_file)
 
-    def _callCli(self, path):
-        cmd = [self.cli, "--config=%s" % self.config_file, "create", "-f", path]
+    def _callCli(self, path, action="create"):
+        cmd = [self.cli, "--config=%s" % self.config_file, action, "-f", path]
 
         if self.dryrun:
             logger.info("Calling: %s", " ".join(cmd))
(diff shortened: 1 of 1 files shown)


goern:
LGTM

vpavlin:
I checked the `oc stop` command and found https://github.com/openshift/origin/blob/master/docs/generated/oc_by_example_content.adoc#oc-stop

which says _Deprecated_, so I asked OpenShift guys and they told me we should use `oc delete` instead.

aweiteka:
@vpavlin good to know! Ok, so let's not merge this as is. Options:

1. Change `atomic stop` to call `oc delete`. This doesn't feel right; it feels too destructive for a stop action.
1. There is no other option. :)

goern:
If we assume that 'stopping an Atomic App' means, that we want to leave it deployed but not any part of it is running, than we should
1. figure out if a replication controller is deployed
1.1. if so, set the number of replicas to 0
2. figure out if any pods are deployed
2.1. if so, delete the pods

This will leave all the DC, BC and RC and templates in place (aka deployed) and stop all execution of the application (aka the pods)

Starting the Atomic App again will result in redeploying any pod configuration and/or setting the number of replicas to 1.

aweiteka:
@goern so there *is* another option. I agree. I'll rework this to handle it less destructively. This begs the question: how do we remove the app? `uninstall` seems like a good fit but does not match the current `install` behavior. That topic is another discussion and PR, I think.

aweiteka:
Need to update using same approach as #221 by @navidshaikh

aweiteka:
Based on current OpenShift provider direction (see #321) I don't think this is required any more. The plan to rely on native OpenShift control (`oc` client or web UI) to control lifecycle outside of run.

cdrage:
Hey @aweiteka can we close this as we're close to implement the API only implementation for openshift?

dustymabe:
@kadel is working on the stop for openshift provider using the new openshift API code.

I'll close this once we have a new PR or Issue # to reference.

kadel:
I've just crated issue for stop for new OpenShift provider  #458

dustymabe:
Closed in favor of issue #458 and future work to resolve it.

PR #202, opened 2015-08-08
Bugfix/docker whitespace

### LINKED ISSUE #201: Strip whitespace with docker provider
@jasonbrooks found an issue where if the docker provider artifact run line has two whitespaces it fails. Fix cmd `subprocess.check_call(cmd)` in `provider/docker.py`. This should be super easy I just couldn't get to it.

aweiteka:
resolves #201

### DIFF
--- a/atomicapp/providers/openshift.py
+++ b/atomicapp/providers/openshift.py
@@ -37,6 +37,8 @@ class OpenShiftProvider(Provider):
 
         if "openshiftconfig" in self.config:
             self.config_file = self.config["openshiftconfig"]
+            if self.container:
+                self.config_file = "/host" + self.config_file
         else:
             logger.warning("Configuration option 'openshiftconfig' not found")
 
--- a/atomicapp/providers/openshift.py
+++ b/atomicapp/providers/openshift.py
@@ -38,7 +38,7 @@ class OpenShiftProvider(Provider):
         if "openshiftconfig" in self.config:
             self.config_file = self.config["openshiftconfig"]
             if self.container:
-                self.config_file = "/host" + self.config_file
+                self.config_file = os.path.join("/host", self.config_file.lstrip("/"))
         else:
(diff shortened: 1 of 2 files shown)


atomicapp/providers/openshift.py
```diff
@@ -34,6 +34,8 @@ def init(self):
 
         if "openshiftconfig" in self.config:
             self.config_file = self.config["openshiftconfig"]
+            if self.container:
+                self.config_file = "/host" + self.config_file
```
navidshaikh: For path concatenation, its better to use `os.path.join`

aweiteka:
@navidshaikh `os.path.join` doesn't seem to work in this context where my config string is an absolute path that starts with a "/", for example `/home/user/.kube/config`. I cannot use `os.path.join` to prepend `/host`. Any suggestions? Leave it as is?

navidshaikh:
Although this is not a blocker, rather a suggestion for best practice.

Regarding, the path concatenation using `os.path.join` , I don't see any problem concatenating the absolute paths.

```
In [4]: import os

In [5]: os.path.join("/host", "user/.kube/config")
Out[5]: '/host/user/.kube/config'

```

vpavlin:
Actually, the problem is when both strings start with `/`

Using `self.config_file.lstrip("/")` should help
```
In [1]: import os

In [2]: os.path.join("/host", "/usr/lib/stuff")
Out[2]: '/usr/lib/stuff'

In [3]: one = "/host"

In [4]: two = "/usr/lib/stuff"

In [6]: os.path.join(one, two.lstrip("/"))
Out[6]: '/host/usr/lib/stuff'

In [7]: 

```

navidshaikh:
> Actually, the problem is when both strings start with /
Ah! my bad!

+1 for `self.config_file.lstrip("/")`

PR #207, opened 2015-08-11
Create a lock file and fail if it's already locked to prevent concurrent Atomic Apps running

vpavlin:
Added dependency for `lockfile`

Creates lock file at `/var/lock/atomicapp.lock`

### DIFF
--- a/atomicapp/constants.py
+++ b/atomicapp/constants.py
@@ -11,6 +11,7 @@ ANSWERS_FILE = "answers.conf"
 ANSWERS_FILE_SAMPLE = "answers.conf.sample"
 ANSWERS_FILE_SAMPLE_FORMAT = 'ini'
 WORKDIR = ".workdir"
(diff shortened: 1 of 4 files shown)


vpavlin:
Fixes #136

atomicapp/constants.py
```diff
@@ -11,6 +11,7 @@
 ANSWERS_FILE_SAMPLE = "answers.conf.sample"
 ANSWERS_FILE_SAMPLE_FORMAT = 'ini'
 WORKDIR = ".workdir"
+LOCK_FILE = "/var/lock/atomicapp.lock"
```
goern: aint it `/run/lock/` ? ;)

vpavlin: Hmm..actually both, but it should be `/run`

```
lrwxrwxrwx. 1 root root   11 May 21 20:55 lock -> ../run/lock
```

PR #213, opened 2015-08-13
Don't use /host string directly in code

vpavlin:
This patch replaces all `/host` occurrences in the code with either `HOST_DIR` constant, or call to a `Utils.getRoot()`


goern:
LGTM
seems like it is pretty hard to write a test case for things like this?!

vpavlin:
hmm..I'll try to think about it..

goern:
still LGTM ;)

PR #214, opened 2015-08-13
Providers: Update link to Nulecule's paramsObject

wking:
Catch up with projectatomic/nulecule@ab067ff5 (Refactor spec
directory, 2015-07-09).  There's a tradeoff here between a permanent
link to the dead v0.0.2 spec and a link to the live current spec.  The
former won't break (assuming no further changes like ab067ff5), but it
would require manual updates whenever atomicapp bumps its Nulecule
version.  The live link won't need to be bumped, but it may point at a
version of the spec that doesn't match the atomicapp implementation.

I've gone with the live link here, but I'm happy to reroll if the
maintainers want something that they need to explicitly bump.

### DIFF
--- a/Providers.asciidoc
+++ b/Providers.asciidoc
@@ -80,7 +80,7 @@ As of 0.1.0 of Atomic App, only
 http://docs.openshift.org/latest/dev_guide/templates.html[OpenShift templates] will be
 preprocessed by Atomic App during the run phase: parameters within a
 template file will be replaced by the
-https://github.com/projectatomic/nulecule/blob/master/spec/0.0.2/README.md#paramsObject[params]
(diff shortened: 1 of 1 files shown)


goern:
I would go for explicit bumping, as it a) enforces our discipline on quality checks when releasing atomicapp and b) not all Nulecule Spec features must/will be implemented by Atomic App, so pointing to the exact spec version may be better.

@aweiteka @vpavlin ?

aweiteka:
@wking thanks for the PR!

aweiteka:
@goern sorry, I didn't follow proper procedure here. I took your comment as an ACK. I'm not sure why in hindsight. Better next time!

goern:
man oh man ;)

Issue #226, opened 2015-08-24
No lockfile installation on python setup.py install


charliedrage:
lockfile is not installed by default when running 
```bash
python setup.py install
```
in the default dir

see pull #227 for fix as well as changes to Dockerfile / versioning conventions

PR #227, opened 2015-08-24
update versioning in setup, requirements and Dockerfile

charliedrage:
relates to issue #226


charliedrage:
with this change, we can also remove 
```bash
pip install -r ./requirements.txt && \ 
``` 
from the list of commands when building / tagging images.

setup.py
```diff
@@ -40,16 +40,16 @@ def _install_requirements():
     return requirements
 
 setup(
-    name='atomicapp',
-    version='0.1.9',
```
goern: could you read the value from https://github.com/projectatomic/atomicapp/blob/master/atomicapp/constants.py#L26 ?

charliedrage: wouldn't it be easier to have a file in the root folder named VERSION that you can read it from? seems that whenever a new version comes out you have to s/0.1.9/0.2.0/ every file

charliedrage: also, it would make it easier for contributors which version they're working on by simply
```bash
cat VERSION
```
the dir

vpavlin:
Hmm..we should also fix a project url and maybe change the author to something more generic than my name:)

Dockerfile
```diff
@@ -11,18 +11,16 @@ WORKDIR /opt/atomicapp
...
 
 # lets install pip, and gcc for the native extensions
 # and remove all after use
-RUN yum install -y --setopt=tsflags=nodocs python-pip python-setuptools docker gcc && \
-    python setup.py install && \
-    pip install -r ./requirements.txt && \
-    yum remove -y gcc cpp glibc-devel glibc-headers kernel-headers libmpc mpfr python-pip && \
+RUN yum install -y --setopt=tsflags=nodocs python-pip python-setuptools docker gcc
+RUN python setup.py install
+RUN yum remove -y gcc cpp glibc-devel glibc-headers kernel-headers libmpc mpfr python-pip && \
```
charliedrage: Reason for multiple lines of RUN is to speed up cache when doing 'docker build -t projectatomic .'

charliedrage:
will need an ack from @goern before merging, as there are some Dockerfile changes :+1:

Dockerfile
```diff
@@ -11,18 +11,16 @@ WORKDIR /opt/atomicapp
...
 # add EPEL repo for pip
 RUN echo -e "[epel]\nname=epel\nenabled=1\nbaseurl=https://dl.fedoraproject.org/pub/epel/7/x86_64/\ngpgcheck=0" > /etc/yum.repos.d/epel.repo
 
 # lets install pip, and gcc for the native extensions
 # and remove all after use
-RUN yum install -y --setopt=tsflags=nodocs python-pip python-setuptools docker gcc && \
-    python setup.py install && \
-    pip install -r ./requirements.txt && \
-    yum remove -y gcc cpp glibc-devel glibc-headers kernel-headers libmpc mpfr python-pip && \
+RUN yum install -y --setopt=tsflags=nodocs python-pip python-setuptools docker gcc
```
vpavlin: You need to add `&& yum clean all` at the end, otherwise we will end up carrying worthless data in the image. (i.e. every RUN creates a layer and every layer where yum is called needs to clear the cache otherwise it stays in layer although it's removed the next layer)

Dockerfile.fedora
```diff
@@ -1,28 +1,26 @@
...
 # add EPEL repo for pip
 RUN echo -e "[epel]\nname=epel\nenabled=1\nbaseurl=https://dl.fedoraproject.org/pub/epel/7/x86_64/\ngpgcheck=0" > /etc/yum.repos.d/epel.repo
 
 # lets install pip, and gcc for the native extensions
 # and remove all after use
-RUN dnf install -y --setopt=tsflags=nodocs python-pip python-setuptools docker gcc && \
-    python setup.py install && \
-    pip install -r ./requirements.txt && \
-    dnf remove -y gcc cpp glibc-devel glibc-headers kernel-headers libmpc mpfr python-pip && \
+RUN dnf install -y --setopt=tsflags=nodocs python-pip python-setuptools docker gcc
```
vpavlin: Same as above - missing `dnf clean all`

Dockerfile.fedora
```diff
@@ -1,28 +1,26 @@
...
+RUN python setup.py install
+RUN dnf remove -y gcc cpp glibc-devel glibc-headers kernel-headers libmpc mpfr python-pip && \
```
goern: and pls consolidate all RUNs in one RUN line, so we will create one layer and have it cleaned up in the end

Dockerfile
goern: caching seems to be irrelevant for a base image, I would go for small size rather than the time needed to build to image.

charliedrage:
Will be re-doing a cleaner pull request**

PR #233, opened 2015-08-26
Update versioning of constants and setup as well as cleanup Dockerfiles

charliedrage:
Fixes issue #226


goern:
LGTM @vpavlin @aweiteka

dustymabe:
This breaks when you build on top of atomicapp. `constants.py` tries to open the `VERSION` and `NULECULE` files in the current workding directory except that the WORKDIR changed from `/opt/atomicapp` to `/atomicapp` later in the Dockerfile for atomicapp. 

Should we be using absolute file paths here? I think I would prefer that to using WORKDIR.

dustymabe:
just so you know.. here is what you get when you try to run a helloapache that was built on top of an atomicapp that has this commit:

```
[root@f22 foo]$ atomic run delloapache
docker run -it --rm  --privileged -v /tmp/foo:/atomicapp -v /run:/run -v /:/host --net=host --name delloapache -e NAME=delloapache -e IMAGE=delloapache delloapache -v  run  /atomicapp
  File "/usr/bin/atomicapp", line 9, in <module>
    load_entry_point('atomicapp==0.1.3', 'console_scripts', 'atomicapp')()
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 378, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 2566, in load_entry_point
    return ep.load()
  File "/usr/lib/python2.7/site-packages/pkg_resources.py", line 2260, in load
    entry = __import__(self.module_name, globals(),globals(), ['__name__'])
  File "/usr/lib/python2.7/site-packages/atomicapp-0.1.3-py2.7.egg/atomicapp/cli/main.py", line 22, in <module>
    from atomicapp.run import Run
  File "/usr/lib/python2.7/site-packages/atomicapp-0.1.3-py2.7.egg/atomicapp/run.py", line 26, in <module>
    from nulecule_base import Nulecule_Base
  File "/usr/lib/python2.7/site-packages/atomicapp-0.1.3-py2.7.egg/atomicapp/nulecule_base.py", line 26, in <module>
    from constants import MAIN_FILE, GLOBAL_CONF, DEFAULT_PROVIDER, PARAMS_KEY, \
  File "/usr/lib/python2.7/site-packages/atomicapp-0.1.3-py2.7.egg/atomicapp/constants.py", line 35, in <module>
    __ATOMICAPPVERSION__ = _get_version('VERSION')
  File "/usr/lib/python2.7/site-packages/atomicapp-0.1.3-py2.7.egg/atomicapp/constants.py", line 32, in _get_version
    raise RuntimeError("Can't open file with: %s", repr(ex))
RuntimeError: ("Can't open file with: %s", "IOError(2, 'No such file or directory')")
```

charliedrage:
@dustymabe this means that examples built on 0.1.2 cannot be ran by atomicapp 0.1.3. what alternative where you thinking of doing with WORKDIR?

goern:
oh well, containers always deliver surprises... what if the file (whereever it lives) gets overwritten by an layer ontop? can we prevent this? 
Alternatively we could leave the version constants in constants.py and straighten out the release process, which would be a good idea anyways.

PR #237, opened 2015-08-27
Updated tox.ini to execute the pep8 tests

### LINKED ISSUE #236: Update tox.ini to execute the pep8 tests
Currently if i do tox -e pep8 it does not run and fails to create the test env. Need to update the tox.ini and add test-requirements for flake8

coolsvap:
Update tox.ini to execute the pep8 tests

Currently if i do tox -e pep8
- it does not run
- fails to create the test env

Updated following,
- tox.ini to include testenv

Added test-requirements for flake8

Resolves #236


tox.ini
```diff
@@ -1,9 +1,19 @@
...
+deps = -r{toxinidir}/requirements.txt
+       -r{toxinidir}/test-requirements.txt
+
+[testenv:pep8]
+commands =
+    flake8
 
 [flake8]
 ignore = E226,E302,E41,E251,C901
 max-line-length = 160
```
navidshaikh: I think the max-line-length should be a bit smaller, 100 ?

@goern @vpavlin : thoughts?

goern:
thanks for that. do any requirements for our developers result from your change? (i dont think so but unsure)

coolsvap:
Couple of requirements that will be addressed are

- Check code check conformance on dev env instead of submitting to CI multiple times
- We can always expand on the testenvs as per updates to ci

goern:
great, could you check how that releates to https://github.com/projectatomic/atomicapp/blob/master/CONTRIBUTING.md ? thx!

coolsvap:
@goern please have a look at the updated contributing.md

goern:
LGTM @vpavlin @aweiteka

vpavlin:
LGTM

Issue #212, opened 2015-08-12
Issue with multiple tiers app k8s artifacts processing 


navidshaikh:
Multiple tier app:
App having more than one tier for eg: db tier, web tier, app tier.

Kubernetes resources (pods, rc and services) manifests are overridden while processing the manifests for a multiple tier app.

So the last kubernetes artifact per type (pod/rc/services) described in Nulecule file is the one which is deployed while the earlier manifests data is lost.

Example: 
Atomic App: <https://hub.docker.com/r/swordphilic/gitlab-centos7-atomicapp/>
Source of nulecule app: <https://github.com/navidshaikh/nulecule_examples/tree/master/gitlab/v1beta1>

```
$ atomic run swordphilic/gitlab-centos7-atomicapp
docker run -it --rm --privileged --net=host -v `pwd`:/atomicapp -v /run:/run -v /:/host --name gitlab-centos7-atomicapp -e NAME=gitlab-centos7-atomicapp -e IMAGE=swordphilic/gitlab-centos7-atomicapp swordphilic/gitlab-centos7-atomicapp -v run /atomicapp
2015-08-12 11:45:17,676 - atomicapp.install - INFO - App name is swordphilic/gitlab-centos7-atomicapp, will be populated to /atomicapp
2015-08-12 11:45:17,677 - atomicapp.nulecule_base - DEBUG - {'namespace': 'default', 'provider': 'kubernetes'}
2015-08-12 11:45:18,363 - atomicapp.nulecule_base - DEBUG - Output of docker images cmd: 
Trying to pull repository docker.io/swordphilic/gitlab-centos7-atomicapp ...
8cda1932f4f8: Download complete 
f1b10cd84249: Download complete 
c852f6d61e65: Download complete 
7322fbe74aa5: Download complete 
b9e8df520b94: Download complete 
36873d767517: Download complete 
a71d35ab23d4: Download complete 
300fdceaf673: Download complete 
358243682dfa: Download complete 
06697d00bbd3: Download complete 
96925fa4ee9e: Download complete 
ed0a83cfde70: Download complete 
276001559a51: Download complete 
593794fa9ab7: Download complete 
d258dd3252f0: Download complete 
aeea42f054c6: Download complete 
db828900997a: Download complete 
269c050a3679: Download complete 
7c5fc272a6bb: Download complete 
79abafa3f691: Download complete 
d7589e14d58d: Download complete 
7285de96ae9f: Download complete 
[...]

PR #241, opened 2015-08-28
Fixes processing multiple k8s resource artifacts

navidshaikh:
Fixes: https://github.com/projectatomic/atomicapp/issues/212

 - Nulecule multiple resource artifacts defined per kind of resource can now be
 processed without overriding earlier artifacts
 - Modifies the methods names as per <https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables>
 - Alters order of imports as per <https://www.python.org/dev/peps/pep-0008/#imports>
 - Adds docstring to methods

### DIFF
--- a/atomicapp/providers/kubernetes.py
+++ b/atomicapp/providers/kubernetes.py
@@ -154,8 +154,9 @@ class KubernetesProvider(Provider):
         :arg replicas: Replica size to scale to.
         """
         rname = self._resource_identity(path)
-        cmd = [self.kubectl, "scale", "rc", rname, "--replicas=0", "--namespace=%s" %
-               self.namespace]
+        cmd = [self.kubectl, "scale", "rc", rname,
+               "--replicas=%s" % str(replicas),
+               "--namespace=%s" % self.namespace]
 
         if self.dryrun:
             logger.info("DRY-RUN: %s", " ".join(cmd))
--- a/atomicapp/providers/kubernetes.py
+++ b/atomicapp/providers/kubernetes.py
@@ -31,6 +31,7 @@ logger = logging.getLogger(__name__)
 
 
 class KubernetesProvider(Provider):
+
     """Operations for Kubernetes provider is implemented in this class.
     This class implements deploy, stop and undeploy of an atomicapp on
     Kubernetes provider.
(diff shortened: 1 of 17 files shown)


atomicapp/providers/kubernetes.py
```diff
@@ -128,9 +147,14 @@ def _resourceIdentity(self, path):
...
-        rname = self._resourceIdentity(path)
-        cmd = [self.kubectl, "resize", "rc", rname, "--replicas=0", "--namespace=%s" %
+    def _scale_replicas(self, path, replicas=0):
+        """Scales replicationController to specified replicas size
+
+        :arg path: Path to replicationController manifest
+        :arg replicas: Replica size to scale to.
+        """
+        rname = self._resource_identity(path)
+        cmd = [self.kubectl, "scale", "rc", rname, "--replicas=0", "--namespace=%s" %
```
navidshaikh: With latest `v1` APIs, `scale` is introduced and recommended to be used.

vpavlin:
From what I was told earlier the order of manifest kinds matter (i.e. first you deploy `service`, then `rc` and/or `pod`). Brief look at your code gives me an impression that this order is not ensured.

1. Is the order requirement still the case for kubes?
2. Am I right with the assumption that the order went away with your path?

navidshaikh:
@vpavlin : 
I think the order in which the resources should be deployed must be ensured by the Nulecule author and IMO we should not alter the sequence of the k8s resource listing and their deployment.

atomicapp/providers/kubernetes.py
```diff
@@ -105,18 +112,30 @@ def _callK8s(self, path):
                 printErrorStatus("cmd failed: " + " ".join(cmd))
                 raise
 
-    def prepareOrder(self):
+    def process_k8s_artifacts(self):
+        """Processes Kubernetes manifests files and checks if manifest under
```
goern: this is 'manifest after processing' ?

atomicapp/providers/kubernetes.py
```diff
@@ -128,9 +147,14 @@ def _resourceIdentity(self, path):
         else:
             raise ProviderFailedException("Malformed kube file: %s" % path)
 
-    def _resetReplicas(self, path):
-        rname = self._resourceIdentity(path)
-        cmd = [self.kubectl, "resize", "rc", rname, "--replicas=0", "--namespace=%s" %
+    def _scale_replicas(self, path, replicas=0):
```
goern: this is not _scale_replicas() it is _set_replica_count_to_zero()

atomicapp/providers/kubernetes.py
goern: at least here is a hardcoded 0

goern:
I think we dont need any sequence in deploying to kubernetes or openshift.

vpavlin:
As I said, originally there was a requirement for deploying service before pods/rcs

atomicapp/providers/kubernetes.py
navidshaikh: Correct!
Fixed in next patch.

atomicapp/providers/kubernetes.py
navidshaikh: Sorry, I did not get this!

aweiteka:
LGTM

navidshaikh:
Added new atomicapp example of gitlab app for the fix of this issue. This atomicapp app particularly references multiple Kubernetes artifacts for 3 different tiers and the tests perform `install`, `run` and `stop` operations on them.

vpavlin:
Thanks @navidshaikh! LGTM

Issue #244, opened 2015-08-28
merge #233 breaks stuff


goern:
just to have the link to the debugging output in #233

goern:
closed by merge #246

PR #246, opened 2015-08-28
reverts break in versioning

charliedrage:
fixes break on versioning in commit #233 
fixes issue #244


dustymabe:
Would you also want to delete the `VERSION` and the other file and remove them from being included in the Dockerfile? If we decided to do something different for this in the future then that stuff might get left around and forgotten.

charliedrage:
@dustymabe I've pushed a separate commit to the breakin one to remove the dynamic versioning from the Dockerfiles.

dustymabe:
@charliedrage maybe squash into a single commit? Otherwise LGTM

charliedrage:
@dustymabe thought it'd be nice to see what exactly broke :) but gone ahead and squashed!

dustymabe:
LGTM @goern ^^

PR #254, opened 2015-09-01
added .workdir/ to ignore list

### LINKED ISSUE #253: Unnecessary `.workdir` directories present in each test atomicapp examples
While running tests, `.workdir` is created in each atomicapp example directory in `tests/cached_nulecules` directory. And it is part of the git repository.

These directories need to be removed and `.gitignore` should be updated accordingly.

goern:
this fixes #253


.gitignore
```diff
@@ -1,3 +1,6 @@
+# ignore specific to Atomic App
+.workdir/
```
navidshaikh: `tests/cached_nulecules/*/.workdir` ?

goern: no, its all .workdir anywhere below .

tests/test_cli.py
```diff
@@ -163,8 +163,7 @@ def test_install_with_gitlab_app(self):
             tests_root,
             "cached_nulecules/gitlab/.workdir")
 
-        assert set(os.listdir(work_dir)) == \
-            set(["gitlab", "postgresql", "redis"])
+        assert os.path.isdir(work_dir) == False
```
navidshaikh: This `assert` adds the ordering constraint. If this test is moved after `run` it will fail -because the `run` will create the `.workdir`

navidshaikh:
LGTM! Lets get this merged?
 The test ordering as addressed above and the separation of concerns for testing particular behavior should be part of another PR.

PR #234, opened 2015-08-26
Add --provider option

aweiteka:
The --provider option is constrained to supported providers.
Providers can be specified in answers.conf. If --provider opt
is specified it will be overridden. Otherwise the default provider
is used.

Resolves #171

PR #255, opened 2015-09-02
dont ask for missing artifacts if stopping

charliedrage:
While working on undeployment methods for the Docker provider, I found that `atomicapp stop` will ask for missing artifcact values despite stopping / removing the container.

fixes issue #256

### DIFF
--- a/atomicapp/run.py
+++ b/atomicapp/run.py
@@ -126,7 +126,10 @@ class Run(object):
 
     def _applyTemplate(self, data, component):
         template = Template(data)
-        config = self.nulecule_base.getValues(component)
+        if self.stop:
+            config = self.nulecule_base.getValues(component, skip_asking=True)
+        else:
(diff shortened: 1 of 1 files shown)


vpavlin:
LGTM, thanks

navidshaikh:
Please add a test for this.

goern:
I think its https://github.com/projectatomic/atomicapp/blob/master/atomicapp/nulecule_base.py#L170 that needs a unit test...

charliedrage:
once #234 is merged I can do a test with --provider=docker too in order to prevent from asking about artifacts (of course once undeployment works with the docker provider, which I'm working on)

@goern should we start creating a file for unit testing then? https://docs.python.org/2/library/unittest.html

charliedrage:
@navidshaikh will be adding test to this :) looking at / going through unit testing on pytest

Issue #259, opened 2015-09-04
atomicapp run gives entire config, atomicapp stop ignores some config files


charliedrage:
I'm finding that when running atomicapp run it provides:

```
DEBUG - Given config: {u'root_pass': u'MySQLPass', u'db_user': 'test', u'namespace': u'tester', u'db_pass': 'test', u'db_name': 'test', u'provider': u'docker'}
```

However, when you run atomicapp stop it only provides the parameters:
```
DEBUG - Given config: {u'db_pass': None, u'db_name': None, u'db_user': None, u'root_pass': u'MySQLPass'}
```

Thus, when trying:
```python
        logger.debug("Given config: %s", self.config)                                                                                                                                                                                                          
        if self.config.get("namespace"):                                                                                                                                                                                                                       
            self.namespace = self.config.get("namespace")                                                                                                                                                                                                      
        logger.debug("DEFAULT NAMESPACE: %s", self.namespace)    
```

in `providers/docker.py` or `providers/kubernetes.py` results in only the default name.

charliedrage:
closed. #260 fixes this :-)

PR #260, opened 2015-09-05
artifact logic bugfix

charliedrage:
fixes issue #259 

Logic error in nulecule_base when obtaining artifact + answers.conf file for stopping / undeploying a provider.

Namespace + provider variable from answers is not passed thus you cannot obtain the configuration information for undeployment.

### DIFF
--- a/atomicapp/nulecule_base.py
+++ b/atomicapp/nulecule_base.py
@@ -172,7 +172,12 @@ class Nulecule_Base(object):
         return params
 
     def getValues(self, component=GLOBAL_CONF, skip_asking=False):
-        params = self.get(component, not skip_asking)
+        if skip_asking:
+            params = self.get(component, skip_asking)
+            values = self._getComponentValues(params, not skip_asking)
+        else:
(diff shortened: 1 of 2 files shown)


goern:
do you think you can write up some test cases for the whole getValue() scenery?

charliedrage:
Sure.

So you want this implemented in our Python testing (pytest) or for me to
just show examples as to why this is a bug?

goern:
some tests in atomicapp, if you look at https://coveralls.io/builds/3491575/source?filename=atomicapp%2Fnulecule_base.py#L95 and https://coveralls.io/builds/3491575/source?filename=atomicapp%2Fnulecule_base.py#L211 it looks like most of this is not covered by any test. It would be great to improve a little...

charliedrage:
@goern 
I have added coverage for nulecule_base, however, we will need to add more for run.py, although first we need to make run.py a bit more modular so we can integrate some tests more easily.

My other suggestion is we can use a logger capture method within our test environment to catch logging errors / info and parse them.

Issue #182, opened 2015-07-22
add random part to name of containers


goern:
Containers started by atomicapp should carry unique names. This problem is solved to the OpenShift provider, as OpenShift is responsible for generating the names, same applies for the Kubernetes provider. Using the docker provider, an application packager can choose and name (via `--name` of docker command) and this may lead to name collisions and therefore to containers not starting up as expected.

* Either we document a good practice (not use any `--name` within the docker provider) or
* we invent some logic in atomicapp that overwrites any given name with a uuid'ed name

bexelbie:
I believe we should add a unique name generator so that the resulting container remains identifiable as part of the application.  Perhaps we could use something where we start the container, get a docker random name and then `docker rename` the container to application-dockername

goern:
closed by #263

PR #263, opened 2015-09-09
Undeploy added to Docker provider

charliedrage:
Please see issue #264 for info!

Comment here for code issues / suggestions.

### DIFF
--- a/atomicapp/providers/docker.py
+++ b/atomicapp/providers/docker.py
@@ -17,11 +17,13 @@
  along with Atomic App. If not, see <http://www.gnu.org/licenses/>.
 """
 
-from atomicapp.plugin import Provider, ProviderFailedException
 import os
 import subprocess
-
+import re
 import logging
+from atomicapp.constants import DEFAULT_CONTAINER_NAME, DEFAULT_NAMESPACE
+from atomicapp.plugin import Provider, ProviderFailedException
+from atomicapp.utils import Utils
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


dustymabe:
this will most likely take care of https://github.com/projectatomic/atomicapp/issues/182 . 

@charliedrage could you confirm?

charliedrage:
@dustymabe yup, that would solve #182 completely.

This would also (partly) solve PR #200 / Issue #135 in regards to stop.

atomicapp/providers/docker.py
```diff
@@ -50,16 +59,74 @@ def init(self):
...
+            # if --name is located in the artifact file, append it with the docker deployment naming convention
+            if '--name' in labels:
+                cname = labels.index('--name') + 1
+                logger.info("Using name located within artifacts/docker: %s", cname)
+                labels[cname] = "%s_%s_%s" % (self.default_name, labels[cname], Utils.getUniqueUUID())
 
-            cmd = label_run.split()
+            # otherwise, name it using the namespace located in answers.conf
+            else:
+                labels.insert(labels.index('run') + 1, "--name=%s_%s_%s" % (self.default_name, self.namespace, Utils.getUniqueUUID()))
```
rtnpro: PEP8!

charliedrage: @rtnpro I did run PEP8. can you provide code what format I should have it in? it's the comments yeah?

atomicapp/providers/docker.py
```diff
@@ -50,16 +59,74 @@ def init(self):
                    "on host (%s). Please update your host." % (client, server))
             raise ProviderFailedException(msg)
 
+    """
+    Deploys a Docker container based upon the /artifacts file located in the working dir.
+    If --name is provided within /artifacts/docker, atomicapp will tag the corresponsind image as: atomic_NAME_HASH
```
dustymabe: s/corresponsind/corresponding

dustymabe:
@charliedrage should we update https://github.com/projectatomic/atomicapp/blob/master/docs/providers.md since now (I think) docker would require the namespace configuration value as well?

atomicapp/providers/docker.py
rtnpro: I tkink, I should have said, FLAKE8. The line width is more than 80 chars, I had to scroll horizontally to check them out ;)

atomicapp/providers/docker.py
```diff
@@ -50,16 +59,74 @@ def init(self):
...
+            with open(artifact_path, "r") as fp:
+                label_run = fp.read().strip()
+            labels = label_run.split()
+
+            # If any artifacts are labelled by name, add it to a container dict list
+            if '--name' in labels:
+                artifact_names.append(labels[labels.index('--name') + 1])
+                logger.debug("artifact cnames: %s" % artifact_names)
+
+        # TODO: If anyone know's a cleaner way to grep all Docker containers by name; please submit a PR
```
vpavlin: RTFM for docker-ps:-)

```
docker ps -a -q --filter=name=$NAME_STUB
```

where `$NAME_STUB` can be full name or just a part of it:)

atomicapp/providers/docker.py
```diff
@@ -30,6 +32,13 @@ class DockerProvider(Provider):
     key = "docker"
 
     def init(self):
+        self.namespace = DEFAULT_NAMESPACE
+        self.default_name = DEFAULT_CONTAINER_NAME
+
+        logger.debug("Given config: %s", self.config)
+        if self.config.get("namespace"):
+            self.namespace = self.config.get("namespace")
+        logger.debug("DEFAULT NAMESPACE: %s", self.namespace)
```
vpavlin: I don't think this prints "Default namespace" as it can be replaced by user value

atomicapp/providers/docker.py
```diff
@@ -50,16 +59,74 @@ def init(self):
                    "on host (%s). Please update your host." % (client, server))
             raise ProviderFailedException(msg)
 
+    """
```
vpavlin: Shouldn't this be a `__doc__` of the function?

atomicapp/providers/docker.py
```diff
@@ -50,16 +59,74 @@ def init(self):
...
+
         for artifact in self.artifacts:
             artifact_path = os.path.join(self.path, artifact)
             label_run = None
             with open(artifact_path, "r") as fp:
                 label_run = fp.read().strip()
+            labels = label_run.split()
+
+            # if --name is located in the artifact file, append it with the docker deployment naming convention
+            if '--name' in labels:
```
vpavlin: Please don't: https://github.com/projectatomic/atomicapp/issues/264#issuecomment-139470312

atomicapp/utils.py
```diff
@@ -214,3 +215,9 @@ def getRoot():
             return HOST_DIR
         else:
             return "/"
+
+    # generates a unique 12 character UUID
+    @staticmethod
+    def getUniqueUUID():
```
vpavlin: When this patch gets merged, please replace my `random` stuff with this method (https://github.com/projectatomic/atomicapp/blob/master/atomicapp/install.py#L90)

charliedrage: Wouldn't it be better to generate a unique UUID with a combination of letters and numbers similar to Docker and rkt's way of generating ID's? Less chance for name colission if someone does multiple deploys.

atomicapp/providers/docker.py
charliedrage: @vpavlin I tried this and I found that --filter doesn't support regex the filter command only shows container ID's not names.

atomicapp/providers/docker.py
charliedrage: I don't think so, just following by what was in the previous `docker.py` code

atomicapp/providers/docker.py
vpavlin: Ah, I was not aware you need to get a name. I though it's just about killing the container..

It does not support regex, but is it a problem?

```
re.match("%s_+%s+_+[a-zA-Z0-9]{12} % (self.default_name, x), container)]
```

You can match for content of %s_%s_ from above - is it necessary to match also for the UUID? (I mean, it's probably a valid issue on docker to support regular expressions, but can't we for now ignore the uuid and use it just as a tool to have unique names?)

atomicapp/utils.py
vpavlin: I've probably screwed up in expressing myself. I want you to replace my code with this new method so that we have a single way of generation of unique container names. Does it make sense now?

charliedrage: @vpavlin added :+1:

atomicapp/providers/docker.py
```diff
@@ -51,15 +60,41 @@ def init(self):
...
 
     def deploy(self):
+        logger.info("Deploying to provider: Docker")
+
         for artifact in self.artifacts:
             artifact_path = os.path.join(self.path, artifact)
             label_run = None
             with open(artifact_path, "r") as fp:
                 label_run = fp.read().strip()
+            labels = label_run.split()
```
vpavlin: I am sorry that I haven't realized it earlier, but could you rename `labels` to something more matching what it is? Like `run_args` f.e.?

atomicapp/providers/docker.py
```diff
@@ -51,15 +60,41 @@ def init(self):
...
+
+    def undeploy(self):
+        logger.info("Undeploying to provider: Docker")
+
+        docker_cmd = 'docker inspect --format="{{ .Name }}" $(sudo docker ps -aq --no-trunc) | sed "s,/,,g"'
+        containers = dict((line, 1) for line in subprocess.check_output(docker_cmd, shell=True).splitlines())
+
+        # Regex checks for matching container name and kills it. ex. atomic_mariadb-atomicapp-app_9dfb369ed2a0
+        for container in containers:
+            m = re.match("%s_+%s+_+[a-zA-Z0-9]{12}" % (self.default_name, self.namespace), container)
```
vpavlin: I am missing a code where `self.default_name` will get the value of `--name` if it's in the run command. I thought we agreed that we can do the undeployment by parsing this name from run command and using it in both deploy and undeploy (?), but maybe I misunderstood:-)

vpavlin:
LGTM

charliedrage:
@vpavlin @goern rebased. good for merge?

goern:
LOKTM, I guess testing is not really possible? Could you write up some docs how the docker provider handles container naming and why it does so.

charliedrage:
docs added, they :+1: @bexelbie ?

@goern Some CLI testing, yes. I will be adding that with a separate PR later this week.

dustymabe:
Did this take care of #182 ? Should it be closed?

charliedrage:
@dustymabe haha, yup. #182 should be closed now

PR #267, opened 2015-09-10
Add `providerconfig` option to fix #266

### LINKED ISSUE #266: Allow passing kubeconfig to kubernetes provider
Real world kuberenetes is configured with authentication, multi-master, and all sorts of pixie dust. The kubernetes provider should accept a kubeconfig parameter in the ```answers.conf``` to which callers can pass a kubeconfig format file (ie: ```~/.kube/config```) so that kubectl running under atomicapp can find and authenticate against kubernetes.

vpavlin:
There are 2 new methods in `plugin.py`

1. getConfigFile() - which simple looks up the option in `config` dictionary, called automatically by all providers, if there is no such option, default (`~/.kube/config`) is used
2. checkConfigFile() - which checks for the accessibility of the config file, needs to be called explicitly in provider's `init()` function so that providers which do no require configuration (docker) don't fail

### DIFF
--- a/atomicapp/constants.py
+++ b/atomicapp/constants.py
@@ -47,3 +47,5 @@ DEFAULT_ANSWERS = {
         "namespace": DEFAULT_NAMESPACE
     }
 }
+PROVIDER_CONFIG_KEY="providerconfig"
+DEFAULT_PROVIDER_CONFIG="~/.kube/config"
--- a/atomicapp/constants.py
+++ b/atomicapp/constants.py
@@ -47,5 +47,5 @@ DEFAULT_ANSWERS = {
         "namespace": DEFAULT_NAMESPACE
     }
 }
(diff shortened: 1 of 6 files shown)


vpavlin:
One thing I am not sure about is whether we should add this option to `DEFAULT_ANSWERS` in constants or not. It might be useful, but also confusing for providers which do not require configs

vpavlin:
And one more note - this will break many examples on the blog posts as no kubes example will probably  "just work™" because it will miss the note about changing the provider. We could make `checkConfigFile` a bit more permissive with adding an argument like `can_fail = False` and set it to `True` in Kubernetes

vpavlin:
Added unit test for the `providerconfig` option

vpavlin:
Closes #75

Issue #265, opened 2015-09-10
[PROPOSAL] implement docker-compose provider


goern:
source: [Container-tools] Minutes from Tools Cabal Call Sep 8

   Ratnadeep suggestst to add Docker-compose, which should be simple, and
   also document the process so that others can work on adding providers

   Need to clearly articulate how do do so, that's probably more important
   than adding Docker-compose itself.
   Is the issue about the provider APIs closed ? Should probably go first.
    - Aaron issues about how install and run is to be modified should not
      affect the API
  Langdon had a first Compose provider but it was never merged

  A bunch of challenges with kubernetes were mentioned - can we get more
  detail there?
  - It was hit as part of doing the Sentry packaging it's not broken, it's
    more a difference between how kubernete works and how Docker works

rtnpro:
Implementing docker-compose should be pretty straigt forward. I see following steps involved:
- Render docker compose YAML file
- Run the rendered docker compose YAML file using docker-compose command
- Using the same docker compose YAML file, stop the services.

As mentioned above, this is more of a reference implementation of a provider, so that we can document the process to create a new provider. This will make it easy for people from the community to come up and add their own providers.

Also, some thoughts need to be put, to outline the scope of a provider, so that we can expect a base minimum set of features from all providers.

Issue #145, opened 2015-06-04
Rework how install and run work internally


vpavlin:
__This is a post-Summit problem__

Given an example of wordpress application:
```
wordpress-atomicapp
  ├── mariadb-atomicapp
  └── skydns-atomicapp
```

When `install` is called, atomicapp installs `wordpress-atomicapp` and, loads Nulecule and goes through `graph`. When it finds external dependency it recursively starts installation for this dependency with a destination path `external/<dep-app-id>`. Once the installation of dependency is finished, recursion ends and the `graph` cycle goes on.

Example:
```
atomicapp install wordpress-atomicapp
FOUND mariadb-atomicapp: 
    atomicapp install --destination external/mariadb-atomicapp mariadb-atomicapp
FOUND skydns-atomicapp:
    atomicapp install --destination external/skydns-atomicapp skydns-atomicapp
```

It works the same for `run`.

## Problems
* Once install is finished for a dependency, atomicapp cannot access it's data easily
* We cannot easily inspect what a dependency provides (f.e. providers) and thus cannot give a user list of available providers or guess default provider
* It would be better if `provider.deploy()` happened from one "cycle" not from a recursively called instance cause that makes it hard to properly react if something fails.

## Solution
Implement a Python class which will represent a Nulecule application (`nulecule_base.py` is sort of where to start) and
* Create object for the main app
* Create objects for all deps and reference them from main app object
* Rework install and run so that once all objects are created and filled with data, install/run gathers them to a single list and performs needed actions.

Comments?:)

vpavlin:
This is needed for proper `docker-compose` provider

rtnpro:
Fixed in https://github.com/projectatomic/atomicapp/tree/cdk2-beta3-rc1

dustymabe:
This work has been merged to master in 90a1e94

PR #268, opened 2015-09-11
Added docker-compose provider #265

rtnpro:
This implements support for ``docker-compose`` as a provider. This work is not yet ready to merge.

Docker compose YAML file is almost analogous to a Nulecule spec file. We don't need to have different compose artifacts for individual application. In this case, how do we accommodate for docker compose in Nulecule file? It seems that artifact for each application for an enabled provider is a must.

### DIFF
--- a/atomicapp/providers/docker_compose.py
+++ b/atomicapp/providers/docker_compose.py
@@ -43,7 +43,7 @@ class DockerComposeProvider(Provider):
             cmd = "docker-compose -f {} up -d".format(artifact_path)
 
             if self.dryrun:
-                logger.info("DRY-RUN: {}".format(cmd))
+                logger.info("DRY-RUN: %s" % cmd)
             else:
                 subprocess.check_call(cmd.split(" "))
 
@@ -58,6 +58,6 @@ class DockerComposeProvider(Provider):
             cmd = "docker-compose -f {} stop".format(artifact_path)
 
             if self.dryrun:
-                logger.info("DRY-RUN: {}".format(cmd))
+                logger.info("DRY-RUN: %s" % cmd)
             else:
                 subprocess.check_call(cmd.split(" "))
--- a/atomicapp/providers/docker_compose.py
+++ b/atomicapp/providers/docker_compose.py
@@ -44,7 +44,7 @@ class DockerComposeProvider(Provider):
             cmd = "docker-compose -f {} up -d".format(artifact_path)
 
             if self.dryrun:
-                logger.info("DRY-RUN: %s" % cmd)
+                logger.info("DRY-RUN: %s", cmd)
             else:
                 subprocess.check_call(cmd.split(" "))
 
@@ -58,7 +58,7 @@ class DockerComposeProvider(Provider):
             cmd = "docker-compose -f {} stop".format(artifact_path)
 
             if self.dryrun:
-                logger.info("DRY-RUN: %s" % cmd)
+                logger.info("DRY-RUN: %s", cmd)
             else:
                 subprocess.check_call(cmd.split(" "))
 
@@ -68,7 +68,7 @@ class DockerComposeProvider(Provider):
                 data = anymarkup.parse(fp)
                 if not data:
                     logger.info(
-                        'Skipping empty compose YAML file: %s' % artifact_path)
+                        'Skipping empty compose YAML file: %s', artifact_path)
                     return True
             except Exception:
                 msg = "Error processing %s artifacts, Error: %s" % (
--- a/atomicapp/providers/docker_compose.py
+++ b/atomicapp/providers/docker_compose.py
@@ -17,9 +17,11 @@
  along with Atomic App. If not, see <http://www.gnu.org/licenses/>.
 """
 
-from atomicapp.plugin import Provider
+import anymarkup
 import os
 import subprocess
+from atomicapp.plugin import Provider
+from atomicapp.utils import printErrorStatus
 
 import logging
 
@@ -36,8 +38,7 @@ class DockerComposeProvider(Provider):
         for artifact in self.artifacts:
             artifact_path = os.path.join(self.path, artifact)
 
-            if artifact_path.find('noop-compose') >= 0:
-                logger.info('Skipping noop compose YAML file.')
+            if self._is_artifact_empty(artifact_path):
                 continue
 
             cmd = "docker-compose -f {} up -d".format(artifact_path)
@@ -51,8 +52,7 @@ class DockerComposeProvider(Provider):
         for artifact in self.artifacts:
             artifact_path = os.path.join(self.path, artifact)
 
-            if artifact_path.find('noop-compose') >= 0:
-                logger.info('Skipping noop compose YAML file.')
+            if self._is_artifact_empty(artifact_path):
                 continue
 
             cmd = "docker-compose -f {} stop".format(artifact_path)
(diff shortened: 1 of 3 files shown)


atomicapp/providers/docker_compose.py
```diff
@@ -0,0 +1,64 @@
...
+    def undeploy(self):
+        for artifact in self.artifacts:
+            artifact_path = os.path.join(self.path, artifact)
+
+            if artifact_path.find('noop-compose') >= 0:
+                logger.info('Skipping noop compose YAML file.')
+                continue
+
+            cmd = "docker-compose -f {} stop".format(artifact_path)
+            logger.info("DRY-RUN: %s" % cmd)
```
vpavlin: We don't want DRY-RUN be mentioned here, I guess:)

rtnpro: aye :+1:

atomicapp/providers/docker_compose.py
```diff
@@ -0,0 +1,64 @@
...
+    key = "docker-compose"
+
+    def init(self):
+        pass
+
+    def deploy(self):
+        for artifact in self.artifacts:
+            artifact_path = os.path.join(self.path, artifact)
+
+            if artifact_path.find('noop-compose') >= 0:
```
vpavlin: Rather than depend on a specific name, I'd like to see a test where you load the file  as yaml (use anymarkup) and if it provides empty dictionary, you skip the subprocess call.

atomicapp/providers/docker_compose.py
rtnpro: @vpavlin Fixed

atomicapp/providers/docker_compose.py
rtnpro: Makes sense. Fixing it!

rtnpro: @vpavlin Fixed!

atomicapp/providers/docker_compose.py
```diff
@@ -0,0 +1,76 @@
...
+
+            if self.dryrun:
+                logger.info("DRY-RUN: %s" % cmd)
+            else:
+                subprocess.check_call(cmd.split(" "))
+
+    def undeploy(self):
+        for artifact in self.artifacts:
+            artifact_path = os.path.join(self.path, artifact)
+
```
vpavlin: Same chunk of code for skipping artifacts is needed here..

atomicapp/providers/docker_compose.py
```diff
@@ -0,0 +1,76 @@
...
+    def deploy(self):
+        for artifact in self.artifacts:
+            artifact_path = os.path.join(self.path, artifact)
+
+            with open(artifact_path, 'r') as fp:
+                try:
+                    data = anymarkup.parse(fp)
+                    if not data:
+                        continue
```
vpavlin: I'd still print a warning - similar to the one on line 53

rtnpro: ok

atomicapp/providers/docker_compose.py
rtnpro: ah! yes

vpavlin:
@rtnpro One more request:-) Please, mention issue #145 in `providers.md` in compose description to specifically point out this is not the final version of the provider (WRT not being able to really compose the compose file, but having to specify the whole file in a top-level app)

Otherwise looks good.

goern:
@rtnpro could you check the landscape messages and increase (not decrease) the code health ;) thanks!

rtnpro:
@vpavlin @dustymabe It's ready to be merged.

dustymabe:
@rtnpro let's hold off until after we have a baseline for beta3. does that seem reasonable?

please ping us on this after that time.

rtnpro:
@dustymabe ^^ :)

dustymabe:
so issue #145 is closed now. Is there a reason that this is still not the "final implementation of the provider" as mentioned in c4ae394?

dustymabe:
@rtnpro ^^

atomicapp/providers/docker_compose.py
cdrage: @rtnpro update and let's merge this into master in time for next week's release :+1:

rtnpro:
@dustymabe the issues are not that related, and even with the current atomicapp implementation, we cannot cleanly compose a docker compose file. So, I will update the docs.

atomicapp/providers/docker_compose.py
```diff
@@ -0,0 +1,77 @@
...
+    def _is_artifact_empty(self, artifact_path):
+        with open(artifact_path, 'r') as fp:
+            try:
+                data = anymarkup.parse(fp)
+                if not data:
+                    logger.info(
+                        'Skipping empty compose YAML file: %s', artifact_path)
+                    return True
+            except Exception:
+                msg = "Error processing %s artifacts, Error: %s" % (
```
dustymabe: 2 %s but only one arg provided?

docs/providers.md
```diff
@@ -86,6 +88,24 @@ This command undeploys the app in Docker. For the given namespace, the undeploy:
...
+The Docker Compose provider: ``docker-compose``
+will start a set of docker containers for a multi-tier application as
+specified in a Docker Compose YAML file on the host the Atomic App is
+deployed. Since, a docker compose YAML file will represent
+the entire graph of applications in a Nulecule file, you don't need to run
+docker compose on each application in the Nulecule graph. However, since
+each provider for an application must point to an artifact file, you can
+point to an empty ``artifacts/docker-compose/noop-compose.yaml`` artifact
+file for applications you want to skip running docker compose on.
+
```
dustymabe: I guess I need a crash course in this to understand. Can you give me an example of why we would need empty yaml files to make this work?

atomicapp/providers/docker_compose.py
rtnpro: Fixed!

docs/providers.md
rtnpro: Because, artifacts/source are a must for a graph item. And, an entry in ``artifacts`` must point to a valid file. In case of compose, a single file contains instructions to deploy a multi container application.

rtnpro:
@cdrage rebased and made the suggested fixes! :+1: from my side. Let me know, if anything else needs to be done.

atomicapp/providers/docker_compose.py
```diff
@@ -0,0 +1,77 @@
...
+        pass
+
+    def deploy(self):
+        for artifact in self.artifacts:
+            artifact_path = os.path.join(self.path, artifact)
+
+            if self._is_artifact_empty(artifact_path):
+                continue
+
+            cmd = "docker-compose -f {} up -d".format(artifact_path)
```
dustymabe: Is there a docker-compose binary that exists in our container image? I don't think so. Won't this fail when running inside the container?

dustymabe:
@rtnpro considering that we want to use the provider's APIs in the future, does that change anything with regards to this feature? Right now it uses docker-compose command, is it even possible to use an API for this?

dustymabe:
We had a discussion with several team members today and we are re-evaluating if docker-compose even makes sense as a provider to us. We will postpone merge of this until after tomorrows release.

cdrage:
After discussion with both @dustymabe and @rtnpro in regards to the Docker compose provider we have concluded that at the moment, the provider is not a suitable inclusion for Atomic App.

We will be closing this PR, however, we will re-evaluate the provider again in the future and it's role in Atomic App.

PR #277, opened 2015-09-17
Convert global params in Nulecule from list to dict #276

### LINKED ISSUE #276: Convert global params in Nulecule to dict from list during loading params
## Problem
Although the current code base, theoretically, allows handling ``params`` at the root level of a Nulecule file, we do not convert global params list to dict before calling:
https://github.com/projectatomic/atomicapp/blob/master/atomicapp/nulecule_base.py#L193.

As a result call to ``Utils.update`` breaks.

## Solution

Converting global ``params`` to ``dict`` using ``nulecule_base.Nulecule.fromListToDict`` will fix this bug.

rtnpro:
Fixes #276


vpavlin:
LGTM

Issue #142, opened 2015-06-03
Don't unpack files to /atomicapp during run


aweiteka:
Based on user feedback, it is unexpected that a bunch of files show up in current working directory when running an atomic app. This includes visible files as well as the hidden `.workdir/` directory. It *is* expected during install. I propose we use a temporary directory like `/tmp` for these things and only write files during install.

aweiteka:
Related to #136

vpavlin:
Seems like we are reverting the decision we made about the `.workdir` some time ago:) Anyway - I think we write only the templated artifacts on run - the rest is written on install. And we cannot prepare artifacts on install as user might want to change answers.

Writing those artifacts in `/tmp is problematic, because the `/tmp` lives in the container thus we would need to mount some directory there..

nzwulfin:
Based on #136 I did some more digging on containerized atomicapps (the nulecule examples in particular) and came up with a "namespaced" solution with LABELs.

```
LABEL STOP docker run -it --rm --privileged --net=host -v `pwd`/Nulecule:/atomicapp -v /run:/run -v /:/host --name NAME -e NAME=NAME -e IMAGE=IMAGE IMAGE -v stop IMAGE
LABEL RUN  docker run -it --rm --privileged --net=host -v `pwd`/Nulecule:/atomicapp -v /run:/run -v /:/host --name NAME -e NAME=NAME -e IMAGE=IMAGE IMAGE -v run IMAGE
LABEL INSTALL docker run --rm -it --privileged -v /run:/run -v `pwd`/Nulecule:/atomicapp -e IMAGE=IMAGE -e NAME=NAME --name NAME IMAGE -v install --destination IMAGE /application-entity
```

This block added to helloapache will create a Nulecule/nzwulfin/helloapache structure in the current working directory.  This feels better than have the Dockerfile just drop files in PWD, but could rely just on the atomicapp functionality of `run IMAGE` to create the target directory for the app.

For some reason that I can't quite pin down, I *have* to use install before run with this label set.  If this is interesting to other folks, the branch I'm working on is here (https://github.com/nzwulfin/nulecule/tree/nulecule-work-dir)

nzwulfin:
I've looked at a few more methods to encapsulate an app.  What I've run into is that every method of encapsulation aside from creating some sort of working directory on the host runs into the issue @vpavlin brings up about answer files.  

[...]

PR #278, opened 2015-09-18
Refactor nulecule base [Fixes #145]

### LINKED ISSUE #145: Rework how install and run work internally

rtnpro:
This includes initial work on refactoring **Nulecule** base code, as discussed in #145 .


atomicapp/nulecule/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+    def install(self):
+        raise NotImplementedError
+
+    def uninstall(self):
+        raise NotImplementedError
+
+
+class Nulecule(NuleculeBase):
+    """
+    This represents a Nulecule application. A Nulecule instance can have
```
goern: just this morning a little picky... "an Application compliant with the Nulecule Specification" ;)

atomicapp/nulecule/__init__.py
```diff
@@ -0,0 +1,74 @@
...
+        pass
+
+    def uninstall(self):
+        # uninstall the Nulecule application
+        pass
+
+
+class NuleculeComponent(NuleculeBase):
+    """
+    Represents a local service in a Nulecule application. It receives props
```
goern: What do you mean by 'service'?

rtnpro: ``service`` == ``graph item``, or dependency, etc

goern: why dont we call 'a component' rather than 'local service' ?

rtnpro: Agreed :+1:

vpavlin:
Hmm..you are quick, guys:) I put this together yesterday: https://gist.github.com/vpavlin/f46444efd647e0d8a48f

Maybe I missed some discussion, so if that's the case, can we reiterate on it? Or if that didn't happen, can I have one, please?:)

atomicapp/nulecule/__init__.py
```diff
@@ -0,0 +1,116 @@
...
+    def __init__(self, id, specversion, metadata, graph, requirements=None,
+                 params=None, config=None, namespace='general'):
+        self.id = id
+        self.specversion = specversion
+        self.metadata = metadata
+        self.graph = graph
+        self.requirements = requirements
+        self.params = params or []
+        self.namespace = namespace
+        self.config = self.load_config(config, self.params, self.namespace)
```
vpavlin: What about moving this to `load()`, or moving  `load_components()` out of `load` ?

aweiteka:
I haven't done a full code review but I agree this is in the right direction. It will help with my work in progress #142 .

@vpavlin do you consider this foundational to address [your gist](https://gist.github.com/vpavlin/f46444efd647e0d8a48f)?

vpavlin:
@aweiteka Yes, I think this is going into the right direction:)

charliedrage:
:+1: on the docker-py, if we decide to use that we can interact with the docker commands more programatically.

although this does add another dependancy...

@goern thoughts?

aweiteka:
@rtnpro nice work! LGTM

charliedrage:
@rtnpro 

Few questions,
* Is there any possibility to add some testing?
* Is it bad naming practice to have both a NuleculeBase and Nulecule_Base? As that's what we currently have

atomicapp/nulecule/__init__.py
```diff
@@ -0,0 +1,335 @@
...
+
+    def __init__(self, image):
+        self.image = image
+        self.initialize()
+
+    def initialize(self):
+        self.app_name = '{}-{}'.format(
+            Utils.sanitizeName(self.image), uuid.uuid1())
+        self.nulecule = None
+        self.unpack_path = os.path.join(CACHE_DIR, self.app_name)
```
charliedrage: What is /var/atomicapp doesn't exist? fails?

charliedrage:
@rtnpro can you rebase too so I can git clone your branch and test it :)

rtnpro:
@charliedrage I just rebased the branch :)

rtnpro:
On Wed, Oct 7, 2015 at 6:20 PM, Charlie Drage <notifications@github.com> wrote:
>
> @rtnpro
>
> Few questions,
>
> Is there any possibility to add some testing?

Yes, of course, and I will start to work on it, to wire up this new
package and cover all edge cases.
>
> Is it bad naming practice to have both a NuleculeBase and Nulecule_Base? As that's what we currently have

It is, indeed. Python suggests use of `CapWords` for class names and
`_` (underscore) separated names for variable and function names for
better readability. You can read more about it here:
https://www.python.org/dev/peps/pep-0008/#function-names. However, in
our current code, we have a mix of various styles. I'm more inclined
to stick to the PEP8 guidelines, and it reflects in the newer code of
``nulecule`` package. In time, we'll refactor the entire code base to
follow a consistent naming convention. Does it make sense?

rtnpro:
@charliedrage ^^

LalatenduMohanty:
@rtnpro Code is more readable now :)

atomicapp/constants.py
```diff
@@ -29,6 +29,7 @@
 EXTERNAL_APP_DIR = "external"
 GLOBAL_CONF = "general"
 APP_ENT_PATH = "application-entity"
+CACHE_DIR = "/var/atomicapp"
```
vpavlin: Make it `/var/lib/atomicapp`, please:)

atomicapp/nulecule/container.py
```diff
@@ -0,0 +1,54 @@
...
+        subprocess.call(rm_cmd)
+
+    def is_image_present(self, image):
+        output = subprocess.check_output([self.docker_cli, 'images'])
+        image_lines = output.strip().splitlines()[1:]
+        for line in image_lines:
+            words = line.split()
+            image_name = words[0]
+            registry = repo = None
+            if image_name.find('/') >= 0:
```
vpavlin: Maybe use `if "/" in image_name"?

atomicapp/nulecule/container.py
```diff
@@ -0,0 +1,54 @@
...
+                registry, repo = image_name.split('/', 1)
```
dustymabe: what happens here if there is no registry for an image? For example if the image was built on the local machine.

atomicapp/nulecule/lib.py
```diff
@@ -0,0 +1,64 @@
...
+            if config.get(self.namespace) is None:
+                config[self.namespace] = {}
+            config[self.namespace][param['name']] = value
+        self.config = config
+
+    def merge_config(self, to_config, from_config):
+        for group, group_vars in from_config.items():
+            to_config[group] = to_config.get(group) or {}
+            if group == GLOBAL_CONF:
+                if self.namespace == GLOBAL_CONF:
```
vpavlin: This `if` is a dead code, isn't it?

If `self.namespace` is `GLOBAL_CONF`, use `GLOBAL_CONF` (== `self.namespace`), otherwise, use `self.namespace`...i.e. it's `self.namespace` everytime..

atomicapp/nulecule/container.py
vpavlin: see next line - `image_name == image`

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+    This represents an application compliant with Nulecule specification.
+    A Nulecule instance can have instances of Nulecule and Nulecule as
+    components. A Nulecule instance knows everything about itself and its
+    componenents, but does not have access to it's parent's scope.
+    """
+    def __init__(self, id, specversion, metadata, graph, basepath,
+                 requirements=None, params=None, config=None,
+                 namespace=GLOBAL_CONF):
+        self.id = id
```
vpavlin: These should be declared in `NuleculeBase` as methods there are using them and it's not cleare where they come from..(and then obviously set here)

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+    def run(self, provider_key=None, dry=False):
+        provider_key, provider = self.get_provider(provider_key, dry)
+        for component in self.components:
+            component.run(provider_key, dry)
+
+    def stop(self, provider_key=None):
+        provider_key, provider = self.get_provider(provider_key)
+        # stop the Nulecule application
+        for component in self.components:
+            component.stop(provider_key)
```
vpavlin: stop is missing `dry`

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+    def __init__(self, image, unpack_path=None):
+        self.image = image
+        self.initialize(unpack_path)
+
+    def initialize(self, unpack_path=None):
+        if unpack_path:
+            self.unpack_path = unpack_path
+        else:
+            self.app_name = '{}-{}'.format(
+                Utils.sanitizeName(self.image), uuid.uuid1())
```
dustymabe: so this gives us a unique path for every time we execute, right? 

Is there value in adding the uuid here vs just adding it directly to unpack_path below?

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def install(self):
+        self.unpack()
+        self.nulecule.install()
+
+    def run(self, provider_key=None, dry=False):
+        self.install()
+        self.nulecule.run(provider_key, dry)
+
+    def stop(self, provider_key=None):
```
charliedrage: stop missing dry here too

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+            self._app.run(provider_key, dry)
+            return
+        provider_key, provider = self.get_provider(provider_key, dry)
+        provider.artifacts = self.get_artifact_paths_for_provider(
+            provider_key)
+        provider.init()
+        if not dry:
+            provider.deploy()
+
+    def stop(self, provider_key=None):
```
charliedrage: dry ^^

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def run(self, provider_key, dry=False):
+        if self._app:
+            self._app.run(provider_key, dry)
+            return
+        provider_key, provider = self.get_provider(provider_key, dry)
+        provider.artifacts = self.get_artifact_paths_for_provider(
+            provider_key)
+        provider.init()
+        if not dry:
```
vpavlin: Always do `deploy`...provider methods should take care of dry run itself..dry run should print out commands that would be run if it was real run..

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+        self.uninstall()
+        shutil.rmtree(self.unpack_path)
+        self.initialize()
+
+
+class Nulecule(NuleculeBase):
+    """
+    This represents an application compliant with Nulecule specification.
+    A Nulecule instance can have instances of Nulecule and Nulecule as
+    components. A Nulecule instance knows everything about itself and its
```
dustymabe: `A Nulecule instance can have instances of Nulecule and Nulecule as components.` seems a bit of an awkward sentence to me. Can we rephrase to make it more clear?

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def load_external_application(self, dest):
+        nulecule = Nulecule.unpack(
+            self.source,
+            os.path.join(self.basepath, EXTERNAL_APP_DIR, self.name),
+            namespace=self.namespace)
+        self._app = nulecule
+
+    @property
+    def components(self):
```
vpavlin: Shouldn't it return `None` otherwise?

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+    def get_artifact_paths_for_provider(self, provider_key):
+        artifact_paths = []
+        artifacts = self.artifacts.get(provider_key)
+        for artifact in artifacts:
+            if not isinstance(artifact, basestring):
+                continue
+            if artifact.startswith('file:///'):
+                path = artifact[7:]
+            elif artifact.startswith('file://'):
+                path = os.path.join(self.basepath,
```
vpavlin: Too many `self.basepath` joins maybe?

atomicapp/nulecule/lib.py
```diff
@@ -0,0 +1,64 @@
...
+plugin = Plugin()
+plugin.load_plugins()
+
+
+class NuleculeBase(object):
+    def load(self):
+        pass
+
+    def load_config(self, config={}):
+        config = config or {}
```
dustymabe: what is the point of this? If the user doesn't provide a value for config won't the default {} be used?

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+                                    os.path.join(self.basepath, artifact[7:]))
+            else:
```
vpavlin: I am missing `inherit` here (see https://github.com/projectatomic/nulecule/tree/master/spec#artifacts-example)

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+        return artifact_paths
+
+    def render_artifact(self, path, context):
+        with open(path, 'r') as f:
+            content = f.read()
+            template = Template(content)
+            rendered_content = template.safe_substitute(context)
+
+        with open(path, 'w') as f:
+            f.write(rendered_content)
```
vpavlin: This is effectively rewriting the original artifact and thus prevents user from modifying `answers.conf` and running again with different params. We are using `.workdir` in cwd currently. That is not correct, obviously, but we need to output the templated artifact somewhere else then to original file.

atomicapp/nulecule/lib.py
vpavlin: My guess it, that we would get into trouble if user provides `None` as value and this avoids it..

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+        # uninstall the Nulecule application
+        for component in self.components:
+            component.uninstall()
+
+    def load_config(self, config={}):
+        super(Nulecule, self).load_config(config=config)
+        for component in self.components:
+            _config = {}
+            _config[GLOBAL_CONF] = copy.deepcopy(
+                self.config.get(GLOBAL_CONF)) or {}
```
dustymabe: bug here `copy.deepcopy(self.config.get(GLOBAL_CONF)) or {}` is not what you want. `copy.deepcopy(self.config.get(GLOBAL_CONF) or {}))` is what you want I think

atomicapp/nulecule/container.py
rtnpro: :+1:

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+            if self.namespace != GLOBAL_CONF:
+                _config[GLOBAL_CONF].update(
+                    self.config.get(self.namespace) or {})
+            _config[component.name] = copy.deepcopy(
+                self.config.get(component.name)) or {}
```
dustymabe: same problem here.

atomicapp/nulecule/main.py
vpavlin: The result will be the same

atomicapp/nulecule/lib.py
rtnpro: Yeah! You got it :)

atomicapp/nulecule/main.py
rtnpro: Agreed :+1:

atomicapp/nulecule/main.py
vpavlin: Seems like following the workflow we have right now..The reason I implemented it like this beforeis that as we used PWD, we need to check the content of PWD before we copy files there to make sure it's the same app (although maybe in a newer version).

May be worth to keep it this way and check the `specversion` before we populate the "real" location of the app

atomicapp/nulecule/main.py
rtnpro: I need to port it.

atomicapp/nulecule/main.py
rtnpro: Agreed!

atomicapp/nulecule/main.py
rtnpro: :+1:

atomicapp/nulecule/container.py
```diff
@@ -0,0 +1,54 @@
...
+            if image_name == image or repo == image:
+                return True
```
dustymabe: If only 'repo' is matched (meaning the user did not provide registry in their image name) should we print a warning? Is it valuable to push towards always including the registry in the image specification?

vpavlin: I don't think so. Any reasoning behind this idea? Why should we print a warning if user builds&runs some examples/apps locally?

atomicapp/nulecule/main.py
dustymabe: Indeed. you are right.

atomicapp/nulecule/container.py
dustymabe: Let me think.. I think there are a few cases here we need to consider.. Consider the following possibilities:
- docker.io/dustymabe/centos-etherpad-mariadb-atomicapp - in docker registry
- gcr.io/dustymabe/centos-etherpad-mariadb-atomicapp - in gcr registry
- dustymabe/centos-etherpad-mariadb-atomicapp - built locally

If I just say `dustymabe/centos-etherpad-mariadb-atomicapp` for my image name then if any of the three *images* listed above exist on my system it will pass the test. I don't know if that would be considered a bad thing or not. If we don't specify a registry then we can get ourselves in this situation so I was asking if maybe we should print a warning?

vpavlin: Makes sense! That's actually why previous versions of Atomic App had a code to prepend the image name with registry based on `answers.conf`...though we agreed we will leave this to user's docker settings..so maybe it's time to bring it back:-) 

Problem of requiring the registry if that...

1. MyApp lives in some public registry (pblc.io) and has pblc.io/another-app as dependency
2. To lower the network load, I will pull these nulecules to my local company registry and this will have mycompany.io/MyApp and mycompany.io/another-app...
3. I do `atomic run mycompany.io/MyApp`, but it fails as MyApp still requires pblc.io/another-app

atomicapp/nulecule/main.py
charliedrage: We should keep the same UUID conventions as creating the images && installing images to the localhost.

use `utils.py` getUniqueUUID()

atomicapp/nulecule/container.py
dustymabe: we'll leave it like it is for now and perhaps bring it back up once we have everything merged.

atomicapp/nulecule/container.py
```diff
@@ -0,0 +1,54 @@
...
+class DockerHandler(object):
+    """Interface to interact with Docker."""
+
+    def __init__(self, docker_cli='/usr/bin/docker'):
+        self.docker_cli = docker_cli
+
+    def pull(self, image, update=False):
+        if not self.is_image_present(image) or update:
+            logger.debug('Pulling Docker image: %s' % image)
+            pull_cmd = [self.docker_cli, 'pull', image]
```
dustymabe: At some point we need to pull the `docker://` part off of the image names that are in the Nulecule like is done currently in https://github.com/projectatomic/atomicapp/blob/master/atomicapp/utils.py#L138 

I don't know if I like having  `docker://` prepended. I suspect we will need to change that at some point in the future.. Should we put a XXX comment in there to revisit this?

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+    def unpack(cls, image, path, config={}, namespace=GLOBAL_CONF):
+        docker_handler = DockerHandler()
+        docker_handler.pull(image)
+        docker_handler.extract(image, APP_ENT_PATH, path)
+        nulecule_data = anymarkup.parse_file(
+            os.path.join(path, 'Nulecule'))
+        nulecule = Nulecule(config=config, basepath=path, namespace=namespace,
+                            **nulecule_data)
+        nulecule.load_components()
+        return nulecule
```
dustymabe: At some point we need to make it extremely obvious to the user what directory we are unpacking to. A print statement here would be a first start.. we might also want to consider printing out a very human readable message right before we return that lets them know as well.

dustymabe:
A couple of things I have noticed:
- I don't think the sample answers file is getting populated when it should.
- in the future we need to somehow "detect" when a previous step has already been executed and omit certain steps based on that.

dustymabe:
To get stop to work I needed to modify the code a bit:

```
diff --git a/atomicapp/nulecule/main.py b/atomicapp/nulecule/main.py
index acca17d..614cb7a 100644
--- a/atomicapp/nulecule/main.py
+++ b/atomicapp/nulecule/main.py
@@ -153,7 +153,7 @@ class NuleculeComponent(NuleculeBase):
         self.basepath = basepath
         self.source = source
         self.params = params or []
-        self.artifacts = artifacts
+        self.artifacts = artifacts or {}
         self._app = None
 
     def load(self):
@@ -176,6 +176,8 @@ class NuleculeComponent(NuleculeBase):
         if self._app:
             self._app.stop(provider_key)
         provider_key, provider = self.get_provider(provider_key)
+        provider.artifacts = self.get_artifact_paths_for_provider(
+            provider_key)
         provider.init()
         provider.undeploy()
 
@@ -209,7 +211,7 @@ class NuleculeComponent(NuleculeBase):
 
     def get_artifact_paths_for_provider(self, provider_key):
         artifact_paths = []
-        artifacts = self.artifacts.get(provider_key)
+        artifacts = self.artifacts.get(provider_key, [])
         for artifact in artifacts:
             if not isinstance(artifact, basestring):
                 continue
```

atomicapp/nulecule/main.py
```diff
@@ -0,0 +1,234 @@
...
+
+    def run(self, provider_key=None, dry=False):
+        self.install()
+        self.nulecule.run(provider_key, dry)
+
+    def stop(self, provider_key=None):
+        self.nulecule.stop(provider_key)
+
+    def uninstall(self):
+        self.stop()
```
dustymabe: Need to be able to pass provider_key to uninstall as well.

atomicapp/nulecule/container.py
vpavlin: I don't completely agree here..I saw a CloudFoundry talk at LinuxCon and they use the same notation. Yes, we can get rid of prefix, but we need to be able to distinguish between formats anyway (as Docker doesn't have to be necessary the only backend Nulecule/Atomic App will support in the future)

dustymabe: @vpavlin I agree that we need to be able to distinguish between the formats (appc, oci, etc.. could all be options). The *we will need to change that at some point in the future* part was to indicate that we need to stop hardcoding docker and start supporting these other image formats.

dustymabe:
Remaining issues that I see:
- No sample answers file gets created on "install"
- Code doesn't properly chop off `docker://` in image specifications within nulecule files
- If I do just a `run` with the image name as the arg the app doesn't get unpacked.. I get an error
- The expected format of the artifacts within `providers/docker.py` is different than what we are populating when we set `provider.artifacts = self.rendered_artifacts.get(provider_key, [])`. Getting a lot of errors because it can't find the artifacts.
- Somehow once I get the other things fixed atomicapp tries to deploy things twice (i.e two docker containers for mariadb) which causes problems. 

These are a few things I notice.. We'll iterate on them as we go.

atomicapp/nulecule/main.py
rtnpro: Makes sense :+1:

rtnpro: Done!

atomicapp/nulecule/container.py
rtnpro: Done!

atomicapp/constants.py
rtnpro: Done!

atomicapp/nulecule/container.py
rtnpro: :+1:

atomicapp/nulecule/container.py
rtnpro: Done!

rtnpro:
TODOs:
- write ``answers.conf.sample`` during install step
- check for duplicate containers getting run on running ethrepad application
- Do not copy directories inside ``load_from_path``
- Inline comments around complicated logic

charliedrage:
@rtnpro Can you rebase to master again now that todo is done? I'm going to look at testing and see what I can add.

We've got to flake8 this and then start adding tests / making sure they pass and there's no regresssions.

dustymabe:
TODO:

- [ ] `install` shouldn't query the user for answers. It should make a sample answers.conf file with `None` in place of the values the user needs to populate. This is an example of one that is created: http://ur1.ca/o0trz

dustymabe:
@aweiteka @goern @vpavlin do you mind taking a look at this PR now? I think we are close to merge. Keep in mind that it doesn't have to be perfect. We just need something *good enough* so that others of us can start building on top of what @rtnpro has built. 

if we choose not to merge until it is perfect then we need to make a topic branch others can contribute to because it is becoming a bottleneck.

rtnpro:
@dustymabe 

All the TODOs above except the 3rd one has been fixed. I defer with you in the 3rd item in the TODO list:
``remove config=self.answers from call to self.nulecule = self.unpack() in nulecule/main.py:82``

The install option takes ``--answers`` data, so, why not consider the supplied ``answers`` data when generating ``answers.conf.sample``?

dustymabe:
@rtnpro your code from before didn't accept a `config` option to `unpack()` so I got this error: 
```
TypeError: unpack() got an unexpected keyword argument 'config'
```

You fixed this in 1640020 though so no need to fix it now :)

dustymabe:
OK. I think most everything looks good. LGTM.

rtnpro:
\o/

rtnpro:
Rebased!

dustymabe:
Ok this has been incorporated in https://github.com/projectatomic/atomicapp/commits/cdk2-beta3-rc1 I'm going to lock comments here.. new things can go into issues/PRs to https://github.com/projectatomic/atomicapp/commits/cdk2-beta3-rc1.

Issue #134, opened 2015-05-28
refactor logging


goern:
Let's refactor out logging, so that logging will append to console and to context.globalCtx. REST API will provide a new /log showing the log.

charliedrage:
@goern no more REST api due to cockpit integration correct?

vpavlin:
REST API is still in plan, but I don't think it's high prio right now.

PR #288, opened 2015-09-23
[WIP] [Do not merge] Beautify logging, add more CLI commands

charliedrage:
Beautifies the logging by using a class to display the output to console. Outputs to stderr and stdout consoles

(ignore my terminal motd's)

Turns this:
old

Into this:
new

Need some input on:
 * Colour(s)
 * Do we touch / create an atomicapp.log on creation or remove logging module completely (since we output to err and info on console anyways)? SEE TOP OF `display.py` 
 * Is there ANY WAY to remove the console __Cockpit__ requirements? (pass a --output=cockpit param?)

Need some testing on:
 * Cockpit (atomicapp.status.info.message) to see if there is any regressions. If anyone has a current dev environment with Cockpit + atomicapp

Fixes: #134

### DIFF
--- a/atomicapp/constants.py
+++ b/atomicapp/constants.py
@@ -37,6 +37,7 @@ ANSWERS_FILE_SAMPLE = "answers.conf.sample"
 ANSWERS_FILE_SAMPLE_FORMAT = 'ini'
 WORKDIR = ".workdir"
 LOCK_FILE = "/run/lock/atomicapp.lock"
+LOG_FILE = "/var/log/atomicapp.log"
 HOST_DIR = "/host"
 
 DEFAULT_PROVIDER = "kubernetes"
--- a/atomicapp/run.py
+++ b/atomicapp/run.py
@@ -21,18 +21,15 @@ from __future__ import print_function
 import os
 from string import Template
 
-import logging
-
 from nulecule_base import Nulecule_Base
 from utils import Utils, printStatus, printErrorStatus
 from constants import GLOBAL_CONF, DEFAULT_PROVIDER, MAIN_FILE, ANSWERS_FILE_SAMPLE_FORMAT
 from plugin import Plugin, ProviderFailedException
 from install import Install
+from display import Display
 
-logger = logging.getLogger(__name__)
 
-
-class Run(object):
+class Run:
     debug = False
     dryrun = False
     nulecule_base = None
@@ -55,6 +52,7 @@ class Run(object):
             answers_format=ANSWERS_FILE_SAMPLE_FORMAT, **kwargs):
 
         self.debug = debug
+        self.display = Display()
         self.dryrun = dryrun
         self.stop = stop
         self.kwargs = kwargs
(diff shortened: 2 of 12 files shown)


atomicapp/constants.py
```diff
@@ -37,6 +37,7 @@
 ANSWERS_FILE_SAMPLE_FORMAT = 'ini'
 WORKDIR = ".workdir"
 LOCK_FILE = "/run/lock/atomicapp.lock"
+LOG_FILE = "/var/log/atomicapp.log"
```
goern: That /var/log inside the container? outside? Or.. `atomic` somehow manages it? `needs_doc`

atomicapp/run.py
```diff
@@ -55,6 +52,7 @@ def __init__(
             answers_format=ANSWERS_FILE_SAMPLE_FORMAT, **kwargs):
 
         self.debug = debug
+        self.display = Display()
```
vpavlin: Why is `display` global var in one case and class attribute in another? I'd rather like to see this unified.

atomicapp/constants.py
vpavlin: I'd suggest using `Utils.getRoot()` https://github.com/projectatomic/atomicapp/blob/master/atomicapp/utils.py#L213 and log to host if it's available.

And maybe condition the file logging with some argument (`--persistent-log`, `--log-file`...?)

goern:
If you do colorful logging, please include a --no-colors command line argument. Additionally, have a look at https://www.patternfly.org/styles/color-palette/ our UX guys have put a lot of work into defining a color schema.

atomicapp/run.py
charliedrage: I'll unify this.

For some reason, if you have display() initialized from within the `main.py/install.py/run.py` classes, it grabs `logging.getLogger("atomicapp").isEnabledFor(logging.DEBUG)` correctly, but if it's outside of the class, it doesn't. For the life of me I couldn't fix it.

Once we have the unified base class from #145 we can just grab from the base class whether self.verbose is set or not.

aweiteka:
@charliedrage much improved.

1. what are the cockpit requirements? where are they defined?
1. could you remove all header information from stdout? IMO this should provide basic info to the user about what is going on. It gives them confidence there is something happening, status, etc. So I would recommend fairly minimal output.

charliedrage:
@goern @vpavlin 
So... adding log arguments would be great and I agree on your suggestions :+1: 

My thoughts:
--log-driver = [cockpit, systemd, stdout, awesomealternativeapi, etc.]
* This will determine where, which will be handled by `display.py`. Will help removing header info from logging. This will introduce a break in Cockpit versioning / Nulecule.. but this is a __much__ needed break, we need to beautify this logging for a user's perspective.

--log-opt color=none max-width=30 etc...

vpavlin:
Nice, adding a systemd driver i.e. logging to journal is a great idea:)

Yeah, I am probably fine with all of that:-)

charliedrage:
**update**

I'm going to close this for now and come back to this when #288 is completed so I can call to the base class in order to get a consitent variable whether something is verbose or not, instead of calling `logger`.

PR #305, opened 2015-10-06
Added Makefile

charliedrage:
Added Makefile, creates some sweet aliases. Satisfies the developer OCD.

Makes everything a little bit easier for local dev / install and for future installations / testing. More aliases can be added here. HINT HINT: `make dev` to Vagrant possibly? :)

```
COMMAND
ALIAS
```

```
make 
(default) python -m pytest -vv
```

````
make install
python setup.py install .
```

```
make test
python -m pytest -vv
```

```
make tag=atomicapp image
docker build -t atomicapp image
```

```
make syntax-check
flake8 atomicapp
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,14 @@
+all:
+	python -m pytest -vv
+
+install:
+	python setup.py install .
+
+test:
+	python -m pytest -vv
+
+image:
+	docker build -t $(tag) .
+
+syntax-check:
+	flake8 atomicapp


LalatenduMohanty:
@charliedrage I like the idea :) . LGTM

charliedrage:
@vpavlin @goern @dustymabe 

need one more opinion before merge :) want to get this in so we can add it more easily to a testing unit test bash script

vpavlin:
hmm..I would probably expect the default for `make`  to be a `make image` with some predefined `$tag`..but that's just me:)

Otherwise LGTM

goern:
hehe... needs_doc :)

PR #307, opened 2015-10-08
Reverts "artifact logic bugfix". Fixes 306

dustymabe:
This reverts commit 931740f, which causes the user to get prompted
for answers during an 'install' operation.


charliedrage:
@dustymabe 
Just tested and it fixes it. *doh* on my previous PR. 

LGTM :+1: 

I will write some more testing on this so there are no regressions in the future.

PR #314, opened 2015-10-13
Updating travis.yml script

charliedrage:
Cleans up the TravisCI script.

* Added Makefile aliases
* Change organization of before and after install
* Installs atomicapp completely instead of using `python -m pytest --cov atomicapp` then runs tests.

ping @goern for :+1: since he setup the previous travis script :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,26 @@
+# Travis CI checks for atomicapp
+
 language: python
+
 python:
   - "2.7"
+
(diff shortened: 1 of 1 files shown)


goern:
LGTM, even though I dont see the added clearness :P

charliedrage:
@goern seperates a few things so it doesn't unexpected fail during `install` when it should of been `before_install` :P

PR #318, opened 2015-10-14
Docker terminology fix

charliedrage:
Fix to when docker provider stops to say `STOPPING CONTAINER` rather than `REMOVING CONTAINER`.


aweiteka:
ACK

PR #322, opened 2015-10-15
Corrects the exit codes output by the cli.

dustymabe:
In shell 0 is success and non-zero is failure. True in python translates to a failure in shell:

```
[dustymabe@media ~]$ python -c 'import sys; sys.exit(True)' || echo "FAIL"
FAIL
```


rtnpro:
:+1:  Good to merge.

charliedrage:
:+1: LGTM

PR #324, opened 2015-10-15
Fix answers-format ordering

charliedrage:
Pushes answers-format back to the beginning of the CLI (rather than after the `install` or `run` commands). 

Needed to prevent cockpit integrattion break. 

https://github.com/cockpit-project/cockpit/blob/54a9405a90d7df48e8f2c4d060cf50e33537abf3/pkg/kubernetes/nulecule.js

^^ see line 241


vpavlin:
LGTM

dustymabe:
LGTM

Issue #321, opened 2015-10-15
OpenShift Integration support


aweiteka:
To make atomicapp work on OpenShift we need to change the way remote atomic apps are pulled in. This primarily means eliminating calls to docker, replaced by a call `oc new-app`. See high-level proposal image.

![](https://docs.google.com/drawings/d/13mfTkxv_M3jM6WMsgpJtoKkX4nuVuTPcPSfsebN3qKA/pub?w=884&h=320)

aweiteka:
Checklist:

- [ ] auto-detect openshift provider #346 
- [x] add LABELs #334 
- [ ] run as non-root (scope ASAP--openshift cannot run a container as root)
- [ ] enable client commands inside atomicapp container #336

aweiteka:
This work is to enable "phase 0"/PoC. Limitations:

- no support for remote apps or cascading nulecules
- no support for passing in params

We need to scope phase 1 to address these items for the MVP.

aweiteka:
Basic openshift `oc new-app <atomicapp_image>` is in origin 1.0.7.

aweiteka:
@dustymabe the remaining items are these from above in priority order.

1. enable client commands inside atomicapp container via [API](https://docs.openshift.org/latest/rest_api/openshift_v1.html)
1. remote apps need to be called with `oc new-app <remote_nulecle_image>`, not docker.

dustymabe:
This efforts here are largely accomplished in the open pull request #454.

There are a few items that re left un-settled:

- Support for nesting of nulecule applications (can't yet point to another nulecule)
- Support for passing in an answers.conf file

dustymabe:
Closing this issue. The remaining items are:

- Support for nesting of nulecule applications: https://github.com/projectatomic/atomicapp/issues/461
- Support for passing in an answers.conf file: https://github.com/projectatomic/atomicapp/issues/462

dustymabe:
The bulk of this work was done in #454.

PR #334, opened 2015-10-16
Add openshift LABELs

aweiteka:
In order for the OpenShift provider to natively
support atomicapp we need to pass in a flag that
identifies the image as an openshift self-executing
job. We also have to instruct OpenShift to pass
in user auth token so we can run oc as that user.

cc @smarterclayton to confirm LABEL correct key=value

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,8 @@ FROM centos:centos7
 MAINTAINER Red Hat, Inc. <container-tools@redhat.com>
 
 LABEL io.projectatomic.nulecule.atomicappversion="0.1.11" \
+      io.openshift.generate.job=true \
+      io.openshift.generate.token.as=env:TOKEN_ENV_VAR \
       RUN="docker run -it --rm \${OPT1} --privileged -v `pwd`:/atomicapp -v /run:/run -v /:/host --net=host --name \${NAME} -e NAME=\${NAME} -e IMAGE=\${IMAGE} \${IMAGE} -v \${OPT2} run \${OPT3} /atomicapp" \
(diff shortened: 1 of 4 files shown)


dustymabe:
A couple of thoughts here. I'd actually like the labels to not be specific to openshift in case other providers would like to benefit from these labels in a similar way. Perhaps we can transform them into something like:

```
io.openshift.generate.job=true --> io.projectatomic.nulecule=true
io.openshift.generate.token.as=env:TOKEN_ENV_VAR --> io.projectatomic.nulecule.authtoken=env:AUTHTOKEN_ENV_VAR
```

My new label proposals may not make sense but maybe we can come up with something that isn't openshift specific that does? @smarterclayton @aweiteka what do you think?

smarterclayton:
The point of the labels being specific to openshift is that they are only used in one context - the OpenShift new-app generator.  In the future, making these labels generic is great - but for that, there needs to be use of them.  We can always add more labels, but I'm generally not in support of making generic things and hoping people use them.  I want to make concrete things and prove their utility.  I would prefer these to be openshift specific for now.

dustymabe:
Fair enough. I just don't want to deal with migration issues in the future if we decide to move to a more generic labeling scheme. I guess since we have a heavy influence in the openshift community it would be easier to change there than if we were working with some community where we didn't have influence.

goern:
On the other hand, a LABEL is just a label, no one will hinder some 3rd party platform/provider/... to evaluate a label called `io.openshift...`

goern:
doc and test (plan) are missing, otherwise LGTM

aweiteka:
@goern docs added. Test plan will be part of the code change PR that addresses #321 that enables openshift-native running of atomic app.

smarterclayton:
Initial code is here https://github.com/openshift/origin/pull/5378

Issue #172, opened 2015-07-14
PROPOSAL: run provider client as another container


aweiteka:
Currently we rely on the host having the provider installed as installed software. There is a move to run these CLI clients as containers. Since we don't want to ship all providers in the atomic app itself it would be ideal to execute the provider call as an atomic run call. This requires two things:

1. A library of supported providers is available to call.
1. Update the providers to call atomic run <provider_image> from item no.1 above.

whitel:
+1

vpavlin:
@aweiteka by library you meen a list of images in Docker Hub which represent specific providers?

aweiteka:
@vpavlin yes, that was the thought. A set of images in `<some_registry>`. I think it's just a matter of time before clients (openshift, heat, kube, etc) will be containerized by default.

goern:
so we wait a little before client containers do appear and in the meantime we spec out how that could be used with atomicapp?

kadel:
Wouldn't it be better to use providers API directly (if they have one) instead of relying on CLI clients?

aweiteka:
Yes, I think that would be a valid approach. The biggest issue might be tracking the API version.

I don't think the containerized client would be too hard to implement. If we provide a config key to specify the client image we could simply run the container. We would need to consider how provider artifacts are passed into the container.

aweiteka:
@kadel I think you're right about the API in the near-term, especially for openshift since we currently don't have a model to run "sidecar" images in pods. I closed #336.

dustymabe:
If we are moving to using the providers APIs rather than using the CLI then we should be able to close this: https://github.com/projectatomic/atomicapp/issues/407 

For openshift provider I think we have PR already: https://github.com/projectatomic/atomicapp/pull/420

@aweiteka should we close this?

PR #336, opened 2015-10-16
[Do Not Merge] Add oc client to image

aweiteka:
The OpenShift oc client will soon run atomicapp
natively. However the current design requires
the oc CLI client be inside the atomicapp image.
The package is 93M. This is considered a temporary
workaround until another design is proposed.

### DIFF
--- a/Dockerfile.rhel7
+++ b/Dockerfile.rhel7
@@ -20,7 +20,11 @@ RUN yum install -y yum-utils --setopt=tsflags=nodocs && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable epel && \
     yum install -y --setopt=tsflags=nodocs python-pip docker && \
-    yum clean all
+    yum clean all && \
+    curl -L https://github.com/openshift/origin/releases/download/v1.0.2/openshift-origin-v1.0.2-325c7b7-linux-amd64.tar.gz | tar -zx && \
+    mv oc /usr/local/bin && \
+    rm oadm && \
+    rm openshift
 
 WORKDIR /opt/atomicapp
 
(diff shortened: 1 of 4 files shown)


Dockerfile.rhel7
```diff
@@ -18,7 +18,11 @@ RUN yum install -y yum-utils --setopt=tsflags=nodocs && \
     yum-config-manager --enable rhel-server-rhscl-7-rpms && \
     yum-config-manager --enable epel && \
     yum install -y --setopt=tsflags=nodocs python-pip docker && \
-    yum clean all
+    yum clean all && \
+    curl -L https://github.com/openshift/origin/releases/download/v1.0.2/openshift-origin-v1.0.2-325c7b7-linux-amd64.tar.gz | tar -zx && \
```
goern: using origin code on a rhel7? as a tmp workaround that seems to be ok...

goern:
and why is it that we cant require the host to have oc installed? as we do with docker or kubectl?

Dockerfile.rhel7
aweiteka: Right, installing from upstream source is not ideal. An oc RPM pkg is coming soon, I believe.

aweiteka:
@goern regarding oc client on host, see image in #321. The issue is OpenShift needs to blindly run the container since it's tagged as an "executable job" (see #334). The job container run unprivileged.

dustymabe:
Adding [Do Not Merge] to title so that this doesn't get merged before beta3.

aweiteka:
I think @kadel 's suggestion to use the API (#172) is the right way to go. We're only making a couple calls anyway. I'm going to close this. It's easy enough to reopen if we prove otherwise.

PR #330, opened 2015-10-16
Cdk2 beta3 rc1 unittests

rtnpro:
This branch adds unittests for ``atomicapp/nulecule``. This is still **WORK IN PROGRESS**.

PR #338, opened 2015-10-19
remove failing/unfinished test files

charliedrage:
cleaning this up, unfinished testing code was added when pulling refactor PR into branch. tests will be re-added in #330


charliedrage:
need :+1: from @rtnpro in-case he needs to save this data.

PR #340, opened 2015-10-19
Fix dry-run not stopping

charliedrage:
Fixes issue #339.

Dry-run isn't passed to stop

Found fix via `python -m trace -t` and some logging

PR #342, opened 2015-10-19
fix gitlab test

charliedrage:
As per the refactor PR by @rtnpro, we are no longer using .workdir.

Dotfiles of the original artifacts are created in the providers folder when installing / running / stopping. Thus `.workdir` is no longer needed / used for tests.

This PR fixes the gitlab example. Instead of checking the .workdir folder, it checks all the artifacts in the kubernetes folder.

### DIFF
--- a/tests/units/cli/test_cli_gitlab_example.py
+++ b/tests/units/cli/test_cli_gitlab_example.py
@@ -26,7 +26,6 @@ import unittest
 import pytest
 
 import atomicapp.cli.main
-from atomicapp.constants import WORKDIR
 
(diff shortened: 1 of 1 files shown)


charliedrage:
__NOTE__: tests/units/cli/test_cli_gitlab_example.py::TestGitLabCli::test_stop_gitlab_app is going to fail until #340 is fixed :)

dustymabe:
Seems to work for me. LGTM

rtnpro:
:+1: @charliedrage

PR #345, opened 2015-10-20
Don't unpack files to cwd.

### LINKED ISSUE #142: Don't unpack files to /atomicapp during run

dustymabe:
This should fix #142.


atomicapp/cli/main.py
```diff
@@ -37,10 +37,19 @@
...
+
 
 def cli_install(args):
     try:
-        NuleculeManager.do_install(**vars(args))
+        argdict = args.__dict__
+        nm = NuleculeManager(app_spec=argdict['app_spec'],
+                             destination=argdict['destination'])
+        nm.install(**argdict)
+        print(APP_MESSAGE % nm.app_path)  # msg for users
```
charliedrage: should this be logging.info?

atomicapp/cli/main.py
```diff
@@ -52,7 +61,11 @@ def cli_install(args):
 
 def cli_run(args):
     try:
-        NuleculeManager.do_run(**vars(args))
+        argdict = args.__dict__
+        nm = NuleculeManager(app_spec=argdict['app_spec'],
+                             destination=argdict['destination'])
+        nm.run(**argdict)
+        print(APP_MESSAGE % nm.app_path)  # msg for users
```
charliedrage: should this be logging.info?

charliedrage:
need rebase with @rtnpro 's docs merge, then I can code review more && test

atomicapp/cli/main.py
dustymabe: I decided specifically to not use logging.info mainly because the output is so verbose that people tend to ignore it (at least I do). Either that or it becomes a chore to sift through it to find something useful. 

I made a decision to spit this out as the last thing that gets output and separate from the other lines so that people would read it and say "Oh, ok.. that's where my application files are". I think what we'll find is that it is much better for user experience.

atomicapp/cli/main.py
```diff
@@ -188,12 +209,12 @@ def set_arguments(self):
...
         parser_install.add_argument(
             "--destination",
-            dest="target_path",
+            dest="destination",
             default=None,
             help="Destination directory for install")
 
         parser_install.add_argument(
-            "APP",
+            "app_spec",
```
dustymabe: @rtnpro @charliedrage let me know if you can come up with a better variable name than `app_spec`

atomicapp/cli/main.py
charliedrage: Agreed.

Perhaps when I push the logging PR that I _was_ working on there prompts / info will be a bit more __bold__.

rtnpro:
:+1: LGTM

charliedrage:
LGTM :+1: Merge away and I'll add some testing to it too.

atomicapp/cli/main.py
rtnpro: @dustymabe ok :+1:

atomicapp/cli/main.py
rtnpro: what about ``app_src``?

dustymabe: this PR is already merged.. we can update to app_src in new PR in the future if you want.

PR #346, opened 2015-10-20
Auto-detect OpenShift provider

aweiteka:
With the primary use of OpenShift we are
not able to specify the provider, which
makes having an arbitrary default kubernetes provider
problematic. This moves in the direction
of not requiring a default provider but
detecting what the environment capabilities
are. This is preferable to an arbitrary default
provider.

Addresses #321

### DIFF
--- a/atomicapp/nulecule/lib.py
+++ b/atomicapp/nulecule/lib.py
@@ -3,6 +3,8 @@ from atomicapp.constants import (GLOBAL_CONF, DEFAULT_PROVIDER,
                                  DEFAULT_ANSWERS)
 from atomicapp.utils import Utils
 from atomicapp.plugin import Plugin
+import os
+import subprocess
 
 plugin = Plugin()
 plugin.load_plugins()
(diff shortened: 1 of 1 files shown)


atomicapp/nulecule/lib.py
```diff
@@ -98,12 +100,32 @@ def get_provider(self, provider_key=None, dry=False):
...
+    @property
+    def is_openshift(self):
+        """
+        The KUBERNETES_SERVICE_HOST env var should only exist
+        on an openshift or kubernetes environment.
+        Here we check if the "kubernetes" host has an openshift
+        API endpoint. If so, we're calling openshift.
+        """
+        if os.getenv("KUBERNETES_SERVICE_HOST"):
+            curl_cmd = ("curl -ks -w '%{http_code}' "
```
dustymabe: why not use urllib right here instead of shelling out to curl?

aweiteka: I agree that would be better. I was concerned about adding another python dependency.

charliedrage:
To be frank, I do not believe this to be a good approach for detecting default providers.

What would happen if I am running both Kubernetes and Openshift in containers on my system? (on two different API end-points), or if I'm running both Docker and Openshift and I only want to deploy to Docker? There's a lot that can go wrong.

I believe that if we were to incorporate finding / setting a default provider based upon what is on the underlying system, it should not be a hard-setting and the user should be able to override it based upon providing `--provider`. 

A better approach as well would be to add a new function to each provider (for ex. `def detect`). When the Nulecule base code loads all available plugins it could use this function to detect what is running on the system. If it detects both, warn the user, etc.

dustymabe:
@charliedrage I think I had the same thought you did when I first looked at this. I thought it was trying to detect based on the host system and make a decision. This isn't the case though, as really what it is doing is saying "am I already running in an openshift pod?, if so then I'm going to default to talking to openshift".

charliedrage:
@dustymabe OHHHHHHHHHHHHHH, now that makes sense.

dustymabe:
I say use urllib and then we could merge.

aweiteka:
@dustymabe that's exactly the behavior. I'll implement using urllib. thanks.

aweiteka:
This may "just work" but it's hard to test until we can run this on OpenShift. (I need to think through how we might do this.) Please review and comment where this might go release-wise. See #321 for how this fits into the "big picture".

dustymabe:
Added this functionality in 0b4be42 and added code to act on it in 470beff.

PR #352, opened 2015-10-22
Fixes broken builds bc of lockfile dependency.

dustymabe:
The new version of lockfile in pypi requires pip itself which we
remove in the Dockerfile. Let's use the older version for now.
See [1] for a more complete explanation.

[1] - https://www.redhat.com/archives/container-tools/2015-October/msg00102.html

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 anymarkup>=0.4.1
-lockfile
+lockfile==0.10.2


charliedrage:
LGTM :+1: 

Do we need a quick :+1: from @vpavlin before merge?

dustymabe:
I don't think he cares.. but @vpavlin ^^

PR #353, opened 2015-10-22
 Support artifact directory path in NuleculeComponent.

### LINKED ISSUE #143: Support kubernetes provider artifacts path
Kubernetes is able to deploy all of the artifact files if given a path. For example:

```
$ ls -1 artifacts/kubernetes/
pod.yaml
service.yaml
$ kubectl create -f artifacts/kubernetes/
pod/mypod
service/myservice
```

I would expect the atomic app provider to work the same way. This will also reduce developer pain when listing all of the correct paths to files. The template can call `- file://artifacts/kubernetes/` and if all the files are placed there no additional work needs to be done.

rtnpro:
Fixes #143 

This adds support for handling directory paths in artifacts list in a Nulecule ``graph``. It does actually check if the pathname endswith '/' or not to check if the path is a directory, rather, it uses ``os.path.isdir`` to check if the path is a directory or not.

### DIFF
--- a/atomicapp/nulecule/base.py
+++ b/atomicapp/nulecule/base.py
@@ -362,7 +362,7 @@ class NuleculeComponent(NuleculeBase):
                 path = Utils.sanitizePath(artifact)
                 path = os.path.join(self.basepath, path) \
                     if path[0] != '/' else path
-                artifact_paths.append(path)
+                artifact_paths.extend(self._get_artifact_paths_for_path(path))
             elif isinstance(artifact, dict) and artifact.get('inherit') and \
                     isinstance(artifact.get('inherit'), list):
                 for inherited_provider_key in artifact.get('inherit'):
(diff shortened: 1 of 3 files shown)


atomicapp/nulecule/base.py
```diff
@@ -403,3 +403,24 @@ def render_artifact(self, path, context):
...
+            path (str): Local path
+
+        Returns:
+            list: A list of artifact paths
+        """
+        artifact_paths = []
+        if os.path.isfile(path):
+            artifact_paths.append(path)
+        elif os.path.isdir(path):
+            for dir_child in os.listdir(path):
```
navidshaikh: The `path` given by user is a path of directory having a set of k8s manifests in it. Do we need to `os.listdir` and traverse in sub directories as well?
@aweiteka : am I missing something here?

rtnpro: You got a point. We'll wait for @aweiteka to comment.

aweiteka: I don't think we need to worry about subdirs at this point.

aweiteka:
@rtnpro could you add a note about this behavior to this doc? https://github.com/projectatomic/atomicapp/blob/master/docs/file_handling.md

atomicapp/nulecule/base.py
rtnpro: Cool :)

dustymabe: Do you mind adding a note that says we don't deal with nested artifacts at this time?

dustymabe: Otherwise everything LGTM

rtnpro: @dustymabe Added notes :+1: and updated docs on ``file_handling`` as suggested by @aweiteka .

rtnpro:
@dustymabe GTM from my side :+1:

PR #359, opened 2015-10-26
Remove dotfiles created by tests

charliedrage:
Currently a bunch of dotfiles are created when running `make test`. This PR makes sure that the files are removed on test tear down.


vpavlin:
LGTM

Issue #193, opened 2015-08-06
file exists error for /etc/kubernetes when running w/ local docker-based kube setup


jasonbrooks:
Running the ["Local Docker-based (recommended starting point)"](https://github.com/GoogleCloudPlatform/kubernetes/blob/release-1.0/docs/getting-started-guides/docker.md) method from the upstream project, using [this Vagrantfile](https://gist.github.com/jasonbrooks/2244af4cf9e411ff2116):

```
$ vagrant up
$ vagrant ssh
$ mkdir wp && cd wp
$ sudo atomic run projectatomic/wordpress-centos7-atomicapp
```
```
2015-08-06 17:45:13,588 - kubernetes - INFO - Using namespace default
Traceback (most recent call last):
  File "/usr/bin/atomicapp", line 9, in <module>
    load_entry_point('atomicapp==0.1', 'console_scripts', 'atomicapp')()
  File "/usr/lib/python2.7/site-packages/atomicapp/cli/main.py", line 89, in main
    cli.run()
  File "/usr/lib/python2.7/site-packages/atomicapp/cli/main.py", line 70, in run
    args.func(args)
  File "/usr/lib/python2.7/site-packages/atomicapp/cli/main.py", line 21, in cli_run
    ae.run()
  File "/usr/lib/python2.7/site-packages/atomicapp/run.py", line 187, in run
    self._dispatchGraph()
  File "/usr/lib/python2.7/site-packages/atomicapp/run.py", line 99, in _dispatchGraph
    self._processComponent(component, graph_item)
  File "/usr/lib/python2.7/site-packages/atomicapp/run.py", line 168, in _processComponent
    provider.init()
  File "/usr/lib/python2.7/site-packages/atomicapp/providers/kubernetes.py", line 29, in init
    os.symlink("/host/etc/kubernetes", "/etc/kubernetes")
OSError: [Errno 17] File exists
```

aweiteka:
We check for `/etc/kubernetes`, but if we're in a container and that file is not found we create `/host/etc/kubernetes`, which is the default kube config dir bindmounted from the host.

@jasonbrooks it doesn't look like you're running from the vagrant box. Do you have an /etc/kubernetes dir on your host or in the vagrant box?

jasonbrooks:
I'm running from a default centos 7 vagrant box, the Vagrantfile is linked above. There's no /etc/kubernetes in there, because I'm running everything in containers, w/ the project's local Docker-based installation method.

[...]

PR #368, opened 2015-10-29
do not create symlink if source doesn't exist

kadel:
create symlink for /etc/kubernetes only if /host/etc/kubernetes exists, so we don't end up with invalid symlink

### DIFF
--- a/atomicapp/providers/kubernetes.py
+++ b/atomicapp/providers/kubernetes.py
@@ -22,7 +22,6 @@ import logging
 import os
 from subprocess import Popen, PIPE
 
-from atomicapp.constants import HOST_DIR
 from atomicapp.plugin import Provider, ProviderFailedException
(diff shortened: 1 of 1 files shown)


vpavlin:
LGTM

dustymabe:
Since this was opened against master branch we'll look at it more closely once the cdk branch is merged.

dustymabe:
hey @kadel. This looks good except for one thing.. can you change the `os.path.exists(kube_conf_path)` to `os.path.lexists`. see the comment for more info: https://github.com/projectatomic/atomicapp/issues/193#issuecomment-153155529

dustymabe:
@kadel actually nevermind. Looks good like it is. If we never create the symlink in the first place (if the target doesn't exist) then there is no reason to use lexists().

dustymabe:
will merge after beta branch is merged tomorrow.

PR #370, opened 2015-10-29
Run as non-root user and provide default command

### LINKED ISSUE #252: Add a `USER` in atomicapp image
If no user is set in image the default user is `root`, IMO we need to configure user other than `root`. I propose we use user `atomicapp`.

aweiteka:
resolves #252 
ref: #321


dustymabe:
will need some work after beta3 branch is merged I think. Let's chat after that happens (tomorrow we think).

dustymabe:
Closing. A different approach was taken to enable the functionality we needed for native openshift support. This was done in #454.

PR #376, opened 2015-11-04
Sync up labels and make run label work again

dustymabe:


charliedrage:
LGTM :+1: tested and works against
```
atomic --version
1.6

~
```

PR #381, opened 2015-11-09
Fetch stderr response from kubectl/oc command

sub-mod:
@dustymabe kindly review
Cockpit needs stderr value from kubectl and oc commands

### DIFF
--- a/atomicapp/utils.py
+++ b/atomicapp/utils.py
@@ -30,7 +30,7 @@ import uuid
 from distutils.spawn import find_executable
 
 import logging
-
+from subprocess import Popen, PIPE
 from constants import (ANSWERS_FILE,
                        APP_ENT_PATH,
                        CACHE_DIR,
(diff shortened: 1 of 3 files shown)


atomicapp/utils.py
```diff
@@ -197,6 +197,36 @@ def sanitizePath(path):
...
+            p = Popen(cmd, stdin=PIPE, stdout=PIPE, stderr=PIPE)
+            stdout, stderr = p.communicate(args_str)
+            logger.debug("stdout = %s", stdout)
+            logger.debug("stderr = %s", stderr)
+            """
+            stderr can be None.
+            We need to raise an exception for the stderr value
+            if the response is from Kubernetes or Openshift.
+            """
+            if stderr and stderr.strip() != "":
```
dustymabe: what if a process gives a bad exit code but doesn't have any stderr output? 

what if a process has stderr output but doesn't fail?

atomicapp/utils.py
```diff
@@ -197,6 +197,36 @@ def sanitizePath(path):
             return path[7:]
 
     @staticmethod
+    def run_cmd(cmd, args_str=None):
+        """
+        Runs a command with its arguments.
+
+        Args:
+            args_str (str): arguments to the command.
```
dustymabe: This is weird because args_str this is actually being passed as `stdin` to the command and not actually as args. Since nothing is actually using it I would remove it.

atomicapp/utils.py
dustymabe: I would modify to only raise Exception if the exit code is bad.

atomicapp/utils.py
```diff
@@ -197,6 +197,36 @@ def sanitizePath(path):
             return path[7:]
 
     @staticmethod
+    def run_cmd(cmd, args_str=None):
```
dustymabe: add arg that allows for specifying if we should check the exit code or not. ignore_exit_code=False or something like that.

dustymabe:
closing in favor of #418

PR #388, opened 2015-11-10
Merge config sets value for key if it or it's value is missing at source

### LINKED ISSUE #385: answers.conf.gen not getting all of the answers in it
@dustymabe reported this, need to reproduce it.

rtnpro:
Fixes #385

# Issue
When merging config data, we're initially checking whether a ``key`` exists in a section or not. If the ``key`` is missing in the section, then only we set it's value from data coming from child nulecule components. This failed in cases, like:
```
[some_section]
key1=None
key2=None
```

# Solution

We now set the value for a key if it is missing in the section, or it's value is None.


dustymabe:
LGTM

PR #392, opened 2015-11-11
Marathon provider

### LINKED ISSUE #357: Mesos provider?
Has anyone had any experience with Mesos know if it's a viable provider compatible with `atomicapp`?

kadel:
fixes #357

### DIFF
--- a/atomicapp/providers/marathon.py
+++ b/atomicapp/providers/marathon.py
@@ -50,6 +50,8 @@ class Marathon(Provider):
         self._process_artifacts()
 
     def deploy(self):
+        """ Deploys the app by given resource manifests.
+        """
         for artifact in self.marathon_artifacts:
             url = urlparse.urljoin(self.marathon_api, "apps/")
 
@@ -71,6 +73,9 @@ class Marathon(Provider):
                 raise ProviderFailedException(msg)
 
     def undeploy(self):
+        """ Undeploys the app by given resource manifests.
+        Undeploy operation deletes Marathon apps from cluster.
+        """
         for artifact in self.marathon_artifacts:
             url = urlparse.urljoin(
                 self.marathon_api,
@@ -95,8 +100,8 @@ class Marathon(Provider):
                 raise ProviderFailedException(msg)
 
     def _process_artifacts(self):
-        """
-        Parse and validate Marathon artifacts and save them to self.marathon_artifacts
+        """ Parse and validate Marathon artifacts
+        Parsed artifacts are saved  to self.marathon_artifacts
         """
         for artifact in self.artifacts:
             logger.debug("Procesesing artifact: %s", artifact)
--- a/atomicapp/providers/marathon.py
+++ b/atomicapp/providers/marathon.py
@@ -62,9 +62,7 @@ class Marathon(Provider):
             try:
                 res = requests.post(url, json=artifact)
             except requests.exceptions.Timeout, e:
-                printErrorStatus(e)
-                # TODO - raise Exception or do sys.exit(1)
-                sys.exit(1)
+                raise ProviderFailedException(e.message)
             if res.status_code == 201:
                 logger.info(
                     "Marathon app %s sucessfully deployed.",
@@ -90,9 +88,7 @@ class Marathon(Provider):
             try:
                 res = requests.post(url, json=artifact)
             except requests.exceptions.Timeout, e:
-                printErrorStatus(e)
-                # TODO - raise Exception or do sys.exit(1)
-                sys.exit(1)
+                raise ProviderFailedException(e.message)
             if res.status_code == 200:
                 logger.info(
                     "Marathon app %s sucessfully deleted.",
--- a/atomicapp/providers/marathon.py
+++ b/atomicapp/providers/marathon.py
@@ -18,12 +18,12 @@
 """
 
 import anymarkup
-import requests
 import urlparse
 import logging
 import os
 from atomicapp.plugin import Provider, ProviderFailedException
 from atomicapp.utils import printErrorStatus
+from atomicapp.utils import Utils
 from atomicapp.constants import PROVIDER_API_KEY
 
 logger = logging.getLogger(__name__)
@@ -58,17 +58,15 @@ class Marathon(Provider):
                 continue
 
             logger.debug("Deploying appid: %s", artifact["id"])
-            try:
-                res = requests.post(url, json=artifact)
-            except requests.exceptions.Timeout, e:
-                raise ProviderFailedException(e.message)
-            if res.status_code == 201:
+            (status_code, return_data) = \
+                Utils.make_rest_request("post", url, data=artifact)
+            if status_code == 201:
                 logger.info(
                     "Marathon app %s sucessfully deployed.",
                     artifact["id"])
             else:
                 msg = "Error deploying app: %s, Marathon API response %s - %s" % (
-                    artifact["id"], res.status_code, res.text)
+                    artifact["id"], status_code, return_data)
                 logger.error(msg)
                 raise ProviderFailedException(msg)
 
@@ -84,17 +82,15 @@ class Marathon(Provider):
                 continue
 
             logger.debug("Deleting appid: %s", artifact["id"])
-            try:
-                res = requests.post(url, json=artifact)
-            except requests.exceptions.Timeout, e:
-                raise ProviderFailedException(e.message)
-            if res.status_code == 200:
+            (status_code, return_data) =  \
+                Utils.make_rest_request("delete", url, data=artifact)
+            if status_code == 200:
                 logger.info(
                     "Marathon app %s sucessfully deleted.",
                     artifact["id"])
             else:
                 msg = "Error deleting app: %s, Marathon API response %s - %s" % (
-                    artifact["id"], res.status_code, res.text)
+                    artifact["id"], status_code, return_data)
                 logger.error(msg)
                 raise ProviderFailedException(msg)
 
--- a/atomicapp/providers/marathon.py
+++ b/atomicapp/providers/marathon.py
@@ -60,9 +60,12 @@ class Marathon(Provider):
             logger.debug("Deploying appid: %s", artifact["id"])
             res = requests.post(url, json=artifact)
             if res.status_code == 201:
-                logger.info("Marathon app %s sucessfully deployed.", artifact["id"])
+                logger.info(
+                    "Marathon app %s sucessfully deployed.",
+                    artifact["id"])
             else:
-                msg = "Error deploying app: %s, Marathon API response %s - %s" % (artifact["id"], res.status_code, res.text)
+                msg = "Error deploying app: %s, Marathon API response %s - %s" % (
+                    artifact["id"], res.status_code, res.text)
                 logger.error(msg)
                 raise ProviderFailedException(msg)
 
@@ -70,7 +73,10 @@ class Marathon(Provider):
         """ Undeploy containers
         """
         for artifact in self.marathon_artifacts:
-            url = urlparse.urljoin(self.marathon_api, "apps/%s" % artifact["id"])
+            url = urlparse.urljoin(
+                self.marathon_api,
+                "apps/%s" %
+                artifact["id"])
 
             if self.dryrun:
                 logger.info("DRY-RUN: %s", url)
@@ -79,9 +85,12 @@ class Marathon(Provider):
             logger.debug("Deleting appid: %s", artifact["id"])
             res = requests.delete(url, json=artifact)
             if res.status_code == 200:
-                logger.info("Marathon app %s sucessfully deleted.", artifact["id"])
+                logger.info(
+                    "Marathon app %s sucessfully deleted.",
+                    artifact["id"])
             else:
-                msg = "Error deleting app: %s, Marathon API response %s - %s" % (artifact["id"], res.status_code, res.text)
+                msg = "Error deleting app: %s, Marathon API response %s - %s" % (
+                    artifact["id"], res.status_code, res.text)
                 logger.error(msg)
                 raise ProviderFailedException(msg)
 
@@ -99,9 +108,11 @@ class Marathon(Provider):
                     # every marathon app has to have id
                     # 'id' key  is also used for showing messages
                     if "id" not in data.keys():
-                        raise ProviderFailedException("Error processing %s artifact. There is no id" % artifact)
+                        raise ProviderFailedException(
+                            "Error processing %s artifact. There is no id" % artifact)
                 except Exception:
-                    msg = "Error processing %s artifcat. Error:" % os.path.join(self.path, artifact)
+                    msg = "Error processing %s artifcat. Error:" % os.path.join(
+                        self.path, artifact)
                     printErrorStatus(msg)
                     raise
                 self.marathon_artifacts.append(data)
--- a/atomicapp/providers/marathon.py
+++ b/atomicapp/providers/marathon.py
@@ -24,6 +24,7 @@ import logging
 import os
 from atomicapp.plugin import Provider, ProviderFailedException
 from atomicapp.utils import printErrorStatus
+from atomicapp.constants import PROVIDER_API_KEY
 
 logger = logging.getLogger(__name__)
 
@@ -42,7 +43,7 @@ class Marathon(Provider):
     def init(self):
         logger.debug("Given config: %s", self.config)
         if self.config.get("providerurl"):
-            self.marathon_api = self.config.get("providerurl")
+            self.marathon_api = self.config.get(PROVIDER_API_KEY)
             self.marathon_api = urlparse.urljoin(self.marathon_api, "v2/")
         logger.debug("marathon_api = %s", self.marathon_api)
         self._process_artifacts()
--- a/atomicapp/providers/marathon.py
+++ b/atomicapp/providers/marathon.py
@@ -111,9 +111,11 @@ class Marathon(Provider):
                     logger.debug("Parsed artifact %s", data)
                     # every marathon app has to have id. 'id' key  is also used for showing messages
                     if "id" not in data.keys():
-                        raise ProviderFailedException(
-                            "Error processing %s artifact. There is no id" % artifact)
+                        msg = "Error processing %s artifact. There is no id" % artifact
+                        printErrorStatus(msg)
+                        raise ProviderFailedException(msg)
                 except anymarkup.AnyMarkupError, e:
                     msg = "Error processing artifact - %s" % e
                     printErrorStatus(msg)
+                    raise ProviderFailedException(msg)
                 self.marathon_artifacts.append(data)
--- a/atomicapp/providers/marathon.py
+++ b/atomicapp/providers/marathon.py
@@ -30,8 +30,6 @@ logger = logging.getLogger(__name__)
 
 
 class Marathon(Provider):
-    """ Marathon provider
-    """
 
     key = "marathon"
     config_file = None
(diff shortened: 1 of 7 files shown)


atomicapp/cli/main.py
```diff
@@ -165,7 +165,7 @@ def set_arguments(self):
         parser_run.add_argument(
             "--provider",
             dest="cli_provider",
-            choices=['docker', 'kubernetes', 'openshift'],
+            choices=['docker', 'kubernetes', 'openshift', 'marathon'],
```
vpavlin: Can we replace these list with a constant rather than having to update them on 2 places (run and stop)

docs/providers.md
```diff
@@ -147,3 +147,37 @@ There are two required labels for OpenShift to run an Atomic App.
...
+This provider requires configuration (`providerurl`) to be able to connect to Marathon API. If no `providerurl` is specified it will use `http://localhost:8080` as
+Marathon API url.
+
+**Configuration values**
+
+Table 1. Marathon default configuration values
+
+Keyword    | Required | Description                                 | Default value
+-----------|----------|---------------------------------------------|--------------------------
+providerurl|   no     |  url for Marathon REST API   | http://localhost:8080
```
vpavlin: What about something like `provider_api_endpoint` ? (I know it's long, but sounds more accurate)

atomicapp/providers/marathon.py
```diff
@@ -0,0 +1,107 @@
...
+            else:
+                msg = "Error deploying app: %s, Marathon API response %s - %s" % (artifact["id"], res.status_code, res.text)
+                logger.error(msg)
+                raise ProviderFailedException(msg)
+
+    def undeploy(self):
+        """ Undeploy containers
+        """
+        for artifact in self.marathon_artifacts:
+            url = urlparse.urljoin(self.marathon_api, "apps/%s" % artifact["id"])
```
vpavlin: Just a nitpicking.. I'd suggest to unify the way you work with URLs - either use string formating everywhere or `urljoin`..just for the sake of consistency

docs/providers.md
kadel: I agree it looks much better. I just wanted to keep it in same style as other options, like `providerconfig` (no word separator)

vpavlin: Right, so maybe `providerapi`? `providerurl` doesn't make sense to me. Let's see what others think:-)

atomicapp/cli/main.py
kadel: Good point, I will do that in another PR.

docs/providers.md
kadel: :+1:  `providerapi` sounds better. If there won't be better suggestion I will use that.

charliedrage:
Awesome work :+1: 

I understand it's still a WIP, but before we merge after GA / Beta-3 I would like to see a few test examples covered in `tests/units/cli/test_cli.py`

atomicapp/providers/marathon.py
rtnpro: :+1:

rtnpro:
:+1: LGTM

charliedrage:
To note: We also changed the way that we pull requirements, so "requests" will need to be pulled from yum/apt-get/epel

requirements.txt
```diff
@@ -1,3 +1,4 @@
 anymarkup==0.5.0
 lockfile==0.10.2
 jsonpointer==1.10.0
+requests==2.8.1
```
charliedrage: To note: We also changed the way that we pull requirements, so "requests" will need to be pulled from yum/apt-get/epel

kadel:
what do you think @cdrage? Is it ready to be merged?

docs/providers.md
```diff
@@ -147,3 +147,37 @@ There are two required labels for OpenShift to run an Atomic App.
...
+
+### Marathon
+
+**Features**
+
+The Marathon provider will deploy an application into Mesos cluster using Marathon
+scheduler.
+
+This provider requires configuration (`providerapi`) to be able to connect to Marathon API. If no `providerapi` is specified it will use `http://localhost:8080` as
+Marathon API url.
```
cdrage: Need docs on this, how do we pass `providerapi` when it's in `constants.py`? We can provide this in `answers.conf` yes?

requirements.txt
```diff
@@ -1,3 +1,4 @@
 anymarkup
 lockfile
 jsonpointer
+requests
```
cdrage: Good that it's added to `requirements.txt`, however, we will need to add this to each `Dockerfile` in regards to pulling in the respective `python-requests` yum/apt-get dependencies. Have a look at how we import anymarkup, lockfile, jsonpointer, etc to our `Dockerfile` :)

atomicapp/providers/marathon.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def deploy(self):
+        """ Deploy containers
+        """
+        for artifact in self.marathon_artifacts:
+            url = urlparse.urljoin(self.marathon_api, "apps/")
+
+            if self.dryrun:
+                logger.info("DRY-RUN: %s", url)
+                continue
```
cdrage: Have a look how we implement this in our Kubernetes and Docker provider, imo this if statement should be at the very end (no `continue`) and show more information (what json would be passed)

atomicapp/providers/marathon.py
```diff
@@ -0,0 +1,122 @@
...
+
+            logger.debug("Deploying appid: %s", artifact["id"])
+            res = requests.post(url, json=artifact)
```
cdrage: should there be a try/rescue here in-case that requests in non-responsive / timeouts / doesn't exist?

atomicapp/providers/marathon.py
```diff
@@ -0,0 +1,122 @@
...
+from atomicapp.plugin import Provider, ProviderFailedException
+from atomicapp.utils import printErrorStatus
+from atomicapp.constants import PROVIDER_API_KEY
+
+logger = logging.getLogger(__name__)
+
+
+class Marathon(Provider):
+    """ Marathon provider
+    """
```
cdrage: no need to have this comment unless we're adding lot's more info

atomicapp/providers/marathon.py
```diff
@@ -0,0 +1,122 @@
...
+    marathon_api_version = "v2"
+    # use localhost as default, when no providerurl is specified
+    marathon_api = "http://localhost:8080/%s/" % marathon_api_version
+    marathon_artifacts = []
+
+    def init(self):
+        self.marathon_artifacts = []
+
+        logger.debug("Given config: %s", self.config)
+        if self.config.get(PROVIDER_API_KEY):
```
cdrage: won't this fail if we run the marathon provider without the providerkey? since `self.config.get` will exit with a cannot get key error?

docs/providers.md
kadel: yes, you have to put it in answers.conf
something like this:
```
providerapi=http://10.1.2.2:8080
```

atomicapp/providers/marathon.py
```diff
@@ -0,0 +1,122 @@
...
+
+class Marathon(Provider):
+    """ Marathon provider
+    """
+
+    key = "marathon"
+    config_file = None
+    marathon_api_version = "v2"
+    # use localhost as default, when no providerurl is specified
+    marathon_api = "http://localhost:8080/%s/" % marathon_api_version
```
cdrage: this isn't even used?

ctrl+f through code and wasn't able to find anything calling marathon_api directly, only self.marathon_api

atomicapp/providers/marathon.py
```diff
@@ -0,0 +1,122 @@
...
+    marathon_artifacts = []
+
+    def init(self):
+        self.marathon_artifacts = []
+
+        logger.debug("Given config: %s", self.config)
+        if self.config.get(PROVIDER_API_KEY):
+            self.marathon_api = self.config.get(PROVIDER_API_KEY)
+            self.marathon_api = urlparse.urljoin(self.marathon_api, "v2/")
+
```
cdrage: imo, else statement to use default from "http://localhost:8080/%s/" % marathon_api_version and set as self.marathon_api

requirements.txt
dustymabe: @cdrage I'm confused. is python-requests not an rpm? It should be brought in just fine with the new way we do request processing in the dockerfiles.

also.. we should evaluate new dependencies pretty heavily from this point on. We should not do anything to break GA and adding a new dep may cause issues with release engineering. @kadel can you verify python-requests exists in CentOS/RHEL/Fedora ?

atomicapp/providers/marathon.py
```diff
@@ -0,0 +1,122 @@
...
+            url = urlparse.urljoin(
+                self.marathon_api,
+                "apps/%s" %
+                artifact["id"])
+
+            if self.dryrun:
+                logger.info("DRY-RUN: %s", url)
+                continue
+
+            logger.debug("Deleting appid: %s", artifact["id"])
```
cdrage: We should _not_ be doing delete at the moment. Currently `undeploy` simply stops the container. Is there the functionality in Mesos to do stop rather than delete?

We will be adding `uninstall` later or something similar that will actually delete the containers on each provider (docker/kubenetes/openshift/etc.)

atomicapp/providers/marathon.py
```diff
@@ -0,0 +1,122 @@
...
+            res = requests.delete(url, json=artifact)
```
cdrage: again here about timeout / 404 / etc. should there be a try / rescue?

atomicapp/providers/marathon.py
```diff
@@ -0,0 +1,122 @@
...
+        """
+        for artifact in self.artifacts:
+            logger.debug("Procesesing artifact: %s", artifact)
+            data = None
+            with open(os.path.join(self.path, artifact), "r") as fp:
+                try:
+                    data = anymarkup.parse(fp)
+                    logger.debug("Parsed artifact %s", data)
+                    # every marathon app has to have id
+                    # 'id' key  is also used for showing messages
```
cdrage: either 1-line comment or we should have this at the top of function

atomicapp/providers/marathon.py
```diff
@@ -0,0 +1,122 @@
...
+                    if "id" not in data.keys():
+                        raise ProviderFailedException(
+                            "Error processing %s artifact. There is no id" % artifact)
+                except Exception:
+                    msg = "Error processing %s artifcat. Error:" % os.path.join(
+                        self.path, artifact)
+                    printErrorStatus(msg)
+                    raise
```
cdrage: raise what? `ProviderFailedException(msg)?`

tests/units/cli/test_examples/helloapache/artifacts/marathon/helloapache.json
```diff
@@ -0,0 +1,20 @@
...
+  "id": "helloapache",
+  "container": {
+    "type": "DOCKER",
+    "docker": {
+      "network": "BRIDGE",
+      "image": "$image",
+      "portMappings": [
+        {
+          "containerPort": 80,
+          "servicePort": $hostport,
```
cdrage: $hostport should be "$hostport" or else it's invalid json.

requirements.txt
cdrage: we still have to do it in `requirements.txt` since our Debian Dockerfile still installs via `requirements.txt` :)

kadel: @cdrage  this should be enough no? Dockerfiles uses this list for generating package names https://github.com/projectatomic/atomicapp/blob/master/Dockerfiles.git/Dockerfile.centos#L22

@dustymabe  it sould be everywhere i check CentOs RHEL and Fedora and it is there.

cdrage: @kadel @dustymabe 
Ah, yeah, that should be good, just tested it under Debian and it works. I was a tad confused for a second.

atomicapp/providers/marathon.py
kadel: Basically next line after this statement is calling Martahon api, and then checking response.
Only thing that can be done to get rid of `continue` is put everything after that to `else` block, but I thing this (using `continue`) is more readable.

atomicapp/providers/marathon.py
kadel: We could scale them down to 0 instances. But in that case redeploy would fail because appid still exists. This is same thing we currently do in kubernetes provider.

This deletes all application definitions in Marathon, Docker containers are stoped, but they are not deleted immediately. Mesos deletes them automatically after some time (configurated inside Mesos)

atomicapp/providers/marathon.py
dharmit: We might want to do it for timeouts.

I think the Dockerfile takes care of "doesn't exist" part. An atomicapp image without requests wouldn't be created. That's what I inferred from [1] & [2]. Is that right?

[1] https://github.com/kadel/atomicapp/blob/master/Dockerfiles.git/Dockerfile.centos#L22
[2] https://github.com/kadel/atomicapp/blob/master/Dockerfiles.git/Dockerfile.debian#L24

atomicapp/providers/marathon.py
dharmit: I think it won't because it's been set earlier by [this](https://github.com/projectatomic/atomicapp/pull/392#discussion-diff-46292700R40).

atomicapp/providers/marathon.py
kadel: I think @cdrage had in mind, what if url we do request to doesn't exist. In that case status_code 404 and error message is shown.

When there is timeout `requests.exceptions.Timeout` is raised. I agree we should catch this.

atomicapp/providers/marathon.py
kadel: @dharmit you are right ;-) Even I missed that and I wrote those lines :-) I need to comment this.

dharmit: @kadel so it also takes care of next two comments from @cdrage, right?

tests/units/cli/test_examples/helloapache/artifacts/marathon/helloapache.json
kadel: it is invalid json only before templating

Marathon expects `servicePort` to be integer, if you do `"$hostport"` then after templating it is string and that is invalid for Marathon

atomicapp/providers/marathon.py
kadel: "empty' raise re-rise previews exeption ;-) I took this from kubernetes provider.

Besides `ProviderFailedException`  only exception that can be raised in that block is `AnyMarkupError`. This could be handled in better way, for sure.

atomicapp/providers/marathon.py
cdrage: yup, it does.

atomicapp/providers/marathon.py
cdrage: @kadel Shouldn't matter, at the moment we should be scaling down to 0 / _stopping_ the application from running, if the user chooses to spin it back up, so be it (this is what we do in the kubernetes provider at the moment). We'll be adding uninstall in the future which would completely remove the application :)

tests/units/cli/test_examples/helloapache/artifacts/marathon/helloapache.json
cdrage: Yeah, you're right. LGTM. Maybe in the future we'll have valid JSON in case someone wants to convert / parse to yaml.

atomicapp/providers/marathon.py
kadel: But in the Kubernetes we are deleting objects (see https://github.com/projectatomic/atomicapp/blob/master/atomicapp/providers/kubernetes.py#L203) or am I missing something?
ReplicationControllers are scaled down to 0, but then they are also deleted.

I can change it here, but we should change it in Kubernetes also, so it is consistent across all providers.

cdrage:
New code LGTM :+1: 

@dustymabe @rtnpro can we get a secondary review and let's get this merged? :)

@kadel PR needs another rebase btw :dancers:

kadel:
rebased 

@dharmit  can you please have a look on docs `providers.md` if there is something missing we should add?

cdrage:
@kadel still has conflicts / needs re-basing to master

kadel:
@cdrage  rebased again :wink:

cdrage:
@dustymabe @rtnpro Can we have a secondary review before merge? Everything LGTM from my perspective :+1:

dustymabe:
Does marathon not need any authentication? Would we ever want to connect to a remote instance of mesos/marathon that does require one to authenticate?

atomicapp/providers/marathon.py
```diff
@@ -0,0 +1,119 @@
...
+                try:
+                    data = anymarkup.parse(fp)
+                    logger.debug("Parsed artifact %s", data)
+                    # every marathon app has to have id. 'id' key  is also used for showing messages
+                    if "id" not in data.keys():
+                        raise ProviderFailedException(
+                            "Error processing %s artifact. There is no id" % artifact)
+                except anymarkup.AnyMarkupError, e:
+                    msg = "Error processing artifact - %s" % e
+                    printErrorStatus(msg)
```
dustymabe: still need to raise exception right?

atomicapp/providers/marathon.py
```diff
@@ -0,0 +1,119 @@
...
+                "apps/%s" %
+                artifact["id"])
+
+            if self.dryrun:
+                logger.info("DRY-RUN: %s", url)
+                continue
+
+            logger.debug("Deleting appid: %s", artifact["id"])
+            try:
+                res = requests.post(url, json=artifact)
```
dustymabe: should we move `_make_request()` from the openshift provider into utils.py and rename it to `_make_rest_request()` and then use that in both the openshift provider and in the mesos provider?

If we do that I would suggest we do it in a separate PR.. zip it in real quick.. and then update this PR to reflect that.

dustymabe:
Adding docs to each function definition in marathon.py would be nice.

dustymabe:
Also are we doing ourselves a disservice by not using the work "Mesos" anywhere in here? I think people identify more with "Mesos" than with "Marathon". Would it be technically wrong to call this entire thing Mesos? Probably, but just thought I would air it out there.

atomicapp/providers/marathon.py
cdrage: and add a small comment on there (moving function to utils.py in future) or something so we can at least keep track of it.

@dustymabe agreed that a make_request function should exist in utils if we're going to be using this in openshift / marathon / kubernetes (in the future)

atomicapp/providers/marathon.py
kadel: oh :blush: , you are right

kadel:
On one hand you are right about identifying with Mesos, but on other hand if we rename entire thing to Mesos it would be technically incorrect :-( 
I will at least try mention Mesos more in docs.

kadel:
About auth: Marathon alone doesn't have any authentication. It is usually run in closed environment, or some people put reverse proxy (Nginx) in front of Marathon a do HTTP Basic Auth there.
If this is case, it should be enough to set `providerapi` to something like this `http://user:pass@127.0.0.1:8080` and it should work ;-)

dharmit:
Adding a bit what @kadel already said regarding identifying with Mesos. Marathon is just one framework. You could very well use Aurora or K8s on top of Mesos to get long running jobs done. But Marathon is used the most today. Agree about adding Mesos in the docs.

kadel:
@cdrage @dustymabe  I've just opened PR #463 that moves `_make_request` to Utils.
When it is merged I'll update this PR.

cdrage:
So make_rest_request has been added to utils and docs have been updated.

Am I missing anything or is this close to be merged?

dustymabe:
LGTM.

cdrage:
I did one last quick check and it seems everything is in order!

:sparkles: Congrats @kadel @dharmit !! :D Thanks for the hard work to getting this merged! :sparkles:

PR #395, opened 2015-11-12
Rework Dockerfiles and requirements

dustymabe:


dustymabe:
@LalatenduMohanty since you maintain some of the atomicapp rpms downstream I am interested in your comments on this

@cdrage since you have reworked requirements in the past and you pretty much maintain debian support for us can you please take a look as well? As part of this we are no longer pulling dependencies from pip for our rpm based distros and as a result I have removed strict versioning on our requirements; having it this way makes it harder to build for different platforms. It's better to introduce version requirements when we find an issue, (like we did for lockfile).

dustymabe:
@charliedrage ^^

Dockerfile
```diff
@@ -11,20 +11,24 @@ LABEL io.projectatomic.nulecule.atomicappversion="0.2.1" \
...
-# lets install pip, and gcc for the native extensions
-# and remove all after use
-RUN yum -y install epel-release && \
-    yum install -y --setopt=tsflags=nodocs python-pip python-setuptools docker gcc && \
-    python setup.py install && \
-    yum remove -y gcc cpp glibc-devel glibc-headers kernel-headers libmpc mpfr python-pip && \
+# Install needed requirements
+RUN yum install -y epel-release \
+    yum install -y --setopt=tsflags=nodocs docker && \
+    yum install -y --setopt=tsflags=nodocs $(sed s/^/python-/ requirements.txt) && \
```
goern: aint that layer carrying too much .rpm? like the gcc and glibc-devel?

rtnpro:
:+1: LGTM

Dockerfile
dustymabe: we are no longer installing gcc etc.. any more. only docker and the python dependencies via rpm. We are not using pip to get deps and then compiling them, so no need for gcc/glibc-devel.

charliedrage:
All good Debian side, everything LGTM :+1: 

ping @LalatenduMohanty for comments

dustymabe:
Will wait for final comments from @LalatenduMohanty and @goern before merging.

vpavlin:
I like it! LGTM:-) THanks @dustymabe

goern:
LGTM :+1:

LalatenduMohanty:
@charliedrage @dustymabe  LGTM!

PR #402, opened 2015-11-17
Change tmp dir location

charliedrage:
Imo, we should be extracting to 

`/var/lib/atomicapp/projectatomic-helloapache-latest-cf68076601d3` 
rather than 
`/var/lib/atomicapp/projectatomic-helloapache:latest-cf68076601d3`

Since metacharacters within dir names is bad practice. (including semicolons, spaces, backslashes, dollar signs, question marks, and asterisks)

PR #403, opened 2015-11-17
Change tmp dir location

cdrage:
Renamed branch for https://github.com/projectatomic/atomicapp/pull/402 pull

### DIFF
--- a/atomicapp/utils.py
+++ b/atomicapp/utils.py
@@ -117,7 +117,7 @@ class Utils(object):
 
     @staticmethod
     def sanitizeName(app):
-        return app.replace("/", "-")
+        return app.replace("/", "-").replace(":", "-")
 
     @staticmethod
     def getNewAppCacheDir(image):


vpavlin:
lgtm

dustymabe:
Can you add the wording in the first comment in #402 into the commit message? I think it is useful.

cdrage:
@dustymabe sorry, i'm a bit confused from both comments, what wording into the commit? I believe I did by changing the commit message to "Change tmp dir location"

dustymabe:
@cdrage I'm talking about the first comment on #402 . The one that starts with "Imo, we should be extracting to". I think that information would be useful in the commit message. So just git `commit --amend` the commit and edit the commit message if you agree.

rtnpro:
:+1: LGTM

cdrage:
@dustymabe updated

dustymabe:
looks good

PR #406, opened 2015-11-18
Update requirements doc

cdrage:
To reflect the changes to the Dockerfile's as well as adding a link to the current Nulecule spec (we rely on the 0.2.2 release)

### DIFF
--- a/docs/requirements.md
+++ b/docs/requirements.md
@@ -2,10 +2,12 @@
 
 Requirements needed to run Atomic App.
 
+
 ## Python requirements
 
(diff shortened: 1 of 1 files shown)


dustymabe:
Should we remove the specific versioning on the requirements ?

cdrage:
@dustymabe Yeah, we should, since we use yum/apt-get now and pip is in there by default too so we wouldn't run into the lockfile issue again.

updated the PR ^^

cdrage:
closing in favor of #427

PR #415, opened 2015-11-23
[Review please!] Logging refactor / cleanup

cdrage:
This is a work-in-progress.

It currently looks like this:
logging

### DONE!

Review away :)


dustymabe:
Should the `[OK]` messages in your picture of example output be `[INFO]` instead?

cdrage:
@dustymabe Yeah, they should :)

sub-mod:
@cdrage
All the display messages cannot be shown on UI. I would like a method where I can add cockpit specific message like display("step 1 completed", cockpit=true) and these messages should show up when i put cockpit flag

also,  can journald output be another PR so that we can have some code for me to test with cockpit

cdrage:
Yup. JournalD will be another PR. At the moment its not possible.

I'll update the PR and add a function for cockpit only info messages so you
can at least test it with cockpit.

cdrage:
The above failure is normal due to atomicapp loading via the new logging class rather than the actual logger module ^^. This will be fixed once we convert all `logger.info` to `display.info`, etc.

cdrage:
Hey all,

So I've added a different commit to re-implement cockpit support but unfortunately there are a few things which are missing from the changes of atomicapp after the refactor. I've noted these in the original post as well as opened up the issues pertaining to them.

The hard part is over, the only thing left is converting all the `logging.info("foobar")` calls to `display.info("foorbar")`.

cdrage:
Update: PR is mostly done. Cockpit integration added and got a :+1: from @sub-mod in regards to the current code.

Travis is failing due to "logger.info" code not converted yet to "display.info". 

If anyone wants to review the code, please go ahead and do so!

rtnpro:
@cdrage I checked logging with some hacks on cockpit's nulecule implementation, and the logs from `atomicapp` show up fine. :+1: from me.

cdrage:
@rtnpro 

Yayyyyyyyy :sparkles:

I'll do that last bit of this PR then to get the standard logging in. Going to be *a lot* of rebases after this is merged due to changing all the logger.info messages.

atomicapp/display.py
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    # Upon initialization we grab the log level value and verbose level (-v or not)
+    def __init__(self):
+        self.logger = logging.getLogger(LOG_NAME)
+        self.verbose_level = self.logger.getEffectiveLevel()
+        self.log_level = logging.getLogger(LOG_CUSTOM_NAME).getEffectiveLevel()
+
+    def debug(self, msg, only=None):
+        self.display(msg, 10, 'cyan', only)
```
dustymabe: should we be using the `DEBUG` constant here instead of `10`? Same goes for the statements below

dustymabe:
Talked with @cdrage. He is going to clean up commits and implement a few suggestions and update the PR.

atomicapp/display.py
```diff
@@ -0,0 +1,248 @@
...
+        Convert all input to utf-8 for multi language support
+        """
+        if type(input) != unicode:
+            input = input.decode('utf-8')
+            return input
+        else:
+            return input
+
+
+def set_logging(args):
```
dustymabe: Let's modify this to take vars as input. Looks like the few that we use are: `verbose`, `quiet`, and `logging_output`.

That way we aren't coupled to changes in cli/main.py.

atomicapp/display.py
```diff
@@ -0,0 +1,248 @@
...
+            msg = "[WARNING] %s" % msg
+        elif code == LOG_LEVELS['error']:
+            msg = "[ERROR] %s" % msg
+        else:
+            msg = "[INFO] %s" % msg
+
+        self._sysflush()
+        print(self._make_unicode(msg))
+
+    def _stdout_with_logging(self, msg, code, color):
```
dustymabe: `_stdout_via_logger` might be a good name. leaving the name as is or updating to that suggestion is fine.

atomicapp/nulecule/base.py
```diff
@@ -100,6 +101,7 @@ def unpack(cls, image, dest, config=None, namespace=GLOBAL_CONF,
             docker_handler = DockerHandler(dryrun=dryrun)
             docker_handler.pull(image)
             docker_handler.extract(image, APP_ENT_PATH, dest, update)
+            Display().info("All dependencies installed successfully.", "cockpit")
```
dustymabe: The NuleculeBase base class has `self.display = Display()`. Why not use that here?

atomicapp/nulecule/base.py
```diff
@@ -266,11 +269,13 @@ def __init__(self, name, basepath, source=None, params=None,
         self.rendered_artifacts = defaultdict(list)
         self._app = None
         self.config = config
+        self.display = Display()
```
dustymabe: The NuleculeBase base class has `self.display = Display()`. Why do we need to redefine it here?

atomicapp/cli/main.py
```diff
@@ -25,7 +25,7 @@
 from lockfile import LockFile
 from lockfile import AlreadyLocked
 
-from atomicapp import set_logging
+from atomicapp.display import set_logging
```
dustymabe: Doesn't work for me:

```
# aa run ./
  File "/opt/atomicapp/atomicapp/cli/main.py", line 28, in <module>
    from atomicapp.display import set_logging
ImportError: No module named atomicapp.display
```

atomicapp/nulecule/base.py
cdrage: @dustymabe 
see @classmethod at the beginning of the function. We can define it again however within it though if you'd prefer.

atomicapp/nulecule/base.py
cdrage: @dustymabe you're right, removing this ^^

atomicapp/nulecule/base.py
dustymabe: ahh I see. Yeah just makes it not seem uniform. Feel free to leave it like this or change it to make it clearer.

cdrage:
Okay so I've gone ahead and pushed all my other commits.

I have **one** failing error right now that could be due to a bug in the test suite. Otherwise, good for another review!

atomicapp/display.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def _stdout(self, msg, code):
+        """
+        Colorless logging output using the standard [DEBUG], [WARNING], etc tags.
+        """
+        if self.verbose_level is not LOG_LEVELS['debug'] and code is LOG_LEVELS['debug']:
+            return
+
+        if code == LOG_LEVELS['debug']:
+            msg = "[DEBUG] %6d %0.2f %s" % (os.getpid(), time.time(), msg)
```
dustymabe: After running it a couple of times I feel like the extra output in the debug statements (the pid/time) doesn't really help debug anything any more. We don't run multiprocess so I don't think this is useful. Thoughts?

atomicapp/display.py
```diff
@@ -0,0 +1,248 @@
...
+    def _stdout_via_logger(self, msg, code, color):
+        """
+        Colorful logging with the logger library.
+        """
+        if self.verbose_level is not LOG_LEVELS['debug'] and code is LOG_LEVELS['debug']:
+            return
+
+        if code == LOG_LEVELS['debug']:
+            self.logger.info(msg)
+            msg = "[DEBUG] %6d %0.2f %s" % (os.getpid(), time.time(), msg)
```
dustymabe: Same as comment above.

dustymabe:
So one really nice thing (for me) about the old output is that it included the name of the file in the output:

```
[root@f23 redis-centos7-atomicapp (upstreammaster *%=)]# aa run ./ --provider kubernetes                                                                                                                                                                                       
2016-02-17 20:37:21,499 - __main__ - INFO - Action/Mode Selected is: run
2016-02-17 20:37:21,532 - atomicapp.plugin - WARNING - Configuration option 'providerconfig' not found
2016-02-17 20:37:21,532 - atomicapp.plugin - WARNING - Configuration option 'providerconfig' not found
2016-02-17 20:37:21,532 - kubernetes - INFO - Using namespace default
2016-02-17 20:37:21,532 - kubernetes - INFO - trying kubectl at /host/usr/bin/kubectl
2016-02-17 20:37:21,533 - kubernetes - INFO - found kubectl at /host/usr/bin/kubectl
2016-02-17 20:37:21,535 - kubernetes - INFO - Deploying to Kubernetes
2016-02-17 20:37:21,808 - atomicapp.plugin - WARNING - Configuration option 'providerconfig' not found
2016-02-17 20:37:21,808 - kubernetes - INFO - Using namespace default
2016-02-17 20:37:21,809 - kubernetes - INFO - trying kubectl at /host/usr/bin/kubectl
2016-02-17 20:37:21,809 - kubernetes - INFO - found kubectl at /host/usr/bin/kubectl
2016-02-17 20:37:21,810 - kubernetes - INFO - Deploying to Kubernetes
```

Is there a way to preserve that with the new system?

atomicapp/display.py
cdrage: @dustymabe updated. yeah, too much output!

cdrage:
@rtnpro @dustymabe 

Hey all!

So I've updated the PR for final review.

One thing to note is that I will be removing the syslog logging output provider until a later date. I want to do more thorough testing with that.

I've also had to move a few things around in main.py in order to get the logging to output correctly. I checked to see if it would do any harm and from the looks if it, it wouldn't.

Otherwise, feel free to review and test!

atomicapp/display.py
```diff
@@ -0,0 +1,223 @@
...
+    else:
+        level = logging.INFO
+
+    # Let's check to see if any of our choices match the LOG_LEVELS constant! --logging-output
+    for k in LOG_LEVELS:
+        if logging_output == k:
+            custom_level = LOG_LEVELS[logging_output]
+            break
+        else:
+            custom_level = LOG_LEVELS['default']
```
dustymabe: LOG_LEVELS is a list right? could we do something like this:

```
if logging_output in LOG_LEVELS:
    custom_level = LOG_LEVELS[logging_output]
else:
    custom_level = LOG_LEVELS['default'] 
```

cdrage: @dustymabe Ahhhhh, forgot Python had that functinon >.> I'll add it

cdrage:
@dustymabe updated based on your comments for the `for` loop

dustymabe:
Closing in favor of #565

PR #416, opened 2015-11-24
Lib.py should load plugins in init

cdrage:
We shouldn't be loading plugins in file import, but rather in _init.

This blocks my logging PR as debug messages are incorrectly being displayed without --verbose being passed through. (set_logging within `cli/main.py` isn't set before loading `nulecule/lib.py`). We should be loading plugins via init of NuleculeManager.

Don't know why this was outside of the class in the first place. Tests and everything seem to pass!

Ping @rtnpro since it relates to your refactor.


cdrage:
timeout...
pinging @rtnpro again :)

dustymabe:
LGTM

PR #418, opened 2015-11-24
Adds in run_cmd() function to utils.py

dustymabe:


dustymabe:
@sub-mod please review. This is an attempt to fix the issues with #381.

cdrage:
@dustymabe @sub-mod 

I'm assuming that when we integrate an API provider with Kubernetes (rather than straight command lines), we could simply use `print` and `sys.stdout.flush()` correct?

dustymabe:
@cdrage - I'm not really sure I understand your question. if we aren't calling out to an external command then there should be no stdout or stderr to report. We would simply print a good error message based on the library call that failed and the returned output from the API.

cdrage:
The way that Cockpit has Nulecule/AtomicApp integrated is through CLI commands and grepping the terminal. If we were to implement the web-based API with kubernetes / openshift rather than CLI commands, we would instead simply use `sys.stdout.flush()` and `print` the error message from the http response rather than having to grep from the CLI commands.

dustymabe:
Cockpit would continue to screen scrape and grep when calling atomicapp the way they currently are. atomicapp doesn't screen scrape `kubectl` and `oc` currently but we want to save the output of the commands so we can report it when they fail.

In the case of a REST API call to an `oc` or `kube` endpoint we would simply check the response from the endpoint and capture the error message and report it via our standard logging. we would print out an error message in the format cockpit expects.

dustymabe:
ping @cdrage

cdrage:
Ah, I didn't reply to this?

LGTM :+1: I understand where we are coming from in regards to the kubectl command, let's merge this.

sub-mod:
@cdrage @dustymabe LGTM

Issue #407, opened 2015-11-18
[epic milestone] Switching to provider's API rather than direct command line.


charliedrage:
So we're changing Openshift from running the command to communicating to the API #321.

We are also adding Mesos that's _also_ talking via the API #392 

We should do the same with __Docker__ as well as __Kubernetes__. Sure, it'll add a lot more dependencies, but this gives us *a lot* more freedom for manipulation as well as error-checking / testing.

aweiteka:
+1

rtnpro:
:+1: @charliedrage, as I had already worked with ``docker-py`` in the past, I will start with migrating the ``docker`` provider to use ``docker`` API.

dustymabe:
- [ ] openshift #321 
- [ ] mesos #392
- [ ] docker #445
- [ ] kubernetes #446

PR #420, opened 2015-11-24
Use OpenShift API rather that `oc` cmd

kadel:
some todos TODOs:
- [x] handle templates
- [ ] implement stop (partially working)
- [ ] add tests
- [x] providerconfig support
- [x] update docs
- [x] handle timeouts
- [ ] turn on ssl_verify -> #451 

ref: #407  #321

### DIFF
--- a/atomicapp/constants.py
+++ b/atomicapp/constants.py
@@ -62,3 +62,5 @@ DEFAULT_ANSWERS = {
 PROVIDER_CONFIG_KEY = "providerconfig"
 
 PROVIDERS = ['docker', 'kubernetes', 'openshift']
+PROVIDER_API_KEY = "providerapi"
+ACCESS_TOKEN_KEY = "accesstoken"
--- a/atomicapp/providers/openshift.py
+++ b/atomicapp/providers/openshift.py
@@ -18,12 +18,13 @@
 """
 
 from atomicapp.plugin import Provider, ProviderFailedException
-from atomicapp.utils import Utils, find_binary
+from atomicapp.utils import printErrorStatus
 
-from collections import OrderedDict
 import os
 import anymarkup
-from distutils.spawn import find_executable
+import requests
+import urlparse
+from atomicapp.constants import PROVIDER_API_KEY, ACCESS_TOKEN_KEY, DEFAULT_NAMESPACE
 
 import logging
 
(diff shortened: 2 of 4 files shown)


aweiteka:
Do we want to deprecate support for the `oc` command? I can't think of a use case for it.

aweiteka:
I added tests TODO

goern:
I wouldnt deprecate anything on a "first attempt to use OpenShift api instead of oc command" commit, but ja. @dustymabe could plan to release that dependency (thats what I see, rather than a feature)

atomicapp/providers/openshift.py
```diff
@@ -36,91 +35,205 @@ class OpenShiftProvider(Provider):
...
-            artifact_path = os.path.join(self.path, artifact)
-            with open(artifact_path, "r") as fp:
+            with open(os.path.join(self.path, artifact), "r") as fp:
                 data = anymarkup.parse(fp, force_types=None)
-            if "kind" in data:
-                if data["kind"].lower() == "template":
-                    logger.info("Processing template")
-                    artifact = self._processTemplate(artifact_path)
-                kube_order[data["kind"].lower()] = artifact
+                # kind has to be specified in artifact
```
dustymabe: Any reason why the rest of this code can't be moved over out of the "with open()" block? We shouldn't need the fp around anymore after it has been read.

atomicapp/providers/openshift.py
```diff
@@ -17,16 +17,15 @@
...
-
-from collections import OrderedDict
 import os
 import anymarkup
-from distutils.spawn import find_executable
+import requests
+from urlparse import urljoin
+from atomicapp.plugin import Provider, ProviderFailedException
+from atomicapp.constants import (PROVIDER_API_KEY, ACCESS_TOKEN_KEY,
+                                 DEFAULT_NAMESPACE, NAMESPACE_KEY)
```
dustymabe: can you put these one on each line.

atomicapp/providers/openshift.py
```diff
@@ -36,91 +35,205 @@ class OpenShiftProvider(Provider):
...
+
+        logger.debug("Openshift resources %s", self.oapi_resources)
+        logger.debug("Kubernetes resources %s", self.kapi_resources)
+
+        self._process_artifacts()
+
+    def _get_namespace(self, artifact):
+        """ return artifacts namespace
+
+        if specified use namespace from artificaft else return self.namespace
```
dustymabe: spelling

atomicapp/providers/openshift.py
```diff
@@ -36,91 +35,205 @@ class OpenShiftProvider(Provider):
...
+                self.openshift_artifacts[kind].append(data)
+
+    def _process_template(self, template):
+        """ process tempalate
+
+        return list of templated object from template
+        """
+        logger.debug("processing template: %s", template)
+        url = self._get_url(self._get_namespace(template), 'processedtemplates')
+        res = requests.post(url, json=template, verify=self.ssl_verify)
```
dustymabe: what is this request doing exactly?

atomicapp/providers/openshift.py
```diff
@@ -36,91 +35,205 @@ class OpenShiftProvider(Provider):
...
+
+    def _kind_to_resource(self, kind):
+        """ converts kind to resource name
+
+        same logic as in k8s.io/kubernetes/pkg/api/meta/restmapper.go
+        KindToResource
+        """
+        singular = kind.lower()
+        if singular.endswith("status"):
+            plular = singular + "es"
```
dustymabe: so you mean plural?

atomicapp/providers/openshift.py
```diff
@@ -36,91 +35,205 @@ class OpenShiftProvider(Provider):
...
+        else:
+            msg = "%s %s" % (res.status_code, res.content)
+            logger.error(msg)
+            raise ProviderFailedException(msg)
+
+    def _kind_to_resource(self, kind):
+        """ converts kind to resource name
+
+        same logic as in k8s.io/kubernetes/pkg/api/meta/restmapper.go
+        KindToResource
```
dustymabe: examples of the few cases below would be nice.

atomicapp/providers/openshift.py
```diff
@@ -36,91 +35,205 @@ class OpenShiftProvider(Provider):
...
+    def _get_url(self, namespace, kind, name=None):
+        """ return url for given kind
+
+        return either openshift or kubernetes url
+        depending on what kind is passed
+        """
+        url = None
+
+        resource = self._kind_to_resource(kind)
+
```
dustymabe: Explain the url you are piecing together in all of these following commands. Showing what the url is composed of and one resulting url would be nice.

atomicapp/providers/openshift.py
```diff
@@ -36,91 +35,205 @@ class OpenShiftProvider(Provider):
     cli = None
     config_file = None
     template_data = None
+    providerapi = "https://127.0.0.1:8443"
```
dustymabe: So this is the default? Is that where openshift listens by default if on local machine? - We should probably make this get set in constants.py and used here.

dustymabe:
Would be nice for each function to have a full description as well as each argument to the function described. Easier to do them now than go back and do them later :) 

Additionally is it possible with this new design to still pick up credentials from the user's system `~/.kube/config` and parse the providerapi and token out of that? This would require us to identify the user and find the file (which is not something you need to do) but I'd like us to consider this as an option. For now it would be sufficient that if someone has provided a providerconfig, to be able to use it to get the providerapi and access token. Is that reasonable?

atomicapp/constants.py
```diff
@@ -61,3 +61,5 @@
 PROVIDER_CONFIG_KEY = "providerconfig"
 
 PROVIDERS = ['docker', 'kubernetes', 'openshift']
+PROVIDER_API_KEY = "providerapi"
```
dustymabe: I wonder if this should include the word "endpoint".. not really sure

atomicapp/providers/openshift.py
kadel: In returns "rendered" template according to its parameters. Main reason for this is that OpenShift supports generating values from expression for parameters (https://docs.openshift.com/enterprise/3.0/architecture/core_concepts/templates.html)

It is a bit confusing because now we have two levels of templating :-( 
Example of this is here: https://github.com/kadel/openshift3mlbparks/blob/nulecule-two-templates/mlbparks-app-template.json
All `$app_*` variables are Nulecule parameters rest (uppercase variables) are OpenShifts

atomicapp/constants.py
vpavlin: I immediately imagined a token while reading `PROVIDER_API_KEY` so the name is probably not good

atomicapp/providers/openshift.py
vpavlin: Ha, `constant.py` was my first thought too, but it's provider specific thing and as providers are plugins, they should be removable without leaving traces behind (like unused constants etc.)

atomicapp/providers/openshift.py
```diff
@@ -36,91 +35,205 @@ class OpenShiftProvider(Provider):
...
+    openshift_api = None
+    kubernetes_api = None
```
vpavlin: Why is kubernetes referenced here? I know it's a base of OpenShift, but why would we ever wanted to use it in OS provider?

atomicapp/providers/openshift.py
vpavlin: I love the work `artificaft`..I think we need to leave it there:D

atomicapp/providers/openshift.py
```diff
@@ -36,91 +35,205 @@ class OpenShiftProvider(Provider):
...
+                    logger.error(msg)
+                    raise ProviderFailedException(msg)
+
+    def _process_artifacts(self):
+        """  parse artifact files
+
+        save parsed artifacts to self.openshift_artifacts
+        """
         for artifact in self.artifacts:
+            logger.debug("Procesesing artifact: %s", artifact)
```
vpavlin: spelling

atomicapp/constants.py
kadel: We already discussed it here https://github.com/projectatomic/atomicapp/pull/392#discussion_r44532586 :-)
I was also thinking about `providerendpoint` and `providerurl`, but first is longer :-) and second is not good.

atomicapp/providers/openshift.py
kadel: You are right.
Fixed

atomicapp/providers/openshift.py
kadel: done

atomicapp/providers/openshift.py
kadel: fixed :-)

atomicapp/providers/openshift.py
kadel: fixed

atomicapp/providers/openshift.py
kadel: done

atomicapp/providers/openshift.py
kadel: done

atomicapp/providers/openshift.py
kadel: fixed

atomicapp/constants.py
vpavlin: Yeah, I realized it later:)

atomicapp/providers/openshift.py
```diff
@@ -36,91 +37,302 @@ class OpenShiftProvider(Provider):
...
+        url = urljoin(url, "?access_token={}".format(self.access_token))
+        logger.debug("url: %s", url)
+        return url
+
+    def _parse_kubeconf(self, f):
+        """"
+        Parse kubectl config file
+
+        Args:
+            f (string): path to configuration file (e.g. ./kube/config)
```
dustymabe: could we make this 'file' vs 'f' ?

atomicapp/providers/openshift.py
```diff
@@ -36,91 +37,302 @@ class OpenShiftProvider(Provider):
...
+        logger.debug("Parsing %s", f)
+
+        with open(f, 'r') as fp:
+            kubecfg = anymarkup.parse(fp.read())
+
+        current_context = kubecfg["current-context"]
+
+        logger.debug("current context: %s", current_context)
+
+        context = None
```
dustymabe: The code is beatifully written but I do think if we had an example of what the file looks like somewhere here in the comments it would make the code easier to read

dustymabe:
Thanks so much for the update you made yesterday to support providerconfig again. Can we get the docs updated to mention it again in an either/or fashion? So now the user can either provide a kube config or they can provide api/token. 

One thing I see as confusing to the user is the fact that we pick up the namespace from the providerconfig. I agree that we should do that, but maybe we can make it better.

I propose that we always use the namespace from the answers.conf file, but we should actually error out if the user has chosen to use providerconfig && (the namespace in the answers.conf + the namespace in providerconfig file don't match).

One example is that for openshift I am using the namespace `proj1`, but in my answers.conf file the namespace is `default`. Rather than just using `proj1` (picked up from providerconfig) I would rather it error and let me know the namespace should be set to match.

Thoughts?

kadel:
At first I printed warning about two different namespaces and that one from providerconfig is used, but for some reason I changed to silently ignoring values from answers.conf (I don't remember why :blush: )
Error out is maybe better, so there will be no confusion. I would do same thing for providerapapi and accesstoken in answers, currently they are also silently ignored with providerconfig is in answers.

atomicapp/providers/openshift.py
kadel: sure, I am going to change it to `filename`, `file` is a build-in function

dustymabe:
@cdrage do you mind giving this PR a quick look (looking for any bugs). You don't need to spend a lot of time on it, just look for any errors.

dustymabe:
@kadel.. we're almost there! do you mind updating the docs again now that providerconfig is an option. Also mentioning that providerconfig and access_token + providerapi are mutually exclusive  as well that namespace information could come from answers or from providerconfig. 

Thanks man. Have a good weekend

kadel:
I tried to update docs. But I think that someone should fix grammar after me.  I am sure that there is a lot of mistakes that I cannot see. :-(

dustymabe:
LGTM. For the remaining 3 items let's break them out into different issues and make them a priority. 

- implement stop
- add tests
- turn on ssl_verify --> I think this issue covers it: https://github.com/projectatomic/atomicapp/issues/451

@cdrage, can you give a final review and then merge.

kadel:
OK, I will remove `undeploy()`, for now, so we don't release it when it is not working properly.

cdrage:
LGTM :+1: from initial testing!

Let's merge this  :sparkles:

dustymabe:
Merging.. @kadel please open up issues or new [WIP] Prs for the "stop" and "unit tests" and comment here linking to them. 

Great work!

kadel:
@dustymabe  done.
Thank you for helping with this PR!

Issue #397, opened 2015-11-12
Changing namespace format to namespace_IMAGENAME_hashvalue


cdrage:
When deploying an atomicapp it's namespace is commonly used as `default`. From our use cases as well as multiple examples that have appeared, many users are encountering  the issue of "this namespace is already in use" when deploying another example without a configured answers.conf file.

@dustymabe and I encountered these issues recently when working on a multi-inherited Nuleculized application.

I propose that we change namespacing containers from:

`atomic_NAMESPCE_HASHVALUE`

to

`NAMESPACE_IMAGENAME_HASHVALUE`

In order to indicate what containers have been deployed by atomicapp.

I also propose that we change the value 'default' as a namespace to 'atomicapp' to better indicate what has been deployed.

dustymabe:
@charliedrage To be clear we are only talking about doing this if the user didn't provide "--name" in the docker run command. We can't change that right now because of docker links. 

I like `NAMESPACE_IMAGENAME_HASHVALUE`. I don't know if I like making `atomicapp` the default namespace though. I like keeping it set as `default`, but I also like some indication that this is an atomicapp. Perhaps we can just do `atomicapp_NAMESPACE_IMAGENAME_HASHVALUE`.

charliedrage:
@dustymabe 
only [a-zA-Z0-9][a-zA-Z0-9_.-] are only allowed in the image name.

For example, if we have centos/httpd:latest. What should we convert it as, centos-httpd-latest, centos.httpd.latest?

dustymabe:
I would go with dashes

dustymabe:
cdrage is this fixed now with #422?

cdrage:
Yup. Thought that this would close automatically with the merge comment ^^ Closin'.

Issue #398, opened 2015-11-13
Nested atomicapp applications error on the same namespace


charliedrage:
This semi-relates to issue #397

When using:
 - The Docker provider
 - A `atomicapp` container with inherited / nested `atomicapp` application
 - and.. all namespaces are default / not assigned

You encounter an error where you're unable to deploy the second inherited `atomicapp` application, ending up with a `this namespace is already in use` error.

This is due to the fact that you cannot assign a unique namespace to the inherited `atomicapp` container __as well__ as the fact that our current regex expression ` if re.match("%s_+%s+_+[a-zA-Z0-9]{12}" % (self.default_name, self.namespace), container):` within `docker.py` will match the first `atomicapp` container that was deployed.

dharmit:
I hit this issue on `0.2.1`.

dustymabe:
@cdrage I thought we talked about the implementation for this.. If we are deploying two instances of the same application:

* instance of exampleapp (nulecule) - with namespace "default"
    - appa (nulecule)
        + image1
        + appaa (nulecule)
            - image2
    - appb (nulecule)
        + image3
        + image4
* instance of exampleapp (nulecule) - with namespace "default1"
    - appa (nulecule)
        + image1
        + appaa (nulecule)
            - image2
    - appb (nulecule)
        + image3
        + image4

we would create image names:
default_appa_image1_$(RANDOM)
default_appa_appaa_image2_$(RANDOM)
default_appb_image3_$(RANDOM)
default_appb_image4_$(RANDOM)
default1_appa_image1_$(RANDOM)
default1_appa_appaa_image2_$(RANDOM)
default1_appb_image3_$(RANDOM)
default1_appb_image4_$(RANDOM)

or something along those lines

cdrage:
@dustymabe 

Yes, you're right, we should do it along those lines (especially for the Docker provider).

I was a tad confused with https://github.com/projectatomic/nulecule/issues/187 and @rtnpro proposal. But his one has to do more with namespacing in the answers.conf.

The docker provider should detect if it is an inherited application and act accordingly with naming the Docker containers.

rtnpro:
@cdrage ping? we need to sync up on this.

cdrage:
@rtnpro this will be fixed before GA. are you good with @dustymabe 's proposal on the naming scheme?

PR #422, opened 2015-11-24
Change naming on Docker deploy, fix tests

cdrage:
This commit not only changes how we deploy containers.

Using: NAMESPACE_IMAGENAME_HASHVALUE
instead of: atomic_NAMESPACE_HASHVALUE

But cleans up the tests to reflect the change

If an image isn't provided in Nulecule (we use it directly from the Docker artifact), we will instead using a default image name instead.

Ex. If image is not provided.
````yaml
graph:
  - name: helloapache-app
    params:
      - name: image
        description: The webserver image
        default: centos/httpd
```

Fixes issue #397

### DIFF
--- a/atomicapp/providers/docker.py
+++ b/atomicapp/providers/docker.py
@@ -40,6 +40,13 @@ class DockerProvider(Provider):
             self.namespace = self.config.get("namespace")
         logger.debug("Namespace: %s", self.namespace)
 
+        if "image" in self.config:
+            self.image = Utils.sanitizeName(self.config.get("image"))
+        else:
+            self.image = Utils.getUniqueUUID()
+            logger.warning("The artifact name has not been provided within Nulecule, using a UUID instead")
+            logger.debug("No image name found for artifact, using UUID %s in container name" % self.image)
+
         if self.dryrun:
             logger.info("DRY-RUN: Did not check Docker version compatibility")
         else:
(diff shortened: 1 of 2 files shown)


dustymabe:
It's unfortunate that the image could only be in the artifact and not in the nulecule.. Could we possibly inspect the image from the docker artifact? I know it's more work but kinda sucks to have to use a default name.

otherwise LGTM

cdrage:
@dustymabe 
Yeah... It'll be a bit difficult. 

For example, out of this output, we need to detect what part is the image name:
`docker run --net=host -d gcr.io/google_containers/etcd:2.0.9 /usr/local/bin/etcd --addr=127.0.0.1:4001 --bind-addr=0.0.0.0:4001 --data-dir=/var/etcd/data`

I'll see what I can do.

dustymabe:
don't worry about it.. but I think we shouldn't have a default name. if there is no image name then I say we come up with something random and use that instead.

cdrage:
@dustymabe another uuid gen perhaps?

dustymabe:
@cdrage yeah that would be fine

cdrage:
@dustymabe This has been updated to include the UUID if an image is not specified in the Nulecule spec (and by 

This will fix the issues that people have been having running multiple images with the same names (fixes issue @kanarip had)

Shall we merge this in with the UUID change and when we convert the `docker.py` client to the API dependency we can add the image detection?

dustymabe:
@cdrage I thought @kanarip's issue was https://github.com/projectatomic/atomicapp/issues/398 and not this one.

cdrage:
@dustymabe I believe it's both, since I myself ran into the same issue too when using one or more of the same images with the default name. 

updated the PR from your comments ^^

atomicapp/providers/docker.py
```diff
@@ -40,6 +40,13 @@ def init(self):
             self.namespace = self.config.get("namespace")
         logger.debug("Namespace: %s", self.namespace)
 
+        if "image" in self.config:
+            self.image = Utils.sanitizeName(self.config.get("image"))
+        else:
+            self.image = Utils.getUniqueUUID()
+            logger.warning("The artifact name has not been provided within Nulecule, using a UUID instead")
+            logger.debug("No image name found for artifact, using UUID %s" % self.image)
```
cdrage: @dustymabe updated again ;) haha

dustymabe:
@cdrage do you want to merge this after tomorrows release?

cdrage:
@dustymabe Yes, let's do just that, merge after 0.2.3

cdrage:
merginnnnnnnnn :dancers:

PR #424, opened 2015-11-28
Don't run container to extract files.

dustymabe:
In the past we didn't run a container to extract files but in cb0625b
we started doing that to workaround bug [1]. Now that the bug is fixed
let's revert back to just creating the containers and not actually
running them. This allows us to support nulecule metadata containers
that just contain nulecule metadata in them and are build "FROM scratch".

[1] - https://bugzilla.redhat.com/show_bug.cgi?id=1252168

### DIFF
--- a/atomicapp/nulecule/container.py
+++ b/atomicapp/nulecule/container.py
@@ -76,8 +76,8 @@ class DockerHandler(object):
 
         # Create dummy container
         run_cmd = [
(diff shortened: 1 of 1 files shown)


aweiteka:
For a bit of context, this enables us to run the atomicapp base container with a metadata-only Nulecule image as an argument.

cdrage:
LGTM :+1: since the bug was fixed in 1.8

PR #426, opened 2015-11-30
Fix tests not removing answers.conf.gen

cdrage:
This commit fixes answers.conf.gen not being removed after a failed
test.

Flake8 was also ran on both test_cli.py and gitlab testing files.
Logging and shutil dependancies were removed due to no use from them (we
collect the log anyways on test ran).


dustymabe:
LGTM.

PR #431, opened 2015-12-02
Move global cli options to sub command level.

### LINKED ISSUE #332: Should ordering matter for options?
We have had discussion about our strict ordering for optional arguments: i.e. if it is a global option then it must go before the "verb", if it is a specific option then it must go after the verb. Some of us would rather not have ordering for options, which we think is reasonable as long as the number of global arguments doesn't get to be too many. 

Let's revisit this after beta3.

rtnpro:
In atomicapp CLI, we are moving global options to sub command level, so that there's little confusion which options go before a sub command and which go after.

Fixes #332.

### DIFF
--- a/atomicapp/cli/main.py
+++ b/atomicapp/cli/main.py
@@ -108,7 +108,9 @@ class CLI():
 
     def set_arguments(self):
 
-        self.parser.add_argument(
+        base_parser = ArgumentParser(add_help=False)
+
+        base_parser.add_argument(
             "-V",
             "--version",
             action='version',
@@ -117,7 +119,7 @@ class CLI():
             help="show the version and exit.")
         # TODO refactor program name and version to some globals
 
-        self.parser.add_argument(
+        base_parser.add_argument(
             "-v",
             "--verbose",
             dest="verbose",
@@ -125,7 +127,7 @@ class CLI():
             action="store_true",
             help="Verbose output mode.")
 
-        self.parser.add_argument(
+        base_parser.add_argument(
             "-q",
             "--quiet",
             dest="quiet",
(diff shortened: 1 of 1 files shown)


cdrage:
@rtnpro tests failed due to no update in test components.

I don't see this affecting any other projects other than cockpit, which @sub-mod has been working on. We will need to push upstream a PR in order to fix the integration since we will have to move the arguments around. But I'll open up another issue for that :)

rtnpro:
@cdrage Thanks :)

cdrage:
otherwise, LGTM :+1:

dustymabe:
I want to spend some time looking at this before we merge. I was doing that tonight but got sidetracked with centos CI stuff and then docker 1.9 hosed my fedora host. Will look tomorrow and comment

dustymabe:
@rtnpro please check out the small bit of code in https://github.com/dustymabe/atomicapp/commit/a8cf5fd5a273e13afc89b6fc55355488424985ca. It allows us to support options anywhere by figuring out what keyword will be used to activate a subparser and then moving that the the beginning of the options. If you agree this is a viable solution then we can commit this PR and I'll open a PR with that code (that builds on top of this PR).

@cdrage.. this should make it so we don't break any existing code that calls atomicapp.

rtnpro:
@dustymabe :+1:

cdrage:
errrrhm ^^ tests failed as noted from my comment before. @rtnpro can you please open up a new PR to fix them.

dustymabe:
The tests are fixed by my PR https://github.com/projectatomic/atomicapp/pull/443. Otherwise I wouldn't have merged

PR #439, opened 2015-12-04
Adds more information to running --dry-run

cdrage:
More information when running `sudo atomicapp -v --dry-run run APP`, since we are required to provide the actual extracted Nulecule application as a form of a path rather than a docker-image.


dustymabe:
I would just use `/path/to/yourapp` rather than putting in the `CACHEDIR`

cdrage:
Updated!

PR #440, opened 2015-12-04
Add default provider

### LINKED ISSUE #380: If there's only one provider, provided, we should default to it.
^^

If we have only a _docker_ provider in the `/artifacts/` folder, we should default to it while deploying.

cdrage:
__First fix:__
One of them being that a default provider isn't passed when you don't
have anything else but Kubernetes in the artifacts folder.

__Second fix:__
With the implementation of the first fix, it seems that it caused a bug within the nulecule core library. Wrong data is passed to the provider when overriding the current default provider. 

ex. `atomicapp -v run —provider=docker projectatomic/helloapache`

is passed Given config: {u‘image’: u’centos/httpd’, u’hostport’: 80,
’namespace’: ’default’, ’provider’: ’kubernetes’}

rather than Given config: {u‘image’: u’centos/httpd’, u’hostport’: 80,
’namespace’: ’default’, ’provider’: ’docker’}

__Other:__
Tests have been added to reduce any regressions

fixes #380


atomicapp/constants.py
```diff
@@ -52,10 +53,10 @@
 DEFAULT_PROVIDER = "kubernetes"
 DEFAULT_CONTAINER_NAME = "atomic"
 DEFAULT_NAMESPACE = "default"
-DEFAULT_ANSWERS = {
+DEFAULT_ANSWERS_FORMAT = {
     "general": {
-        "provider": DEFAULT_PROVIDER,
-        "namespace": DEFAULT_NAMESPACE
+        "provider": None,
+        "namespace": "default"
```
cdrage: This will have to be changed in the future, I _don't_ like how we use "default"

atomicapp/nulecule/lib.py
```diff
@@ -21,7 +21,7 @@ def __init__(self, basepath, params, namespace):
     def load(self):
         pass
 
-    def load_config(self, config=None, ask=False, skip_asking=False):
+    def load_config(self, config, ask=False, skip_asking=False):
```
cdrage: There has never been a case where load_config has been passed __None__ _yet_.

tests/units/cli/test_examples/helloapache/answers.conf.sample
```diff
@@ -7,4 +7,4 @@
     "namespace": "default",
     "provider": "kubernetes"
   }
-}
\ No newline at end of file
+}
```
cdrage: fix indentation yo, cause by auto flake8

cdrage:
I need to add a doc to this before merge or else people will think it's unexpected behavior.

atomicapp/nulecule/lib.py
```diff
@@ -36,7 +37,6 @@ def load_config(self, config=None, ask=False, skip_asking=False):
         Returns:
             None
         """
-        config = config or DEFAULT_ANSWERS
```
cdrage: redundant and not needed :)

atomicapp/constants.py
```diff
@@ -52,10 +53,10 @@
 DEFAULT_PROVIDER = "kubernetes"
 DEFAULT_CONTAINER_NAME = "atomic"
 DEFAULT_NAMESPACE = "default"
-DEFAULT_ANSWERS = {
+DEFAULT_ANSWERS_FORMAT = {
```
dustymabe: would rather leave this as DEFAULT_ANSWERS

atomicapp/constants.py
dustymabe: I think we should leave this as it was. I know the change here was to specify "None" but I don't really think we need that and will comment on how/why.

atomicapp/nulecule/lib.py
```diff
@@ -1,16 +1,17 @@
 # -*- coding: utf-8 -*-
-from atomicapp.constants import (GLOBAL_CONF, DEFAULT_PROVIDER,
-                                 DEFAULT_ANSWERS, NAME_KEY,
-                                 DEFAULTNAME_KEY, PROVIDER_KEY)
+from atomicapp.constants import (
+    GLOBAL_CONF, NAME_KEY, DEFAULTNAME_KEY, PROVIDER_KEY)
```
dustymabe: would prefer to have each one on its own line

atomicapp/nulecule/main.py
```diff
@@ -183,6 +184,11 @@ def run(self, cli_provider, answers_output, ask,
         # it does exist it will be just be loaded and returned
         self.nulecule = self.unpack(dryrun=dryrun, config=self.answers)
 
+        # If no provider is specified within answers.conf or --provider=
+        # we should try and get a default one from the artifacts folder
+        if self.answers[GLOBAL_CONF][PROVIDER_KEY] is None and cli_provider is None:
+            self.answers[GLOBAL_CONF][PROVIDER_KEY] = Utils.getDefaultArtifact(self.app_path)
+
```
dustymabe: I believe this would be better served by simply leaving the default as it was set in the DEFAULT_ANSWERS and calling out to a function that returns the list of supported providers. If that list has a length of 1 then just set that provider:

```
# comment about single provider means we should default to it
supportedproviders = Utils.getAppSupportedProviders(self.app_path)
if cli_provider is None and len(supportedproviders) ==1:
    self.answers[GLOBAL_CONF][PROVIDER_KEY] = supportedproviders[0]
```

atomicapp/nulecule/main.py
```diff
@@ -183,6 +184,11 @@ def run(self, cli_provider, answers_output, ask,
...
         self.nulecule.load_config(config=self.nulecule.config, ask=ask)
         self.nulecule.render(cli_provider, dryrun)
```
dustymabe: I wonder why we pass in cli_provider to all of these rather than just setting it in the config and passing the information that way.

atomicapp/nulecule/main.py
```diff
@@ -254,9 +261,11 @@ def _get_runtime_answers(self, config, cli_provider):
             dict
         """
         _config = copy.deepcopy(config)
+        # If a provider is provided via CLI, override the config parameter
+        if cli_provider:
+            _config[GLOBAL_CONF][PROVIDER_KEY] = cli_provider
```
dustymabe: why is this above line 267?

dustymabe:
Can this be broken out into two commits? 

1 - set the config data so that it gets passed in correctly
2 - do the logic to determine if only one provider exists and make it the default if so.

Would make them much more consumable than "numerous little fixes"

atomicapp/constants.py
cdrage: If I specify None, then it's much easier to pick it up via determine if a provider has been passed or not. If we have to revert back to using DEFAULT_PROVIDER, I would have to determine and check to see if something is "Kubernetes" or not which would be bad practice.  I don't know how else I should explain, but passing in a template would be more ideal than relying on a default preference.

atomicapp/nulecule/main.py
cdrage: If you look at `getDefaultArtifact` it actually get's a default artifact by determining what's been provided in the `/artifacts` folder of the Nuleculized container. Wouldn't determining it that way be much more intuitive than checking to see if "kubernetes/DEFAULT_PROVIDER" has been set or not?

atomicapp/nulecule/main.py
cdrage: shouldn't be, woops, i'll re-add it back to the bottom.

atomicapp/nulecule/main.py
cdrage: me too, we should be determining cli_provider much earlier inthe code rather than at multiple points in `lib.py/main.py/base.py`.

cdrage:
@dustymabe I'll split it into to two commits, got carried away trying to implement one function and finding multiple bugs instead.

atomicapp/nulecule/main.py
dustymabe: I believe what you are doing in getDefaultArtifact and what I have above is essentially the same thing because Utils.getAppSupportedProviders would just return the list of providers that there are artifacts for. This would allow us to leave DEFAULT_ANSWERS the way they were in the past and still get the same functionality.

Let's chat about this tomorrow if you like.

atomicapp/nulecule/lib.py
rtnpro: :+1:

atomicapp/nulecule/lib.py
rtnpro: :+1:

tests/units/cli/test_default_provider.py
```diff
@@ -0,0 +1,76 @@
...
+        command = [
+            "main.py",
+            "--verbose",
+            "--dry-run",
+            "run",
+            self.examples_dir + 'oneprovider-helloapache/'
+        ]
+
+        # Run the dry-run command
+        with pytest.raises(SystemExit) as exec_info:
```
rtnpro: I know that ``pytest`` has been included as a test dependency quite some time back, but are we using ``pytest`` apart from ``pytest.raises``. Python's unittest.TestCase provides ``assertRaises`` method to do similar stuff. Any thoughts?

cdrage: @dustymabe have to talk to @vpavlin about that since pytest.raises is pretty much everywhere in our cli test ;)

dustymabe: @cdrage do you mean @rtnpro ?? :)

cdrage: @rtnpro Woops, yeah, I meant @rtnpro I need more coffee >.>

cdrage:
@dustymabe @rtnpro Updated everything, need secondary review.

atomicapp/nulecule/main.py
```diff
@@ -183,6 +184,12 @@ def run(self, cli_provider, answers_output, ask,
         # it does exist it will be just be loaded and returned
         self.nulecule = self.unpack(dryrun=dryrun, config=self.answers)
 
+        # If no provider is specified within answers.conf or --provider=
+        # we should try and get a default one from the artifacts folder
```
dustymabe: I think the comment needs updating:

```
# If only one provider is supported by app then default to it unless
# the user specified otherwise
```

dustymabe:
Might want to still update the commit message to be more specific that "Fix numerous little bugs"

dustymabe:
Rebase on master should probably fix test failures.

cdrage:
@dustymabe rebased && tests passed

docs/cli.md
```diff
@@ -2,7 +2,7 @@
 
 ## Providers
 
-Providers may be specified using the `answers.conf` file or the `--provider <provider>` option. If not specified the kubernetes provider will be used by default.
+Providers may be specified using the `answers.conf` file or the `--provider <provider>` option. If not specified, Atomic App will try to detect what artifacts are being used within the Nulecule container and use the first one available. If neither are detected, Atomic App will use Kubernetes by default.
```
dustymabe: This isn't exactly right. 

```
If not specified, but artifacts for only one provider are included within the Nulecule contianer, then Atomic App will use that provider.
```

dustymabe:
LGTM -> Merge!!

PR #443, opened 2015-12-07
Support for options anywhere on command line.

dustymabe:
This is done by taking the subparser work that rtnpro did in which
global options were added to the subparsers and simply doing a first
pass to detect which subparser would be used and then moving that
subparsers keyword to the beginning of the line.


dustymabe:
@cdrage please review.. @rtnpro has already given +1 (he looked directly at my commit)

cdrage:
LGTM and works :+1: let's merge.

PR #449, opened 2015-12-10
Minor fix in nulecule testing and flake8

cdrage:
[] is passed as a graph, this should be [{}] as graph is initially an
array but holds a hash table of values.

Flake8 formatting as well as on the test file is changed too.


dustymabe:
LGTM

PR #450, opened 2015-12-11
Adds dockerignore file.

dustymabe:
My git repo is getting large and sending all of that in the build
context to the daemon every time is a bit much. Let's only send the
essentials to the daemon.


cdrage:
LGTM

PR #453, opened 2015-12-14
Adds clean method to Makefile

cdrage:
Adds a clean method to the makefile to get ride of generated bin files and
misc files during install / build process.


dustymabe:
LGTM

PR #454, opened 2015-12-14
OpenShift Integration: support for `oc new-app`

dustymabe:
This PR is an initial attempt at supporting bootstrapping Atomic App 
embedded applications view the `oc new-app` command in OpenShift.

### DIFF
--- a/atomicapp/cli/main.py
+++ b/atomicapp/cli/main.py
@@ -31,6 +31,7 @@ from atomicapp.constants import (__ATOMICAPPVERSION__,
                                  __NULECULESPECVERSION__,
                                  ANSWERS_FILE,
                                  ANSWERS_FILE_SAMPLE_FORMAT,
+                                 APP_ENT_PATH,
                                  HOST_DIR,
                                  LOCK_FILE,
                                  PROVIDERS)
@@ -249,6 +250,13 @@ class CLI():
     def run(self):
         self.set_arguments()  # Set our arguments
         cmdline = sys.argv[1:]  # Grab args from cmdline
+
+        # If we are running in an openshift pod (via `oc new-app`) then
+        # there is no cmdline but we want to default to "atomicapp run".
+        # In this case copy files to cwd and use the working directory.
+        if Utils.running_on_openshift():
+            Utils.copy_dir(os.path.join('/', APP_ENT_PATH), './')
+            cmdline = 'run --verbose ./'.split()
 
         # We want to be able to place options anywhere on the command
         # line. We have added all global options to each subparser,
--- a/atomicapp/nulecule/main.py
+++ b/atomicapp/nulecule/main.py
@@ -90,6 +90,16 @@ class NuleculeManager(object):
             if os.path.isfile(os.path.join(self.app_path, ANSWERS_FILE)):
                 self.answers_file = os.path.join(self.app_path, ANSWERS_FILE)
 
+        # TODO: put this in a better place in the future.
+        # If we are running inside of an openshift pod then override
+        # some of the config by detecting some values from the environment
+        if Utils.running_on_openshift():
+            self.answers[GLOBAL_CONF]['provider'] = 'openshift'
+            self.answers[GLOBAL_CONF]['accesstoken'] = os.environ['TOKEN_ENV_VAR']
+            self.answers[GLOBAL_CONF]['namespace'] = os.environ['POD_NAMESPACE']
+            self.answers[GLOBAL_CONF]['providerapi'] = \
+                Utils.get_openshift_api_endpoint_from_env()
+
     def unpack(self, update=False,
                dryrun=False, nodeps=False, config=None):
         """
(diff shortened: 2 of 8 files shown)


aweiteka:
LGTM

dustymabe:
Note I also updated the provider documentation. It is in a separate PR: #455.

dustymabe:
@rtnpro @cdrage.. This is a big feature for us to add. The more eyes the better on this one.

atomicapp/cli/main.py
```diff
@@ -250,6 +251,13 @@ def run(self):
         self.set_arguments()  # Set our arguments
         cmdline = sys.argv[1:]  # Grab args from cmdline
 
+        # If we are running in an openshift pod (via `oc new-app`)
+        # then there is no cmdline but we want to default to "run". In
```
cdrage: For us non-openshift folks what does 'run' mean? as in, `oc run`?

atomicapp/constants.py
```diff
@@ -47,9 +47,11 @@
 ANSWERS_FILE_SAMPLE = "answers.conf.sample"
 ANSWERS_FILE_SAMPLE_FORMAT = 'ini'
 WORKDIR = ".workdir"
-LOCK_FILE = "/run/lock/atomicapp.lock"
 HOST_DIR = "/host"
 
+# Use /tmp/ so that we can grab the lock as non-root
+LOCK_FILE = "/tmp/atomicapp.lock"
```
cdrage: :+1:

atomicapp/nulecule/main.py
```diff
@@ -90,6 +90,16 @@ def __init__(self, app_spec, destination=None, answers_file=None):
             if os.path.isfile(os.path.join(self.app_path, ANSWERS_FILE)):
                 self.answers_file = os.path.join(self.app_path, ANSWERS_FILE)
 
+        # TODO: put this in a better place in the future.
```
cdrage: Better place could actually just be the `openshift.py` under providers.

I've been thinking lately of refactoring how we use each provider since we're adding new commands in the future. For example, changing `def undeploy` to `def stop` and `def deploy` to `def run`. Making it a bit more modular for new providers to be added. 

Perhaps a `def check_host` or something similar could help

atomicapp/utils.py
```diff
@@ -114,6 +115,45 @@ def __init__(self, target_path, workdir=None):
...
+            response = requests.get(url, verify=False)
+            if response.status_code == 200:
+                logger.debug("Detected environment: openshift pod")
+                return True
+        except:
+            return False
+        return False
+
+    @staticmethod
+    def get_openshift_api_endpoint_from_env():
```
cdrage: sorry about being nitpicky, but other function calls within Utils are camelCase.

@rtnpro i think you brought this up previously before, what does it say in pep-8 specifically? otherwise, we should change utils to conform to how rtnpro did his refactor and made everything snake_case

kadel: I believe that pep8 says that function names should be snake_case and classes CamelCase.

dustymabe: So I should leave this snake case and we'll have another PR that makes everything in here snake_case ?

atomicapp/cli/main.py
dustymabe: No. I mean "atomicapp run". I'll try to update to make it more clear

atomicapp/constants.py
dustymabe: Did some more thinking about this. I have to change it back and solve it a different way. In the past multiple instances of atomicapp running on a system share /run/lock on the host because /run is mounted into the container. So they would be competing for the same lockfile. 

In this case they would not because /tmp from the host isn't mounted in the container.

atomicapp/nulecule/main.py
dustymabe: The problem is that we don't call the provider code until we know what provider we are using. In this case we actually set that information here.

dustymabe:
Re-worked the commits a bit. Fixed lockfile issue in a different way. Added comments about defaulting to "atomicapp run". 

@cdrage, please review.

atomicapp/nulecule/main.py
cdrage: Yeah, let's change this in another PR, no effects, it's just semantics.

atomicapp/utils.py
cdrage: Yeah, it's just semantics, we'll update that in another PR.

atomicapp/nulecule/main.py
cdrage: @dustymabe Okay, I'm okay for merging, but this needs to be changed in the future as I feel as though it's not suppose to belong there.

cdrage:
@dustymabe 
reviewed, just one comment, otherwise, LGTM :+1:

atomicapp/nulecule/main.py
dustymabe: Right.. I feel like if we add support to kubernetes for this same thing then we'll come up with the right place to "detect" this stuff and build out a framework for it.

Actually I'd like a single place where answers.conf + CLI args + defaults + detected environment overrides are all merged into one. Maybe I'll open an issue for that.

PR #455, opened 2015-12-15
Provider documentation update/re-organization.

dustymabe:


cdrage:
I like the re-organization to different dir's as well as removing the dumb version numbers we had in there before :+1: 

LGTM for merge.

PR #460, opened 2015-12-15
Add persistent storage

cdrage:
__DONE:__

~~Initial work on persistent storage. Work's correctly. Just need to clean-up most of the code.~~

~~It's a Work-In-Progress so excuse the multiple log messages, but you get the gist of how I implemented it.~~

  - ~~Code comments / docs~~
  - ~~More tests!~~

```
Name                        Stmts   Miss  Cover   Missing
---------------------------------------------------------
atomicapp/requirements.py      36      3    92%   40, 43, 46
----------------------------------------------------------------------
Ran 5 tests in 0.009s
```
_tests cover 92% and under `kubernetes.py persistent_storage` function too_

  - ~~Clean up log messages~~

__TODO:__
 - [x] Grep `kubectl pvc` output or something to see if there is actually any persistent volumes yet for Kubernetes. Warn the user if so.
 - [x] Logs are a mess (this will be cleaned up in #134)
 - [ ] Documentation in Nulecule spec or nulecule-library example should be added. 
 - [x] Split into different commits

Notes:
 - For now this will **only** be for Kubernetes persistent storage. OpenShift will be added at a future date (Docker maybe too with the Flocker plugin).
 - Only **creating** persistent storage has been adding, no removing (yet?) we need to discuss this.

### DIFF
--- a/atomicapp/nulecule/base.py
+++ b/atomicapp/nulecule/base.py
@@ -15,8 +15,10 @@ from atomicapp.constants import (APP_ENT_PATH,
                                  PARAMS_KEY,
                                  NAME_KEY,
                                  INHERIT_KEY,
-                                 ARTIFACTS_KEY)
+                                 ARTIFACTS_KEY,
+                                 REQUIREMENTS_KEY)
 from atomicapp.utils import Utils
+from atomicapp.requirements import Requirements
 from atomicapp.nulecule.lib import NuleculeBase
 from atomicapp.nulecule.container import DockerHandler
 from atomicapp.nulecule.exceptions import NuleculeException
@@ -27,12 +29,14 @@ logger = logging.getLogger(__name__)
 
 
 class Nulecule(NuleculeBase):
+
     """
     This represents an application compliant with Nulecule specification.
     A Nulecule instance can have instances of Nulecule and Nulecule as
     components. A Nulecule instance knows everything about itself and its
     componenents, but does not have access to its parent's scope.
     """
+
     def __init__(self, id, specversion, metadata, graph, basepath,
                  requirements=None, params=None, config=None,
                  namespace=GLOBAL_CONF):
@@ -132,6 +136,17 @@ class Nulecule(NuleculeBase):
             None
         """
         provider_key, provider = self.get_provider(provider_key, dryrun)
+
+        # Process preliminary requirements
+        # Pass configuration, path of the app, graph, provider as well as dry-run
+        # for provider init()
+        if REQUIREMENTS_KEY in self.graph[0]:
+            logger.debug("Requirements key detected. Running action.")
+            r = Requirements(self.config, self.basepath, self.graph[0][REQUIREMENTS_KEY],
+                             provider_key, dryrun)
+            r.run()
+
+        # Process components
         for component in self.components:
             component.run(provider_key, dryrun)
 
@@ -152,8 +167,8 @@ class Nulecule(NuleculeBase):
         for component in self.components:
             component.stop(provider_key, dryrun)
 
+    # TODO: NOT YET IMPLEMENTED
     def uninstall(self):
-        # uninstall the Nulecule application
         for component in self.components:
             component.uninstall()
 
@@ -225,6 +240,7 @@ class Nulecule(NuleculeBase):
 
 
 class NuleculeComponent(NuleculeBase):
+
     """
     Represents a component in a Nulecule application. It receives props
     from its parent and can add new props and override props at its local
@@ -232,6 +248,7 @@ class NuleculeComponent(NuleculeBase):
     components, but can request the value of sibling's property from its
     parent.
     """
+
     def __init__(self, name, basepath, source=None, params=None,
                  artifacts=None, config=None):
         super(NuleculeComponent, self).__init__(basepath, params, name)
--- a/atomicapp/providers/kubernetes.py
+++ b/atomicapp/providers/kubernetes.py
@@ -220,6 +220,8 @@ class KubernetesProvider(Provider):
                 (action, self.key))
             return
 
+        self._check_persistent_volumes()
+
         # Get the path of the persistent storage yaml file includes in /external
         # Plug the information from the graph into the persistent storage file
         base_path = os.path.dirname(os.path.realpath(__file__))
@@ -239,3 +241,14 @@ class KubernetesProvider(Provider):
                 cmd.append("--kubeconfig=%s" % self.config_file)
             self._call(cmd)
             os.unlink(tmp_file)
+
+    def _check_persistent_volumes(self):
+            cmd = [self.kubectl, "get", "pv"]
+            if self.config_file:
+                cmd.append("--kubeconfig=%s" % self.config_file)
+            lines = self._call(cmd)
+
+            # If there are no persistent volumes to claim, warn the user
+            if not self.dryrun and len(lines.split("\n")) == 2:
+                logger.warning("No persistent volumes detected in Kubernetes. Volume claim will not "
+                               "initialize unless persistent volumes exist.")
--- a/atomicapp/providers/kubernetes.py
+++ b/atomicapp/providers/kubernetes.py
@@ -20,6 +20,7 @@
 import anymarkup
 import logging
 import os
+from string import Template
 
 from atomicapp.plugin import Provider, ProviderFailedException
 from atomicapp.utils import printErrorStatus, Utils
@@ -204,3 +205,37 @@ class KubernetesProvider(Provider):
             if self.config_file:
                 cmd.append("--kubeconfig=%s" % self.config_file)
             self._call(cmd)
+
+    def persistent_storage(self, graph, action):
+        """
+        Actions are either: run, stop or uninstall as per the Requirements class
+        Curently run is the only function implemented for k8s persistent storage
+        """
+
+        logger.debug("Persistent storage enabled! Running action: %s" % action)
+
+        if action not in ['run']:
+            logger.warning(
+                "%s action is not available for provider %s. Doing nothing." %
+                (action, self.key))
+            return
+
+        # Get the path of the persistent storage yaml file includes in /external
+        # Plug the information from the graph into the persistent storage file
+        base_path = os.path.dirname(os.path.realpath(__file__))
+        template_path = os.path.join(base_path,
+                                     'external/kubernetes/persistent_storage.yaml')
+        with open(template_path, 'r') as f:
+            content = f.read()
+        template = Template(content)
+        rendered_template = template.safe_substitute(graph)
+
+        tmp_file = Utils.getTmpFile(rendered_template, '.yaml')
+
+        # Pass the .yaml file and execute
+        if action is "run":
+            cmd = [self.kubectl, "create", "-f", tmp_file, "--namespace=%s" % self.namespace]
+            if self.config_file:
+                cmd.append("--kubeconfig=%s" % self.config_file)
+            self._call(cmd)
+            os.unlink(tmp_file)
--- a/atomicapp/requirements.py
+++ b/atomicapp/requirements.py
@@ -1,0 +1,79 @@
+import logging
+
+from atomicapp.constants import REQUIREMENT_FUNCTIONS
+from atomicapp.plugin import Plugin
+
+logger = logging.getLogger(__name__)
+
+
+class Requirements:
+
+    """
+    Requirements will search what is currently being used under
+    the requirements section of Nulecule and deploy in accordance
+    to the graph variables as well as whether or not said requirement
+    exists within the provider.
+
+    The REQUIREMENTS_FUNCTIONS dictionary maps all current requirement
+    names to the function names for each provider.
+
+    For example, the persistentVolume requirement in Nulecule is mapped as
+    the persistent_storage function within each provider.
+
+    Requirements tries to be as modular as possible.
+    """
+
+    def __init__(self, config, basepath, graph, provider, dryrun):
+        self.plugin = Plugin()
+        self.plugin.load_plugins()
+
+        self.config = config
+        self.basepath = basepath
+        self.graph = graph
+        self.dryrun = dryrun
+
+        # We initialize the provider in order to gather provider-specific
+        # information
+        p = self.plugin.getProvider(provider)
+        self.provider = p(config, basepath, dryrun)
+        self.provider.init()
+
+    def run(self):
+        self._exec("run")
+
+    def stop(self):
+        self._exec("stop")
+
+    def uninstall(self):
+        self._exec("uninstall")
+
+    # Find if the requirement does not exist within REQUIREMENT_FUNCTIONS
+    def _find_requirement_function_name(self, key):
+        logging.debug("Checking if %s matches any of %s" %
+                      (key, REQUIREMENT_FUNCTIONS))
+        if key in REQUIREMENT_FUNCTIONS.keys():
+            return REQUIREMENT_FUNCTIONS[key]
+        raise RequirementFailedException("Requirement %s does not exist." % key)
+
+    # We loop through the given requirements graph and
+    # execute each passed requirement
+    def _exec(self, action):
+        for req in self.graph:
+            key_name = req.keys()[0]
+            requirement_function = self._find_requirement_function_name(key_name)
+
+            # Check to see if the function exists in the provider,
+            # if it does not: fail
+            try:
+                requirement = getattr(self.provider, requirement_function)
+            except AttributeError:
+                raise RequirementFailedException(
+                    "Requirement %s does not exist within %s." %
+                    (requirement_function, self.provider))
+
+            # Run the requirement function
+            requirement(req[key_name], action)
+
+
+class RequirementFailedException(Exception):
+    pass
--- a/atomicapp/utils.py
+++ b/atomicapp/utils.py
@@ -186,6 +186,15 @@ class Utils(object):
     def getTmpAppDir(self):
         return os.path.join(self.tmpdir, APP_ENT_PATH)
 
+    # Create a temporary file with data in order to use kubectl, openshift, etc.
+    # Returns the tmp dir and name of the file
+    @staticmethod
+    def getTmpFile(data, suffix=''):
+        f = tempfile.NamedTemporaryFile(delete=False, suffix=suffix)
+        f.write(data)
+        f.close()
+        return f.name
+
     @staticmethod
     def getComponentName(graph_item):
         # logger.debug("Getting name for %s", graph_item)
--- a/setup.py
+++ b/setup.py
@@ -40,16 +40,18 @@ def _install_requirements():
     return requirements
 
 setup(
-    name = 'atomicapp',
-    version = '0.3.0',
-    description = 'A tool to install and run Nulecule apps',
-    author = 'Red Hat, Inc.',
-    author_email = 'container-tools@redhat.com',
-    url = 'https://github.com/projectatomic/atomicapp',
-    license = "LGPL3",
-    entry_points = {
+    name='atomicapp',
+    version='0.3.0',
+    description='A tool to install and run Nulecule apps',
+    author='Red Hat, Inc.',
+    author_email='container-tools@redhat.com',
+    url='https://github.com/projectatomic/atomicapp',
+    license="LGPL3",
+    entry_points={
         'console_scripts': ['atomicapp=atomicapp.cli.main:main'],
     },
-    packages = find_packages(),
-    install_requires = _install_requirements()
+    packages=find_packages(),
+    package_data={'atomicapp': ['providers/external/kubernetes/*.yaml']},
+    include_package_data=True,
+    install_requires=_install_requirements()
 )
--- a/tests/units/persistent_storage/test_ps.py
+++ b/tests/units/persistent_storage/test_ps.py
@@ -1,0 +1,36 @@
+import unittest
+import tempfile
+from atomicapp.requirements import Requirements, RequirementFailedException
+
+
+class TestPersistentStorage(unittest.TestCase):
+
+    # Create a tmp dir and add graph and config information
+
+    def setUp(self):
+        config = {'helloapache-app': {'image': 'centos/httpd', 'hostport': 80},
+                  'general': {'namespace': 'default', 'provider': 'kubernetes'}}
+        graph = [{'persistentVolume': {'accessMode': 'ReadWrite', 'name': 'var-lib-mongodb-data', 'size': 4}},
+                 {'persistentVolume': {'accessMode': 'ReadWrite', 'name': 'var-log-mongodb', 'size': 4}}]
+        self.tmpdir = tempfile.mkdtemp(prefix="atomicapp-test", dir="/tmp")
+        self.test = Requirements(
+            config=config, basepath=self.tmpdir, graph=graph, provider="kubernetes", dryrun=True)
+
+    def tearDown(self):
+        pass
+
+    def test_missing_requirement(self):
+        with self.assertRaises(RequirementFailedException) as context:
+            self.test._find_requirement_function_name('foo')
+
+        self.assertTrue("Requirement foo does not exist." in context.exception)
+
+    # Rest of these tests checks to see if dryrunning the provider passes correctly
+    def test_run(self):
+        self.test.run()
+
+    def test_stop(self):
+        self.test.stop()
+
+    def test_uninstall(self):
+        self.test.uninstall()
--- a/atomicapp/constants.py
+++ b/atomicapp/constants.py
@@ -40,6 +40,12 @@ NAME_KEY = "name"
 DEFAULTNAME_KEY = "default"
 PROVIDER_KEY = "provider"
 NAMESPACE_KEY = "namespace"
+REQUIREMENTS_KEY = "requirements"
+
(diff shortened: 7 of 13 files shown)


atomicapp/providers/kubernetes.py
```diff
@@ -204,3 +205,36 @@ def undeploy(self):
...
+        # Must write persistent storage to a tmp file as to not have malicious intent
+        # when running direct cmd
+        tmp_file = Utils.getTmpFile(rendered_template, '.yaml')
+
+        # Pass the .yaml file and execute
+        if action is "run":
+            cmd = [self.kubectl, "create", "-f", tmp_file, "--namespace=%s" % self.namespace]
+            if self.config_file:
+                cmd.append("--kubeconfig=%s" % self.config_file)
+            self._call(cmd)
```
cdrage: should we add deleting / removing persistent storage or just running?

setup.py
```diff
@@ -40,16 +40,18 @@ def _install_requirements():
...
-    license = "LGPL3",
-    entry_points = {
+    name='atomicapp',
+    version='0.3.0',
+    description='A tool to install and run Nulecule apps',
+    author='Red Hat, Inc.',
+    author_email='container-tools@redhat.com',
+    url='https://github.com/projectatomic/atomicapp',
+    license="LGPL3",
+    entry_points={
```
cdrage: all stuff above is flake8 fixes ^^

cdrage:
Oh, another note, logging is starting to get SO MESSY. Way too much logging coming out now when running each-application, this will be fixed in #134

dustymabe:
Ok.. ran through it once.. not sure I quite understand it all just yet.. still digesting. I think this would be better/easier to understand if you broke up the big commit into smaller commits with good commit messages. Let's chat later today or next week about the architecture

atomicapp/nulecule/base.py
```diff
@@ -132,6 +136,14 @@ def run(self, provider_key=None, dryrun=False):
             None
         """
         provider_key, provider = self.get_provider(provider_key, dryrun)
+
+        # Process preliminary requirements
+        # Pass configuration, path of the app, graph, provider as well as dry-run for provider init()
+        if REQUIREMENTS_KEY in self.graph[0]:
+            logger.debug("Requirements key detected. Running action.")
+            Requirements(self.config, self.basepath, self.graph[0][REQUIREMENTS_KEY], provider_key, dryrun).run()
```
cdrage: should be requirements = Requirements(etc...)
requirements.run()

atomicapp/providers/kubernetes.py
```diff
@@ -204,3 +205,36 @@ def undeploy(self):
...
                 cmd.append("--kubeconfig=%s" % self.config_file)
             self._call(cmd)
+
+    """
+    All extra methods that belong to 'requirements'
+    Each method will be passed the respective graph from Nulecule
+    and any other extra information that is required.
+
+    Actions are either: run, stop or uninstall
+    """
```
cdrage: move these comments to inside function

atomicapp/requirements.py
```diff
@@ -0,0 +1,74 @@
...
+
+class Requirements:
+
+    """
+    Requirements will search what is currently being used under
+    the requirements section of Nulecule and deploy it in according
+    to the graph variables as well as whether or not said requirement
+    exists within the provider.
+
+    The variable REQUIREMENTS_FUNCTIONS lists all current function names
```
cdrage: should be "the requirements_functions dictionary maps all current requirement names to the function names for each provider"

atomicapp/requirements.py
```diff
@@ -0,0 +1,74 @@
...
+    designated for each provider.
+
+    For example, persistentVolume in Nulecule is noted as the function
```
cdrage: "For example, the persistentVolume requirement in Nulecule is mapped as the persistent_storage function within each provider"

atomicapp/requirements.py
```diff
@@ -0,0 +1,74 @@
...
+    def run(self):
+        self._exec("run")
+
+    def stop(self):
+        self._exec("stop")
+
+    def uninstall(self):
+        self._exec("uninstall")
+
+    # Find if the requirement exists within REQUIREMENT_FUNCTIONS
```
cdrage: "does not exist"

atomicapp/requirements.py
```diff
@@ -0,0 +1,74 @@
...
+    def _find_requirement_name(self, key):
```
cdrage: "find_requirement_function_name"

atomicapp/requirements.py
```diff
@@ -0,0 +1,74 @@
...
+        raise RequirementFailedException("Requirement %s does not exist." % key)
+
+    # Get the attribute from the provider and execute. If it does not exist, fail Pythonesque style
+    def _exec_requirement(self, function):
+        try:
+            return getattr(self.provider, function)
+        except AttributeError:
+            raise RequirementFailedException("Requirement %s does not exist within %s." % (function, self.provider))
+
+    # We loop through the given requirements graph and execute each passed requirement
```
cdrage: "and execute the corresponding provider function for each passed requirement"

tests/units/persistent_storage/test_examples/ps-helloapache/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM projectatomic/atomicapp:0.1.10
+
+MAINTAINER Aaron Weitekamp <aweiteka@redhat.com>
+
+LABEL io.projectatomic.nulecule.providers="kubernetes,docker" \
+      io.projectatomic.nulecule.specversion="0.0.2"
+
+ADD /Nulecule /Dockerfile README.md /application-entity/
```
cdrage: why do we even use dockerfile

atomicapp/requirements.py
```diff
@@ -0,0 +1,74 @@
...
+        if key in REQUIREMENT_FUNCTIONS.keys():
+            return REQUIREMENT_FUNCTIONS[key]
+        raise RequirementFailedException("Requirement %s does not exist." % key)
+
+    # Get the attribute from the provider and execute. If it does not exist, fail Pythonesque style
+    def _exec_requirement(self, function):
+        try:
+            return getattr(self.provider, function)
+        except AttributeError:
+            raise RequirementFailedException("Requirement %s does not exist within %s." % (function, self.provider))
```
cdrage: this function should be squashed into the one below

cdrage:
Merged in changes from @dustymabe 's suggestions :) Let's split this into different commit's and have a look at this after the weekend

cdrage:
Okay, so I've split the commits as well as added checking `kubectl get pv` to see if any persistent volumes actually exist in order to claim them (with persistentVolumeClaim).

The only thing that's left is documentation to be added.

We will need a discussion with @bexelbie about this. With a new feature like this added, it's hard for users to see the significant changes to atomicapp if we only update the Nulecule spec. A detailed documentation that lists all Nulecule functions in atomicapp should be here. I can volunteer to help add this.

atomicapp/nulecule/base.py
```diff
@@ -132,6 +136,15 @@ def run(self, provider_key=None, dryrun=False):
             None
         """
         provider_key, provider = self.get_provider(provider_key, dryrun)
+
+        # Process preliminary requirements
+        # Pass configuration, path of the app, graph, provider as well as dry-run for provider init()
+        if REQUIREMENTS_KEY in self.graph[0]:
+            logger.debug("Requirements key detected. Running action.")
+            r = Requirements(self.config, self.basepath, self.graph[0][REQUIREMENTS_KEY], provider_key, dryrun)
```
dustymabe: Split up this line.. too long

atomicapp/nulecule/base.py
```diff
@@ -148,12 +161,11 @@ def stop(self, provider_key=None, dryrun=False):
             None
         """
         provider_key, provider = self.get_provider(provider_key, dryrun)
-        # stop the Nulecule application
```
dustymabe: why remove this comment?

atomicapp/requirements.py
```diff
@@ -0,0 +1,75 @@
...
+plugin.load_plugins()
+
+logger = logging.getLogger(__name__)
+
+
+class Requirements:
+
+    """
+    Requirements will search what is currently being used under
+    the requirements section of Nulecule and deploy it in according
```
dustymabe: "it in according" -> Probably not what you meant to say

dustymabe: in accordance ?

atomicapp/requirements.py
```diff
@@ -0,0 +1,75 @@
...
+    Requirements tries to be as modular as possible.
+    """
+
+    def __init__(self, config, basepath, graph, provider, dryrun):
+        self.config = config
+        self.basepath = basepath
+        self.graph = graph
+        self.dryrun = dryrun
+
+        # We initialize the provider in order to gather essential information such as kubectl config
```
dustymabe: 100 chars long.. let's keep to under 80

atomicapp/requirements.py
```diff
@@ -0,0 +1,75 @@
...
+        self.provider = plugin.getProvider(provider)(config, basepath, dryrun)
```
dustymabe: Would rather break this up so it isn't two chained function calls. In nulecule/lib.py we do it like this:

```
provider_class = self.plugin.getProvider(provider_key)
provider_class(self.get_context(), self.basepath, dry)
```

atomicapp/requirements.py
```diff
@@ -0,0 +1,75 @@
...
+        self._exec("run")
+
+    def stop(self):
+        self._exec("stop")
+
+    def uninstall(self):
+        self._exec("uninstall")
+
+    # Find if the requirement does not exist within REQUIREMENT_FUNCTIONS
+    def _find_requirement_name(self, key):
```
dustymabe: I think we had discussed renaming to `_find_requirement_function_name`

atomicapp/requirements.py
```diff
@@ -0,0 +1,75 @@
...
+        logging.debug("Checking if %s matches any of %s" % (key, REQUIREMENT_FUNCTIONS))
```
dustymabe: long line

atomicapp/requirements.py
```diff
@@ -0,0 +1,75 @@
...
+        if key in REQUIREMENT_FUNCTIONS.keys():
+            return REQUIREMENT_FUNCTIONS[key]
+        raise RequirementFailedException("Requirement %s does not exist." % key)
+
+    # We loop through the given requirements graph and
+    # execute each passed requirement
+    def _exec(self, action):
+        for req in self.graph:
+            key_name = req.keys()[0]
+            requirement_name = self._find_requirement_name(key_name)
```
dustymabe: requirement_name -> requirement_function would be better I think

atomicapp/requirements.py
```diff
@@ -0,0 +1,75 @@
...
+            except AttributeError:
+                raise RequirementFailedException(
+                    "Requirement %s does not exist within %s." % (requirement_name, self.provider))
+
+            # Run the requirement function
+            requirement_function(req[key_name], action)
+
+
+class RequirementFailedException(Exception):
+
```
dustymabe: do we not need to put "pass" here like we do for the other exceptions we have defined?

atomicapp/requirements.py
```diff
@@ -0,0 +1,75 @@
...
+        for req in self.graph:
+            key_name = req.keys()[0]
+            requirement_name = self._find_requirement_name(key_name)
+
+            # Check to see if the function exists in the provider, if it does not: fail
+            try:
+                requirement_function = getattr(self.provider, requirement_name)
+            except AttributeError:
+                raise RequirementFailedException(
+                    "Requirement %s does not exist within %s." % (requirement_name, self.provider))
```
dustymabe: long line

atomicapp/providers/kubernetes.py
```diff
@@ -204,3 +205,52 @@ def undeploy(self):
             if self.config_file:
                 cmd.append("--kubeconfig=%s" % self.config_file)
             self._call(cmd)
+
+    def persistent_storage(self, graph, action):
+        """
+        All extra methods that belong to 'requirements'
+        Each method will be passed the respective graph from Nulecule
+        and any other extra information that is required.
```
dustymabe: I feel like this description doesn't apply to this function. Are we trying to generically define what requirements functions are? If so I think we need to be more specific or something.

atomicapp/providers/kubernetes.py
```diff
@@ -204,3 +205,52 @@ def undeploy(self):
...
+        """
+
+        logger.debug("Persistent storage enabled! Running action: %s" % action)
+
+        self._check_persistent_volumes()
+
+        # Get the path of the persistent storage yaml file includes in /external
+        # Plug the information from the graph into the persistent storage file
+        base_path = os.path.dirname(os.path.realpath(__file__))
+        template_path = os.path.join(base_path, 'external/kubernetes/persistent_storage.yaml')
```
dustymabe: long line

atomicapp/providers/kubernetes.py
```diff
@@ -204,3 +205,52 @@ def undeploy(self):
...
+        with open(template_path, 'r') as f:
+            content = f.read()
+            template = Template(content)
+            rendered_template = template.safe_substitute(graph)
```
dustymabe: left shift the last two lines. They don't need to be within the "with open"

atomicapp/providers/kubernetes.py
```diff
@@ -204,3 +205,52 @@ def undeploy(self):
...
+
+        # Must write persistent storage to a tmp file as to not have malicious intent
+        # when running direct cmd
```
dustymabe: Change this comment.. As we discussed on friday I don't think there is any chance of malicious intent.. Maybe just change it to "Write out to temporary file to pass to kubectl"

atomicapp/providers/kubernetes.py
```diff
@@ -204,3 +205,52 @@ def undeploy(self):
...
+
+        # Pass the .yaml file and execute
+        if action is "run":
+            cmd = [self.kubectl, "create", "-f", tmp_file, "--namespace=%s" % self.namespace]
+            if self.config_file:
+                cmd.append("--kubeconfig=%s" % self.config_file)
+            self._call(cmd)
+        else:
+            logger.warning(
+                "%s action is not available for provider %s. Doing nothing." % (action, self.key))
```
dustymabe: we should probably do this "check" at the beginning of the function. If the action is not "run" then we just read from a file, did variable replacement, and wrote to a new temporary file for no reason.

atomicapp/providers/kubernetes.py
dustymabe: I guess for now we can stick with this and revisit when we do uninstall.

atomicapp/utils.py
```diff
@@ -186,6 +186,15 @@ def getExternalAppDir(self, component):
     def getTmpAppDir(self):
         return os.path.join(self.tmpdir, APP_ENT_PATH)
 
+    # Create a temporary file with data in order to use kubectl / openshift / docker-compose, etc.
```
dustymabe: long line

atomicapp/providers/kubernetes.py
```diff
@@ -204,3 +205,52 @@ def undeploy(self):
...
+        base_path = os.path.dirname(os.path.realpath(__file__))
+        template_path = os.path.join(base_path, 'external/kubernetes/persistent_storage.yaml')
+        with open(template_path, 'r') as f:
+            content = f.read()
+            template = Template(content)
+            rendered_template = template.safe_substitute(graph)
+
+        # Must write persistent storage to a tmp file as to not have malicious intent
+        # when running direct cmd
+        tmp_file = Utils.getTmpFile(rendered_template, '.yaml')
```
dustymabe: os.unlink file after the _call() below.

tests/units/persistent_storage/test_ps.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def test_missing_requirement(self):
+        with self.assertRaises(RequirementFailedException) as context:
+            self.test._find_requirement_name('foo')
+
+        self.assertTrue("Requirement foo does not exist." in context.exception)
+
+    # Rest of these tests checks to see if dryrunning the provider passes correctly
+    def test_run(self):
+        self.test._exec("run")
```
dustymabe: why not call `self.test.run()` instead?

atomicapp/nulecule/base.py
cdrage: I think that the code explains itself here when stopping (since it's the exact same as within `def run`

atomicapp/providers/kubernetes.py
```diff
@@ -204,3 +205,52 @@ def undeploy(self):
...
+
+    def _check_persistent_volumes(self):
+            cmd = [self.kubectl, "get", "pv"]
+            if self.config_file:
+                cmd.append("--kubeconfig=%s" % self.config_file)
+            lines = self._call(cmd)
+
+            # If there are no persistent volumes to claim, warn the user
+            if not self.dryrun and len(lines.split("\n")) == 2:
+                logger.warning("No persistent volumes detected in Kubernetes. Volume claim will not "
```
dustymabe: long line

atomicapp/providers/kubernetes.py
```diff
@@ -204,3 +205,52 @@ def undeploy(self):
...
+                               "initialize unless persistent volumes exist.")
```
dustymabe: are we just warning here? should we error instead?

atomicapp/providers/kubernetes.py
cdrage: Yeah, this is an old comment. Removing. Thanks for the catch!

atomicapp/providers/kubernetes.py
cdrage: Hmm, how should we do it? Another if-statement at the top of the function?

tests/units/persistent_storage/test_ps.py
cdrage: We pass an action name to _exec, .run() doesn't exist.

atomicapp/providers/kubernetes.py
cdrage: fixed ^^

atomicapp/providers/kubernetes.py
cdrage: Yes, we should *ONLY* warn as it goes against Kubernetes principle, the user may add a persistentVolume to kubernetes after deployment and Kubernetes will change the container states.

atomicapp/providers/kubernetes.py
cdrage: ping @dustymabe how should we check before running this? another if statement at the beginning of the function to check what actions are available?

dustymabe: Right.. I would move "checks" to the beginning of the function. No reason to do work if we fail the checks

tests/units/persistent_storage/test_ps.py
dustymabe:
```
    def run(self):
        self._exec("run")
```

looks like it passes "run" to me.

cdrage: ah, you're right, I'll change it, was mixing it up with the above code that finds the function name.

atomicapp/providers/kubernetes.py
```diff
@@ -204,3 +205,52 @@ def undeploy(self):
...
+        logger.debug("Persistent storage enabled! Running action: %s" % action)
+
+        if action is "run":
+            pass
+        else:
+            logger.warning(
+                "%s action is not available for provider %s. Doing nothing." %
+                (action, self.key))
+            return
+
```
cdrage: @dustymabe like this? don't see any other way since each cmd call will be different for delete / create / etc. when more functions are added.

dustymabe: Would rather do:
```
if action not in ['run']:
    logger.warning()
```

In the future when we add more it would be:
```
if action not in ['run', 'stop', etc..]:
    logger.warning()
```

cdrage:
Tested with changes and can confirm functionality:
```
dropbox/dev/atomicapp  persistent-storage ✗                                                                                                                                                                                                        22h56m ◒  ⍉
▶ kubectl get pv
NAME      LABELS    CAPACITY   ACCESSMODES   STATUS    CLAIM     REASON

dropbox/dev/atomicapp  persistent-storage ✗                                                                                                                                                                                                         22h56m ◒  
▶ kubectl get pvc
NAME                   LABELS    STATUS    VOLUME
var-lib-mongodb-data   map[]     Pending   
var-log-mongodb        map[]     Pending   

dropbox/dev/atomicapp  persistent-storage ✗                                                                                                                                                                                                         22h56m ◒  
▶ kubectl get po
NAME                   READY     STATUS                                                RESTARTS   AGE
helloapache            0/1       Image: centos/httpd is ready, container is creating   0          1m
k8s-master-127.0.0.1   3/3       Running                                               5          1m

```

atomicapp/providers/kubernetes.py
cdrage: updated ^^

atomicapp/requirements.py
```diff
@@ -0,0 +1,79 @@
...
+        self.basepath = basepath
+        self.graph = graph
+        self.dryrun = dryrun
+
+        # We initialize the provider in order to gather provider-specific
+        # information
+        p = plugin.getProvider(provider)
+        self.provider = p(config, basepath, dryrun)
+        self.provider.init()
+
```
cdrage: ^^

dustymabe:
@goern @aweiteka. we would really like to get your input on this before we merge. If you are available today or tomorrow before the holidays that would be great.

atomicapp/requirements.py
```diff
@@ -0,0 +1,79 @@
+import logging
+
+from atomicapp.constants import REQUIREMENT_FUNCTIONS
+from atomicapp.plugin import Plugin
+
+plugin = Plugin()
+plugin.load_plugins()
```
dustymabe: move to inside of __init__

aweiteka:
Sorry I haven't been following this from the start. Question: where is the volume supposed to be mounted so the application can actually use it? In my experience if persistent storage is needed it's defined in a template and referenced in the pod spec so it's actually mounted. But to specify all of that in Nulecule feels like we're duplicating kubernetes.

I think the [spec](https://github.com/projectatomic/nulecule/tree/master/spec#storage-requirements-object) is lacking here but this looks like a good start to implement what we have. LGTM

goern:
@aweiteka I think you got the approach right: the Nulecule Specification is making a statement about what type of storage an application needs, Atomic App is going to auto generate a Kubernetes/OpenShift snippet for that and the application can use it. We can also keep the storage requirements within the  artifacts of the provider, but than we will obfuscate the application, eg the deployment guy will not be able to parameterize the size or type of the storage.

Furthermore, what is the spec missing wrt a storage requirement?

dustymabe:
considering @aweiteka's new comment in https://github.com/projectatomic/nulecule/issues/193 

*the application artefact will define how it wants to use the volume, where it should be mounted. We don't need to provide a path in the spec.* 

I think we should move forward with this review. @aweiteka @goern can you please review so that we can get this merged?

atomicapp/providers/kubernetes.py
```diff
@@ -204,3 +205,50 @@ def undeploy(self):
...
+            self._call(cmd)
+            os.unlink(tmp_file)
+
+    def _check_persistent_volumes(self):
+            cmd = [self.kubectl, "get", "pv"]
+            if self.config_file:
+                cmd.append("--kubeconfig=%s" % self.config_file)
+            lines = self._call(cmd)
+
+            # If there are no persistent volumes to claim, warn the user
```
aweiteka: There are more error cases. I can think of two:

1. claim with that name has already been created.
1. user is claiming more space than allowed.

These (and other errors like this I can't think of) can be generally caught with one check, I would think.

atomicapp/providers/kubernetes.py
aweiteka: Given this is "persistent" we probably don't want to ever be deleting. I believe OpenShift does not delete storage when all objects are destroyed. It's treated *special* so as to do no unexpected harm.

aweiteka:
Nice work! LGTM.

atomicapp/nulecule/base.py
vpavlin: It's not nice to make changes unrelated to the PR:)

atomicapp/providers/kubernetes.py
cdrage: Ah yes, this should error out atomically though when calling the API. But it would be nice to parse and create a good logging output for this :)

atomicapp/nulecule/base.py
cdrage: :stuck_out_tongue:

cdrage:
^^ changed code from @vpavlin 's comment :) I'll update notes / comments in a different PR for changes unrelated to persistent storage.

dustymabe:
@cdrage, since there has been a lot of work in master can we get a rebase on it. Might be good to test on top of master anyway.

cdrage:
Rebased and checks pass!

dustymabe:
With @aweiteka LGTM and a LGTM from me.. merging.

PR #463, opened 2015-12-16
move openshift._make_request() to Utils.make_rest_request()

kadel:
Other providers can make use of this function now.

### DIFF
--- a/atomicapp/providers/openshift.py
+++ b/atomicapp/providers/openshift.py
@@ -19,8 +19,8 @@
 
 import os
 import anymarkup
-import requests
 from urlparse import urljoin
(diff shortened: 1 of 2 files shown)


kadel:
Hmm :-( Now I not sure if this is great idea.
I cannot rise `ProviderFailedException` here :-(

Whole point of this function was so I don't have to catch exceptions each time i want to make request.

kadel:
I had to change raising `ProviderFailedException` to `AtomicAppUtilsException`.

cdrage:
LGTM

dustymabe:
LGTM

PR #465, opened 2015-12-17
Openshift unittests

rtnpro:
Tests and related refactor for openshift provider using Openshift API.

### DIFF
--- a/atomicapp/providers/openshift.py
+++ b/atomicapp/providers/openshift.py
@@ -338,15 +338,29 @@ class OpenShiftProvider(Provider):
             user:
                 token: abcdefghijklmnopqrstuvwxyz0123456789ABCDEF
         """
-
-        url = None
-        token = None
-        namespace = None
-
         logger.debug("Parsing %s", filename)
 
         with open(filename, 'r') as fp:
             kubecfg = anymarkup.parse(fp.read())
+
+        try:
+            return self._parse_kubeconf_data(kubecfg)
+        except ProviderFailedException:
+            raise ProviderFailedException('Invalid %s' % filename)
+
+    def _parse_kubeconf_data(self, kubecfg):
+        """
+        Parse kubeconf data.
+
+        Args:
+            kubecfg (dict): Kubernetes config data
+
+        Returns:
+            A tuple: (url, token, namespace)
+        """
+        url = None
+        token = None
+        namespace = None
 
         current_context = kubecfg["current-context"]
 
@@ -356,6 +370,9 @@ class OpenShiftProvider(Provider):
         for co in kubecfg["contexts"]:
             if co["name"] == current_context:
                 context = co
+
+        if not context:
+            raise ProviderFailedException()
 
         cluster = None
         for cl in kubecfg["clusters"]:
@@ -367,8 +384,8 @@ class OpenShiftProvider(Provider):
             if usr["name"] == context["context"]["user"]:
                 user = usr
 
-        if not context or not cluster or not user:
-            raise ProviderFailedException("Invalid %s", filename)
+        if not cluster or not user:
+            raise ProviderFailedException()
 
         logger.debug("context: %s", context)
         logger.debug("cluster: %s", cluster)
--- a/atomicapp/providers/openshift.py
+++ b/atomicapp/providers/openshift.py
@@ -186,35 +186,46 @@ class OpenShiftProvider(Provider):
             data = None
             with open(os.path.join(self.path, artifact), "r") as fp:
                 data = anymarkup.parse(fp, force_types=None)
-            # kind has to be specified in artifact
-            if "kind" not in data.keys():
-                raise ProviderFailedException(
-                    "Error processing %s artifact. There is no kind" % artifact)
 
-            kind = data["kind"].lower()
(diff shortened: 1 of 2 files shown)


atomicapp/providers/openshift.py
```diff
@@ -179,15 +266,7 @@ def _process_template(self, template):
...
-        (status_code, return_data) = self._make_request("post", url, data=template)
-        if status_code == 201:
-            logger.info("template proccessed %s", template['metadata']['name'])
-            logger.debug("processed template %s", return_data)
-            return return_data['objects']
-        else:
-            msg = "%s %s" % (status_code, return_data)
-            logger.error(msg)
-            raise ProviderFailedException(msg)
+        return self.oc.process_template(url, template)
```
kadel: Is `OpenShiftProvider._process_template` still needed?
I would probably   get rid of if this function (`_process_template`) now. It doesn't do anything useful, everything important is in `OpenshiftClient.process_tempate`.

rtnpro: ah! yes

dustymabe:
still WIP or done now?

rtnpro:
@dustymabe 

Yes, currently having 78% coverage. I need to cover a few more methods.

atomicapp/providers/openshift.py
rtnpro: The reason I did not remove ``_process_template`` is because of the dependency on ``get_url``. Also, I see no reason to move ``get_url`` to ``OpenshiftClient``.

atomicapp/providers/openshift.py
```diff
@@ -32,6 +32,85 @@
 logger = logging.getLogger(__name__)
 
 
+class OpenshiftClient(object):
+
+    def __init__(self, openshift_api, kubernetes_api):
+        self.openshift_api = openshift_api
+        self.kubernetes_api = kubernetes_api
+
+    def get_oapi_resources(self):
```
dustymabe: could probably make this function generic:

get_api_resources(self, type) where type = 'kubernetes' or 'openshift'

don't have to change it, just sayin

tests/units/providers/test_openshift_provider.py
```diff
@@ -0,0 +1,241 @@
+# -*- coding: utf-8 -*-
```
dustymabe: This file could use some comments about the strategy of the test.

tests/units/providers/test_openshift_provider.py
```diff
@@ -0,0 +1,241 @@
...
+    def test_process_artifact_data_error_resource_not_in_resources(self):
+        artifact_data = {
+            'kind': 'foobar'
+        }
+
+        op = self.get_oc_provider()
+
+        self.assertRaises(
+            ProviderFailedException,
+            op._process_artifact_data, 'foo', artifact_data)
```
dustymabe: should this not be:

`op._process_artifact_data('foo', artifact_data))`

atomicapp/providers/openshift.py
```diff
@@ -294,16 +338,24 @@ def _parse_kubeconf(self, filename):
             user:
                 token: abcdefghijklmnopqrstuvwxyz0123456789ABCDEF
         """
```
dustymabe: Should move this example expected file format to the _parse_kubeconf_data function now

atomicapp/providers/openshift.py
```diff
@@ -313,6 +365,9 @@ def _parse_kubeconf(self, filename):
             if co["name"] == current_context:
                 context = co
 
+        if not context:
+            raise ProviderFailedException()
```
dustymabe: Don't really understand why this change would made. Please change it back if there was not good reason.

atomicapp/providers/openshift.py
rtnpro: yes.

tests/units/providers/test_openshift_provider.py
rtnpro: :P but agreed!

tests/units/providers/test_openshift_provider.py
rtnpro: No, this is how ``self.assertRaises`` is supposed to be called.

``self.assertRaises(ExceptionClass, method, arg1, arg2)``

atomicapp/providers/openshift.py
rtnpro: Yes

atomicapp/providers/openshift.py
rtnpro: without this check, the subsequent code used to break when ``context == None``, as they check for ``context[somekey]``.

tests/units/providers/test_openshift_provider.py
dustymabe: ok

atomicapp/providers/openshift.py
dustymabe: got ya. you're right

It would still be nice to print out the Invalid filename like was done on line 327 before?

dustymabe:
LGTM. @kadel can you +1 and we can merge this

kadel:
:+1: LGTM

PR #466, opened 2015-12-18
Allow user to specify both source and destination as directories.

### LINKED ISSUE #400: Bring back install to destination from Nulecule metadata
# Problem statement

Let's say, I have a local library of Nulecule apps, cloned from https://github.com/projectatomic/nulecule-library. I'd like to install/run an application from the library in my current working directory or workspace. I would not like to manually copy the metadata to my destination directory in order to use it.

dustymabe:
This will allow the user to specify both a source path to a directory
with Nulecule metadata as well as a destination path where they would
like to new application to reside. This doesn't make sense in most
situations but could be useful in cases where we don't want to copy
data out of the container or in the case rtnpro stated in #400.

Closes #400.

### DIFF
--- a/atomicapp/nulecule/main.py
+++ b/atomicapp/nulecule/main.py
@@ -61,10 +61,11 @@ class NuleculeManager(object):
         else:
             self.app_path = app_spec
 
-        # Doesn't make sense to provide an app path and destination
+        # Doesn't really make much sense to provide an app path and destination,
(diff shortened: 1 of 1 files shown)


dustymabe:
@rtnpro ^^

rtnpro:
:+1:

dustymabe:
Merging!!!

PR #468, opened 2015-12-18
Add --mode CLI switch. Argparser reorg. 

dustymabe:
In order to change the mode of the application easily when using
Atomic CLI it would be nice to have a way to change the mode even
though we were run through the "run" action.

This could be useful for accessing actions that aren't available in
Atomic CLI or ones that haven't yet been added to atomic CLI.

For example, if we add 'atomicapp takeoverworld' to Atomic App in
order to use it through Atomic CLI we would simply need to run:
`atomic run <IMAGE> --mode=takeoverworld`.

### DIFF
--- a/atomicapp/cli/main.py
+++ b/atomicapp/cli/main.py
@@ -99,43 +99,54 @@ def cli_stop(args):
 class CLI():
 
     def __init__(self):
-        self.parser = argparse.ArgumentParser(
+        self.parser = self.create_parser()
+
+    def create_parser(self):
+
+        # We will have a few parsers that we use. The toplevel parser
+        # will be the parser that ultimately gets called. It will consist
+        # of subparsers for each "action" and each of those subparsers will
+        # inherit from a parser for all global options.
+
+        # === TOPLEVEL PARSER ===
+        # Create the toplevel parser. This is the one we will return
+        toplevel_parser = argparse.ArgumentParser(
             prog='atomicapp',
+            formatter_class=argparse.RawDescriptionHelpFormatter,
             description=(
                 "This will install and run an Atomic App, "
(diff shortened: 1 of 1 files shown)


dustymabe:
@cdrage @rtnpro let me know what you think. this builds up to the next PR for atomicapp genanswers.

atomicapp/cli/main.py
```diff
@@ -100,43 +99,75 @@ def cli_stop(args):
...
+        globals_parser.add_argument(
             "-q",
             "--quiet",
             dest="quiet",
             default=False,
             action="store_true",
             help="Quiet output mode.")
-
-        base_parser.add_argument(
+        globals_parser.add_argument(
```
rtnpro: Is it intuitive to have ``--mode`` as a global option? I think ``run``ning an image/application with some ``mode`` is good enough.

dustymabe: Probably not, it's only really needed for one path (i.e. run would probably be the only way we ever use it).. the one thing I do like about having it as a global is that it will show in the help when someone types `atomcapp --help`.

I can move it to a non-global if you like.

cdrage:
LGTM :+1: from tests with both PR's.

atomicapp/cli/main.py
rtnpro: @dustymabe 

We don't want ``atomicapp`` users to use mode anyways, right? It's just a hack for integration with ``atomic`` CLI. So, I believe that we can safely move it to ``run`` sub command.

dustymabe:
talked with @rtnpro about this and it would be more complicated than it is worth to have this as a non-global option. merging as-is.

Issue #467, opened 2015-12-18
Add ability to just grab answers.conf from app


dustymabe:
For Atomic CLI I can easily envision people living completely in the following few actions for now:

- request and get answers.conf file -> extraced to current working dir
- edit answers.conf file
- run app and provide answers.conf
- stop app

We need to provide them a way to just grab an answers.conf file without having to do a full install/unpack.

PR #469, opened 2015-12-18
Add genanswers toplevel action.

dustymabe:
See #467

### DIFF
--- a/atomicapp/cli/main.py
+++ b/atomicapp/cli/main.py
@@ -179,7 +179,7 @@ class CLI():
             "--answers-format",
             dest="answers_format",
             default=ANSWERS_FILE_SAMPLE_FORMAT,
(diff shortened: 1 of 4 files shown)


rtnpro:
LGTM :+1: 

What about using ``get-anwers`` in place of ``genanswers``?

dustymabe:
> What about using get-anwers in place of genanswers?

I like "generate answers" because it implies that there is some "work" that is being done by the Atomic App code; it has to get all the artifacts from nested nulecules so that it can generate the answers.conf file. The answers.conf file isn't something we simply copy out, but have to actually generate.
 
That is the way it is now.. With your proposed work on answers for nested nulecules would this change?

cdrage:
LGTM :+1:

dustymabe:
Just performed rebase.. merging now

PR #488, opened 2016-01-08
Stop Docker containers more gracefully

cdrage:
In this PR, we are stopping containers a lot more gracefully than our previous code of `docker kill foobar`.

This change also relates to a bug where if a container is already stopped, Atomic App will error out when running `docker kill`.

Instead, we use `docker stop foobar` which has a 10 second timer for the container to stop peacefully, otherwise, it is killed. Container which are already stopped are ignored by the command.


dustymabe:
I think a 10 second timeout is the default timeout for `stop` right?

cdrage:
@dustymabe 

Wait, it is?!? Blah.

Okay, I changed the PR ^^. We should be using stop regardless.

dustymabe:
agreed. LGTM

Issue #461, opened 2015-12-15
native openshift: support for nested nulecule applications


dustymabe:
Currently nested nulecule applications won't work when running `oc new-app` because the pod that runs the atomicapp code doesn't have access to the docker daemon so it can't pull down other containers. I'm not sure of the best way to solve this right now but we need to support running nested nulecules via `oc new-app`.

rtnpro:
I will take a look into it.

dustymabe:
@rtnpro make sure that, whatever you do, you don't try to interface with the docker registry directly via REST API calls. I already have a POC implementation of that but have been told it's not going to be a solution we are happy with.

dustymabe:
@rtnpro openshift has a `./oc rsync` command. Try it out with `--strategy=tar` and see if that is something we can use.

Basically we would need to start the container (in a pod) and then make it sleep or something. Then we would run the equivalent of the `oc rsync` against the sleeping pod and have it copy the files to the local install pod (the one started with `oc new-app`).

rtnpro:
@dustymabe 

Correct me if I am wrong?

So, if the ``atomicapp`` code doesn't have access to ``docker`` daemon, then no docker commands will work, e.g., ``docker cp``, etc. Or, is it just prohibited to access Docker registries?

If only ``docker pull`` is forbidden, then the solution is pretty simple:
```
def unpack(...):
    if openshift_env:
         call oc run image with dummy entrypoint, and this will pull the image
    # Then others step can follow as usual:
    Run docker container
    Extract contents for docker container... 
```

dustymabe:
we don't have access to the docker daemon at all. no docker commands will work.

Here is an example using just oc commands to grabbing a file from an image:

```
[root@f23 dir]# cat /tmp/httpdsleep.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: sleephttpd
spec:
  containers:
  - image: projectatomic/helloapache
    command:
      - sleep
      - "3600"
    imagePullPolicy: IfNotPresent
    name: sleephttpd
  restartPolicy: Always
[root@f23 dir]# oc create -f /tmp/httpdsleep.yaml                                                                                                                                                    
pod "sleephttpd" created
[root@f23 dir]# 
[root@f23 dir]# oc get pods sleephttpd
NAME         READY     STATUS    RESTARTS   AGE
sleephttpd   1/1       Running   0          1m
[root@f23 dir]# 
[...]

PR #480, opened 2016-01-04
openshift provider: add option for skiping tls verification

kadel:
get tls verification settings from `providerconfig` (kubectl config) or `answers.conf`

closes: #451

PR #489, opened 2016-01-10
native openshift: support nested Nulecule apps

rtnpro:
As discussed in #461, since Openshift does not allow access to host **docker** daemon to containers running inside pods. Currently, we use Docker to pull external Nulecule images and extract metadata from them.

This pull request implements unpacking a Nulecule image on Openshift using Openshift API calls.

### DIFF
--- a/atomicapp/nulecule/base.py
+++ b/atomicapp/nulecule/base.py
@@ -22,6 +22,7 @@ from atomicapp.requirements import Requirements
 from atomicapp.nulecule.lib import NuleculeBase
 from atomicapp.nulecule.container import DockerHandler
 from atomicapp.nulecule.exceptions import NuleculeException
+from atomicapp.providers.openshift import OpenShiftProvider
 
 from jsonpointer import resolve_pointer, set_pointer, JsonPointerException
 
@@ -87,9 +88,17 @@ class Nulecule(NuleculeBase):
             A Nulecule instance, or None in case of dry run.
         """
         logger.info('Unpacking image: %s to %s' % (image, dest))
-        docker_handler = DockerHandler(dryrun=dryrun)
-        docker_handler.pull(image)
-        docker_handler.extract(image, APP_ENT_PATH, dest, update)
+        if Utils.running_on_openshift():
+            # pass general config data containing provider specific data
+            # to Openshift provider
+            op = OpenShiftProvider(config.get('general', {}), './', False)
+            op.artifacts = []
+            op.init()
+            op.extract(image, APP_ENT_PATH, dest, update)
+        else:
+            docker_handler = DockerHandler(dryrun=dryrun)
+            docker_handler.pull(image)
+            docker_handler.extract(image, APP_ENT_PATH, dest, update)
         return cls.load_from_path(
             dest, config=config, namespace=namespace, nodeps=nodeps,
             dryrun=dryrun, update=update)
@@ -216,7 +225,8 @@ class Nulecule(NuleculeBase):
             source = Utils.getSourceImage(node)
             component = NuleculeComponent(
                 node_name, self.basepath, source,
-                node.get(PARAMS_KEY), node.get(ARTIFACTS_KEY))
+                node.get(PARAMS_KEY), node.get(ARTIFACTS_KEY),
+                self.config)
             component.load(nodeps, dryrun)
             components.append(component)
         self.components = components
@@ -257,6 +267,7 @@ class NuleculeComponent(NuleculeBase):
         self.artifacts = artifacts
         self.rendered_artifacts = defaultdict(list)
         self._app = None
+        self.config = config
 
     def load(self, nodeps=False, dryrun=False):
         """
@@ -331,6 +342,7 @@ class NuleculeComponent(NuleculeBase):
             nulecule = Nulecule.unpack(
                 self.source,
                 external_app_path,
+                config=self.config,
                 namespace=self.namespace,
                 dryrun=dryrun,
                 update=update
--- a/atomicapp/providers/openshift.py
+++ b/atomicapp/providers/openshift.py
@@ -231,7 +231,9 @@ class OpenshiftClient(object):
             url,
             on_message=lambda ws, message: self._handle_exec_reply(ws, message, results, outfile))
 
-        ws.run_forever(sslopt={'cert_reqs': ssl.CERT_NONE})
+        ws.run_forever(sslopt={
+            'ca_certs': self.provider_ca,
+            'cert_reqs': ssl.CERT_REQUIRED if self.provider_tls_verify else ssl.CERT_NONE})
 
         if not outfile:
             return ''.join(results)
--- a/atomicapp/providers/openshift.py
+++ b/atomicapp/providers/openshift.py
@@ -17,6 +17,7 @@
  along with Atomic App. If not, see <http://www.gnu.org/licenses/>.
 """
 
+import datetime
 import os
 import anymarkup
 import ssl
@@ -250,6 +251,38 @@ class OpenshiftClient(object):
                 f.write(cleaned_msg)
         else:
             results.append(cleaned_msg)
+
+    def get_pod_status(self, namespace, pod):
+        """
+        Get pod status.
+
+        Args:
+            namespace (str): Openshift namespace
+            pod (str): Pod name
+
+        Returns:
+            Status of pod (str)
+
+        Raises:
+            ProviderFailedException when unable to fetch Pod status.
+        """
+        args = {
+            'namespace': namespace,
+            'pod': pod,
+            'access_token': self.access_token
+        }
+        url = urljoin(
+            self.kubernetes_api,
+            'namespaces/{namespace}/pods/{pod}?'
+            'access_token={access_token}'.format(**args))
+        (status_code, return_data) = \
+            Utils.make_rest_request("get", url, verify=self._requests_tls_verify())
+
+        if status_code != 200:
+            raise ProviderFailedException(
+                'Could not fetch status for pod: {namespace}/{pod}'.format(
+                    namespace=namespace, pod=pod))
+        return return_data['status']['phase'].lower()
 
 
 class OpenShiftProvider(Provider):
@@ -765,11 +798,9 @@ class OpenShiftProvider(Provider):
                 'restartPolicy': 'Always'
             }
         }
+
         self.oc.deploy(self._get_url(self.namespace, 'Pod'), artifact)
-        # FIXME: Pod takes some time to come up, so we wait for 3 seconds
-        # The best solution would be to watch for the pod to come to
-        # the running state
-        time.sleep(10)
+        self._wait_till_pod_runs(self.namespace, pod_name, timeout=300)
 
         # Archive content from the container and dump it to tmpfile
         tmpfile = '/tmp/atomicapp-{pod}.tar.gz'.format(pod=pod_name)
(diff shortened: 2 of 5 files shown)


dustymabe:
Ok I did actually get this to work :) - great work @rtnpro. 

I had to make a few small changes just to get it to work:
- Need to pass the config all the way down to the component level
- Needed to increase the timeout in order to get it to work on the openshift environment I am operating on. 

The patch is attached. [patch.txt](https://github.com/projectatomic/atomicapp/files/85272/patch.txt)

There are other comments I will make in the actual code review of the files.

atomicapp/nulecule/base.py
```diff
@@ -87,9 +88,20 @@ def unpack(cls, image, dest, config=None, namespace=GLOBAL_CONF,
...
-        docker_handler = DockerHandler(dryrun=dryrun)
-        docker_handler.pull(image)
-        docker_handler.extract(image, APP_ENT_PATH, dest, update)
+        if Utils.running_on_openshift():
+            _config = {
+                'providerapi': config['general']['providerapi'],
+                'accesstoken': config['general']['accesstoken'],
+                'namespace': config['general']['namespace']
+            }
+            op = OpenShiftProvider(_config, './', False)
```
dustymabe: Why not pass the whole config here (which already contains the values we need in it) rather than just picking out pieces?

atomicapp/providers/openshift.py
```diff
@@ -133,6 +146,49 @@ def process_template(self, url, template):
...
+            'pod': pod,
+            'container': container,
+            'command': ''.join(['command={}&'.format(word) for word in command.split()])
+        }
+        url = 'wss://' + urljoin(
+            self.kubernetes_api,
+            'namespaces/{namespace}/pods/{pod}/exec?'
+            'access_token={token}&container={container}&'
+            '{command}stdout=1&stdin=0&tty=0'.format(
+              **args)).split('://', 1)[-1]
```
dustymabe: So part of what we are doing here is removing `ttps ://` from `self.kubernetes_api`
Can you comment on what you are doing. Even better may be to comment and break it up into two lines.

atomicapp/providers/openshift.py
```diff
@@ -133,6 +146,49 @@ def process_template(self, url, template):
...
+
+        ws.run_forever(sslopt={'cert_reqs': ssl.CERT_NONE})
+
+        return ''.join(results)
+
+    def _handle_exec_reply(self, ws, message, results):
+        """
+        Handle reply message for exec call
+        """
+        results.append(message[1:])
```
dustymabe: Comment on `[1:]`. I still don't know why we need this but we should at comment. Either we should figure out why it is needed and state why or we should just comment that we don't know why.

atomicapp/providers/openshift.py
```diff
@@ -571,3 +622,71 @@ def _get_config_values(self):
...
+        time.sleep(3)
+
+        # Archive content from the container and return the binary data for
+        # for the archive
+        resp = self.oc.execute(
+            self.namespace, pod_name, container_name,
+            'tar -cz {}'.format('/' + src)
+        )
+        tmpfile = '/tmp/atomicapp-{pod}.tar.gz'.format(pod=pod_name)
+        tmpdir = '/tmp/atomicapp-{pod}'.format(pod=pod_name)
```
dustymabe: Since we are already in a throw-away container I would say that extracting to a tmpdir first rather than just extracting directly to the target directory is not necessary.

atomicapp/providers/openshift.py
```diff
@@ -571,3 +622,71 @@ def _get_config_values(self):
...
+                    }
+                ],
+                'restartPolicy': 'Always'
+            }
+        }
+        self.oc.deploy(self._get_url(self.namespace, 'Pod'), artifact)
+        # FIXME: Pod takes some time to come up, so we wait for 3 seconds
+        # The best solution would be to watch for the pod to come to
+        # the running state
+        time.sleep(3)
```
dustymabe: Yeah. We pretty much have to query for the state of the pod and make sure it is ready in this step. We'll have to add that functionality and update this before we merge.

In my testing I had to up the sleep value because it was failing.

atomicapp/providers/openshift.py
```diff
@@ -133,6 +146,49 @@ def process_template(self, url, template):
...
+        }
+        url = 'wss://' + urljoin(
+            self.kubernetes_api,
+            'namespaces/{namespace}/pods/{pod}/exec?'
+            'access_token={token}&container={container}&'
+            '{command}stdout=1&stdin=0&tty=0'.format(
+              **args)).split('://', 1)[-1]
+
+        results = []
+
```
dustymabe: Probably need to comment on why we needed websockets here.

Basic https won't work because the request needed to be upgraded to SPDY, which python-requests doesn't support. The endpoint does work with websockets though, so we had to take this route.

atomicapp/providers/openshift.py
```diff
@@ -133,6 +146,49 @@ def process_template(self, url, template):
...
+        ws = websocket.WebSocketApp(
```
dustymabe: Add debug statement, including url.

atomicapp/providers/openshift.py
```diff
@@ -571,3 +622,71 @@ def _get_config_values(self):
...
+            }
+        }
+        self.oc.deploy(self._get_url(self.namespace, 'Pod'), artifact)
+        # FIXME: Pod takes some time to come up, so we wait for 3 seconds
+        # The best solution would be to watch for the pod to come to
+        # the running state
+        time.sleep(3)
+
+        # Archive content from the container and return the binary data for
+        # for the archive
```
dustymabe: One potential issue I see here is that all the contents of the web request are loaded into memory and then written out to the file. If for some awful reason we ever had a very large reply to the request then we'd probably run into trouble. Would probably be better to use a buffer and write out data as we receive it so we don't keep it all in memory. 

Since I hope we come up with a better solution than this in the future we may be able to leave it like this for now but I would want it commented at the top of the execute function as well as here.

atomicapp/nulecule/base.py
rtnpro: The reason being providers consume a **flat** config dictionary in contrast to the config data in **answers**. And our method ``get_context`` which does the conversion is an instance method written to create a flat **config** data for components. So, I used the above work around instead.

atomicapp/providers/openshift.py
rtnpro: :+1: will do it

atomicapp/providers/openshift.py
rtnpro: :+1:

atomicapp/providers/openshift.py
rtnpro: So, when we do a ``tar -cz /application-entity``, the tar contents start from ``application-entity`` as root dir. And when extracting, it extracts with ``application-entity`` as root directory, even when trying to extract a single member in the tarfile. I could not figure out a simple command to tar the contents of ``/application-entity`` directory, so that we don't need a tmpdir to copy the content from. But I couldn't.

It will be great if you can suggest some work around.

atomicapp/providers/openshift.py
rtnpro: I need to look into it and see if we can figure out a more elegant way.

atomicapp/providers/openshift.py
rtnpro: I agree :+1:

atomicapp/providers/openshift.py
rtnpro: Ah! I missed that :+1:

atomicapp/providers/openshift.py
rtnpro: Nice point :+1: I will dump things to the file as soon as we receive some message.

dustymabe: maybe we should just comment for now and fix later in a new issue/PR.

atomicapp/providers/openshift.py
dustymabe: Try modifying the tar command to be: `tar -cz  --directory /application-entity ./`

atomicapp/nulecule/base.py
dustymabe: Add a comment to that effect. 

You could also simply just pass `config[GLOBAL_CONF]` directly (if you do, make sure to comment :))

atomicapp/providers/openshift.py
dustymabe: should be pretty easy:

- query api for the state of the pod
- is state of the pod == READY ?
    - No? loop
    - Yes? break

atomicapp/nulecule/base.py
rtnpro: :+1:

atomicapp/providers/openshift.py
rtnpro: Done!

dustymabe:
@kadel - this PR could probably also use a review from you. he will be putting up new commits at some point with changes from my code review, but a quick look from you before and after that happens couldn't hurt.

atomicapp/providers/openshift.py
rtnpro: It worked!

rtnpro:
@dustymabe Updated the PR based on review. Good to go from my side.

atomicapp/providers/openshift.py
```diff
@@ -133,6 +146,66 @@ def process_template(self, url, template):
...
+        url = 'wss://' + url.split('://', 1)[-1]
+        logger.debug('url: {}'.format(url))
+
+        results = []
+
+        ws = websocket.WebSocketApp(
+            url,
+            on_message=lambda ws, message: self._handle_exec_reply(ws, message, results, outfile))
+
+        ws.run_forever(sslopt={'cert_reqs': ssl.CERT_NONE})
```
kadel: after #480 is merged we'll need to update this so it uses `self.provider_ca` and `self.provider_tls_verify`

dustymabe:
@rtnpro looks good for the most part.. the one part that is still missing is polling status of the pod until it is ready for us to exec a command. I don't think we should leave that sleep in there.

It should be pretty easy to make a new faction that gets the status of a pod and gives back the data returned as a dict.. then that information can be used to determine if it is ready or not.. the [v1.pod](https://docs.openshift.org/latest/rest_api/kubernetes_v1.html#v1-pod) object that gets returned has a status object in there.

ALSO.. still waiting to get the other PR merged before you will be able to update this one.

kadel:
There is also `watch` parameter for lot of api calls. Docs says: 
"Watch for changes to the described resources and return them as a stream of add, update, and remove notifications." 
source: https://docs.openshift.org/latest/rest_api/kubernetes_v1.html#list-or-watch-objects-of-kind-pod

But I haven't been able to get to work :-(

rtnpro:
@kadel openshift web ui is also using these watch endpoints over websocket to keep the UI for pods up to date.

dustymabe:
@rtnpro we just merged https://github.com/projectatomic/atomicapp/pull/480 so you can integrate it with this PR now.

atomicapp/providers/openshift.py
dustymabe: yes. @rtnpro can you use those two parameters to set the sslopt correctly?

rtnpro: ok

atomicapp/providers/openshift.py
```diff
@@ -655,3 +757,93 @@ def _set_config_values(self):
...
+        self.oc.deploy(self._get_url(self.namespace, 'Pod'), artifact)
+        self._wait_till_pod_runs(self.namespace, pod_name, timeout=300)
+
+        # Archive content from the container and dump it to tmpfile
+        tmpfile = '/tmp/atomicapp-{pod}.tar.gz'.format(pod=pod_name)
+        self.oc.execute(
+            self.namespace, pod_name, container_name,
+            'tar -cz --directory {} ./'.format('/' + src),
+            outfile=tmpfile
+        )
```
dustymabe: We need to put `_wait_till_pod_runs` and `oc.execute` inside of a try/finally where the finally contains the `oc.delete` of the pod.

dustymabe:
@rtnpro great work.. I had one final comment (see above), otherwise LGTM.

atomicapp/providers/openshift.py
rtnpro: agreed!

atomicapp/providers/openshift.py
```diff
@@ -655,3 +757,94 @@ def _set_config_values(self):
...
+            self.oc.execute(
+                self.namespace, pod_name, container_name,
+                'tar -cz --directory {} ./'.format('/' + src),
+                outfile=tmpfile
+            )
+
+            # Extract archive data
+            tar = tarfile.open(tmpfile, 'r:gz')
+            tar.extractall(dest)
+        finally:
```
dustymabe: The two tar statements don't need to be in the `try` - move them below the `finally` and unindent.

rtnpro: yes!

kadel:
great work. :+1: LGTM

dustymabe:
+1 - merging

PR #490, opened 2016-01-10
cli: Add genanswers as a choice for --mode.

dustymabe:
One of the points of --mode was to be able to switch to the
"genanswers" mode from the atomic CLI. I left it out of the choices
of possible arguments to --mode from the CLI so it was broken. This
fixes that.


rtnpro:
:+1:

PR #496, opened 2016-01-11
Move the install of docker to fix /run/lock issue

### LINKED ISSUE #494: /run/lock doesn't exist by default in fedora:23 docker
/run/lock doesn't exist until after the docker package is installed.

I have opened a bug:

https://fedorahosted.org/rel-eng/ticket/6328

However, I believe we should reorganize the dockerfile to fix this. PR coming.

bexelbie:
Pending a proper fix upstream, this rearranges the build
to look like the git build and by happenstance fixes the missing
/run/lock directory error

Fixes #494


cdrage:
Hey @bexelbie 
So we've decided to deprecate locking due to the numerous issues and edge-cases we were getting with it. I'm closing your PR's!

PR #498, opened 2016-01-11
Update to mount /var/run and /var/lock

bexelbie:
This helps with some OSes that don't map /var/run->/run

Also use /var/lock not /var/run/lock for the same reason

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -12,7 +12,8 @@ RUN yum install -y epel-release && \
     yum install -y --setopt=tsflags=nodocs $(sed s/^/python-/ test-requirements.txt) && \
     yum clean all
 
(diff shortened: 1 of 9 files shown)


bexelbie:
This is needed for boot2docker/docker-machine

cdrage:
Sorry for the delay, but this needs a re-base :(

is there a reason why we use the /run dir instead of /var?

Same issue appears for alpine linux ++

bexelbie:
rebased

cdrage:
So tests pass for me in regards to this.

I've ran two atomicapp's on the same system at the same time and it seems to lock correctly. 

@dustymabe @rtnpro for secondary review?

PR #503, opened 2016-01-13
native openshift: support tls verification

dustymabe:


dustymabe:
Darn syntax checks.. I'll fix those.. just review for code quality in the mean time.

dustymabe:
@rtnpro after the ssl verify PR from yesterday we'll need this PR in order to get tls verification working again for native openshift `oc new-app` deployments.

rtnpro:
:+1:

atomicapp/providers/openshift.py
```diff
@@ -40,6 +40,9 @@
 import logging
 logger = logging.getLogger(__name__)
 
+# If running in an openshift POD via `oc new-app`, the ca file is here
+OPENSHIFT_POD_CA_FILE = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
```
cdrage: Should we keep this here on in `constants.py`?

cdrage:
The code LGTM's. I believe I outlined my thoughts on that previous OpenShift code being located within `main.py` but :+1: that it has been moved to `openshift.py`

The tests pass :sparkles: and other than my one comment, LGTM!

atomicapp/providers/openshift.py
dustymabe: The question is: do we put provider specific contstants into constants.py? I'm on the fence.

PR #512, opened 2016-01-15
openshift provider: safer stop (undeploy)

kadel:
Instead of finding and deleting related pods for RC, scale RC down to 0 replicas.

This should be much more cleaner then before.

This is how `kubectl` deletes RCs: https://github.com/kubernetes/kubernetes/blob/439d7a721e087f4015191fe7e2d0b9262abd2530/pkg/kubectl/stop.go#L127

This is how `oc` deletes DCs: https://github.com/openshift/origin/blob/bccc1d35c067d367009ee1406d5f2fc603b1ab5a/pkg/deploy/reaper/reaper.go#L33

TODO: 
- [x] make sure that DC is deleted before its RCs so it is not recreated [source](https://github.com/openshift/origin/blob/bccc1d35c067d367009ee1406d5f2fc603b1ab5a/pkg/deploy/reaper/reaper.go#L39)

### DIFF
--- a/atomicapp/providers/openshift.py
+++ b/atomicapp/providers/openshift.py
@@ -167,6 +167,30 @@ class OpenshiftClient(object):
             logger.error(msg)
             raise ProviderFailedException(msg)
 
+    def scale(self, url, replicas):
+        """
+        Scale ReplicationControllers or DeploymentConfig
+
+        Args:
+          url (str): full url for artifact
+          replicas (int): number of replicas scale to
+        """
+        patch = [{"op": "replace",
+                  "path": "/spec/replicas",
+                  "value": replicas}]
+
+        (status_code, return_data) = \
+            Utils.make_rest_request("patch",
+                                    url,
+                                    data=patch,
+                                    verify=self._requests_tls_verify())
+        if status_code == 200:
+            logger.info("Sucessfully scaled %s to %s replicas", url, replicas)
+        else:
+            msg = "%s %s" % (status_code, return_data)
+            logger.error(msg)
+            raise ProviderFailedException(msg)
+
     def process_template(self, url, template):
         (status_code, return_data) = \
             Utils.make_rest_request("post",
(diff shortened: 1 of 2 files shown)


cdrage:
Yayyyy! :sparkles: during adding the new verbs I **thought** I saw scale replicas missing, and I was right.

Thanks for this, once we have this merged I can get this into the new verbs. It's the only thing missing for the OpenShift provider.

kadel:
Deletion is now  done same way as in `kubectl` and `oc`.

atomicapp/providers/openshift.py
```diff
@@ -371,13 +395,17 @@ def undeploy(self):
         logger.debug("Starting undeploy")
         delete_artifacts = []
         for kind, objects in self.openshift_artifacts.iteritems():
-            delete_artifacts.extend(objects)
+            # Add DCs to beggining of list so they are deleted first.
```
dustymabe: s/beggining/beginning

Also can we add more explanation of why we do DCs first: "Do DC first because if you do RC before DC then the DC will re-spawn the RC before the DC is deleted."

cdrage:
Tests pass for run both run and stop on OpenShift Etherpad and Wordpress examples. :+1: 

Just going to review code now.

cdrage:
From a few manual tests, this LGTM :+1:! Same as the code. Nice and clean.

Pinging @rtnpro or @dustymabe for secondary review.

atomicapp/providers/openshift.py
```diff
@@ -371,13 +395,19 @@ def undeploy(self):
         logger.debug("Starting undeploy")
         delete_artifacts = []
         for kind, objects in self.openshift_artifacts.iteritems():
-            delete_artifacts.extend(objects)
+            # Add DCs to beginning of list so they are deleted first.
```
cdrage: "beginning of the list" 
I think we should have "deployment config" instead of the DC acronym.

atomicapp/providers/openshift.py
```diff
@@ -371,13 +395,19 @@ def undeploy(self):
...
+            # Do DC first because if you do RC before DC then the DC
```
cdrage: use deployment config

kadel:
updated comments based on suggestions

cdrage:
@kadel awesome! LGTM :+1:

dustymabe:
works for me. LGTM

PR #519, opened 2016-01-18
cli: allow specifying target dir during atomic run

dustymabe:
This commit has one simple goal which is to allow `atomic run` to
be executed against an installed directory just a little easier.

Before this commit we need:

    atomic install myapp --destination=/path/to/local/dir
    atomic run myapp --destination=/path/to/local/dir

After this commit:

    atomic install myapp --destination=/path/to/local/dir
    atomic run myapp /path/to/local/dir


rtnpro:
More intuitive. LGTM :+1:

kadel:
why don't we allow same thing for install? eg. `atomic install myapp /path/to/local/dir`

dustymabe:
'/path/to/local/dir' is a path to an "existing" application that has been extracted. In the case of `install` there is not yet an "existing" application, otherwise you wouldn't be calling `install`.

Does that make sense?

cdrage:
LGTM :+1: and all tests pass!

PR #521, opened 2016-01-19
Allow users to provide answers file as url.

dustymabe:
This will allow one to specify a location of an answers file that can
be retrieved by Atomic App during deployment. A common case for this
would be if someone deploying the application wants to deploy many
times with a known good configuration where this configuration is
stored in source control or some other location accessed via a url.

This also may be one avenue for providing answers to applications
started via `oc new-app` in openshift.

### DIFF
--- a/atomicapp/nulecule/main.py
+++ b/atomicapp/nulecule/main.py
@@ -5,6 +5,8 @@ import distutils.dir_util
 import logging
 import os
 import tempfile
+import urlparse
+import urllib
 
 from atomicapp.constants import (GLOBAL_CONF,
                                  ANSWERS_FILE_SAMPLE_FORMAT,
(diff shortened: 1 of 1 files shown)


rtnpro:
LGTM :+1:

kadel:
I just wonder if it is right that if there is already `answers.conf` in app dir we overwrite it with new downloaded one.
I probably answer myself that `--answers` parameter is for overriding it :-)
I'm just not a fan of apps that overwrite files without even warning :relaxed: 

But otherwise LGTM :+1:

dustymabe:
> I just wonder if it is right that if there is already answers.conf in app dir we overwrite it with new downloaded one.

Yes. I did this because we need to support 438ad78. We may come up with a better solution in the future.

dustymabe:
Added another commit to handle creation of the app_path dir in the case where someone has specified the answers file as a url.

cdrage:
I believe this fills one of the requirements within the Nulecule spec in regards to supplying URL's. We can apply this to artifacts as well in the future. 

Otherwise, LGTM :+1:

dustymabe:
rebased on master.. merging now.

PR #527, opened 2016-01-21
Remove symbolic linking of Dockerfile

cdrage:
~~After numerous replies to a ticket with Docker hub, it seems that the issue with symbolically linked Dockerfiles no longer works.~~

~~They have added a feature of specifying a specific Dockerfile as of Monday January 18th, however, you still cannot specify the working directory of the builder. Thus when choosing for example: /Dockerfiles.git/Dockerfile.centos with auto-builder, it does not detect requirements.txt, atomicapp/ folder, etc. as you cannot pass the working dir.~~

~~This PR adds a non-symbolically-linked Dockerfile.dev file to the root dir with the main purpose of building the :dev branch on each commit.~~

Let's go back to how we had it before without the symbolic link to at least fix the Docker hub issue until they get linking fixed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,1 +1,43 @@
-Dockerfiles.git/Dockerfile.centos
+FROM centos:7
+
+MAINTAINER Red Hat, Inc. <container-tools@redhat.com>
+
+ENV ATOMICAPPVERSION="0.4.0"
+
+LABEL io.projectatomic.nulecule.atomicappversion=${ATOMICAPPVERSION} \
+      io.openshift.generate.job=true \
+      io.openshift.generate.token.as=env:TOKEN_ENV_VAR \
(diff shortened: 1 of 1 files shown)


dustymabe:
Rather than do this would it be better just to copy the Dockerfiles.git/Dockerfile.centos over top of the symlink for now and we'll just keep them synced until docker hub sorts these issues out?

cdrage:
@dustymabe We can do that too at least until this issue is fixed.

dustymabe:
yeah lets do that instead.

cdrage:
updated ^^ @dustymabe

dustymabe:
Let's not delete the Dockerfiles.git/Dockerfile.centos file. We want it to stay in place.

dustymabe:
oops. my mistake. LGTM

PR #535, opened 2016-01-26
Update readme

cdrage:
* Removes mentions of install for fetch
* New year, new copyright update
* Cleans up some of descriptions
* Remove grammatical errors

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,15 @@
 # Atomic App
 
-Atomic App is a reference implementation of the [Nulecule Specification](http://www.projectatomic.io/docs/nulecule/). It can be used to bootstrap container applications and to install and run them. Atomic App is designed to be run in a container context. Examples using this tool may be found in the [Nulecule examples directory](https://github.com/projectatomic/nulecule/tree/master/examples).
+Atomic App is a reference implementation of the [Nulecule specification](https://github.com/projectatomic/nulecule). It can be used to bootstrap packaged container environments and then run them. Atomic App is designed to be ran within a container. 
+
+Examples of this tool may be found within the [Nulecule library repo](https://github.com/projectatomic/nulecule/tree/master/examples).
 
 ## Getting Started
 
-Atomic App is packaged as a container. End-users typically do not install the software from source. Instead use the atomicapp container as the `FROM` line in a Dockerfile and package your application on top. For example:
+Atomic App itself is packaged as a container. End-users typically do not install the software from source. Instead use the `atomicapp` container as the `FROM` line in a Dockerfile and package your application on top. For example:
 
 ```
-FROM projectatomic/atomicapp:0.4.2
+FROM projectatomic/atomicapp
 
 MAINTAINER Your Name <you@example.com>
 
@@ -15,17 +17,17 @@ ADD /Nulecule /Dockerfile README.md /application-entity/
 ADD /artifacts /application-entity/artifacts
 ```
 
-For more information see the [Atomic App getting started guide](http://www.projectatomic.io/docs/atomicapp/).
+For more information see the [Nulecule getting started guide](https://github.com/projectatomic/nulecule/blob/master/docs/getting-started.md).
 
 ## Developers
 
 First of all, clone the github repository: `git clone https://github.com/projectatomic/atomicapp`.
 
-### Install this project
+### Installing Atomic App locally
 Simply run
 
 ```
-pip install .
+make install
 ```
 
 If you want to do some changes to the code, I suggest to do:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,47 +27,40 @@ First of all, clone the github repository: `git clone https://github.com/project
...
 ```
-pip install .
-```
-
-If you want to do some changes to the code, I suggest to do:
-
-```
-cd atomicapp
-export PYTHONPATH=`pwd`:$PYTHONPATH
-alias atomicapp="python `pwd`/atomicapp/cli/main.py"
```
cdrage: i don't think this should be included ^^. let python devs do their own way of development

cdrage:
pingarino @goern @vpavlin since you two did most of the initial README work according to git logs :)

cdrage:
Ping @dustymabe @rtnpro for review as this part is old stuff ^^ we don't use `install` anymore.

README.md
```diff
@@ -25,47 +27,40 @@ First of all, clone the github repository: `git clone https://github.com/project
...
-### Build
+### Build an Atomic App
 ```
 docker build -t [TAG] .
 ```
 
-Just a call to Docker to package up the application and tag the resulting image.
+Use 'docker build' to package up the application and tags the resulting image.
 
 ### Install and Run
```
dustymabe: Install --> Fetch ??

README.md
```diff
@@ -25,47 +27,40 @@ First of all, clone the github repository: `git clone https://github.com/project
...
 ```
-atomicapp [--dry-run] [-a answers.conf] install|run [--recursive] [--update] [--destination DST_PATH] APP|PATH
+atomicapp [--dry-run] [-a answers.conf] fetch|run|stop [--recursive] [--update] [--provider docker] [--destination DST_PATH] APP|PATH
```
dustymabe: There is also genanswers. I think maybe we should not put all of the options here because it gets too long.

README.md
```diff
@@ -25,47 +27,40 @@ First of all, clone the github repository: `git clone https://github.com/project
...
 ```
 
-Pulls the application and it's dependencies. If the last argument is
+Pulls the application and its dependencies. If the last argument is
 existing path, it looks for `Nulecule` file there instead of pulling anything.
 
 * `--recursive yes|no` Pull whole dependency tree
 * `--update` Overwrite any existing files
```
dustymabe: are these two options really useful? I wonder if we shouldn't get rid of them

README.md
dustymabe: I don't mind having it. Sometimes I use this and I always come back here as a reference.

dustymabe:
Other than comments it looks good to me.

README.md
cdrage: yeah, you're right, I don't think I've actually used either of them once.

cdrage:
@dustymabe updated with your comments.

README.md
```diff
@@ -36,36 +38,35 @@ export PYTHONPATH=`pwd`:$PYTHONPATH
 alias atomicapp="python `pwd`/atomicapp/cli/main.py"
 ```
 
-### Build
+### Build an Atomic App
```
dustymabe: Can we be more clear about what this is doing? "Build an Atomic App" sounds like I am doing an application build (i.e. helloapache) vs packaging up the atomicapp software. Maybe this would be better: `Build an Atomic App Container Image - for Container Execution`. And then mention running it via Atomic CLI? 

And maybe we even want to change the name of the previous section to `Install Atomic App - for Local Execution`

README.md
```diff
@@ -36,36 +38,35 @@ export PYTHONPATH=`pwd`:$PYTHONPATH
...
 ```
 docker build -t [TAG] .
 ```
 
-Just a call to Docker to package up the application and tag the resulting image.
+Use 'docker build' to package up the application and tags the resulting image.
```
dustymabe: s/tags/tag/

dustymabe:
@cdrage - can we get this updated?

cdrage:
Hey @dustymabe 
So I updated it with the new title as well as the spelling error.

In regards to `atomic cli` execution. Should we add a separate section for alternatively using the atomic CLI? 

Otherwise, rebased again ^^

dustymabe:
`Build an Atomic App for containerized execution` still sounds wrong. 

What we are doing is containerizing the Atomic App software so it can then be run in a container but it isn't actually an "Atomic App". `Building for containerized execution` would be a better title. We should probably update `Install this project` to `Install Atomic App Software Locally` or something along those lines. 

I say just digest these comments and update it based on your best judgement and then MERGE!!

cdrage:
@dustymabe 
Updated based on your comments!

So I'm going to push this PR for now and I'll be adding a separate section for "atomic cli" execution this week. Work on doc updating for our upcoming release :)

surajssd:
I think work on #543 can continue :)

Issue #486, opened 2016-01-07
openshift: make sure namespace value gets back into config


dustymabe:
If I run and my answers.conf doesn't have a `namespace=default` in it then the namespace from the `providerconfig` (if provided) will be used. We need to make sure the namespace that gets used gets put back in the config so that the written out answers.conf.gen file will have it. If we don't then when we try to run a `stop` we will see an error because the wrong value for namespace got put in `answers.conf.gen`:

```
[root@f23 etherpad-centos7-atomicapp (master *%>)]# aa stop ./
2016-01-07 15:07:56,061 - __main__ - INFO - Action/Mode Selected is: stop
2016-01-07 15:07:56,082 - atomicapp.nulecule.base - INFO - Found existing external application for mariadb-centos7-atomicapp. Loading it.
2016-01-07 15:07:56,133 - openshift - ERROR - There are conflicting values in /host/root/.kube/config (proj1) and answers.conf (default)
2016-01-07 15:07:56,133 - __main__ - ERROR - There are conflicting values in /host/root/.kube/config (proj1) and answers.conf (default)
Traceback (most recent call last):
  File "/opt/atomicapp/atomicapp/cli/main.py", line 118, in cli_stop
    nm.stop(**argdict)
  File "/opt/atomicapp/atomicapp/nulecule/main.py", line 272, in stop
    self.nulecule.stop(cli_provider, dryrun)
  File "/opt/atomicapp/atomicapp/nulecule/base.py", line 153, in stop
    component.stop(provider_key, dryrun)
  File "/opt/atomicapp/atomicapp/nulecule/base.py", line 272, in stop
    self._app.stop(provider_key, dryrun)
  File "/opt/atomicapp/atomicapp/nulecule/base.py", line 153, in stop
    component.stop(provider_key, dryrun)
  File "/opt/atomicapp/atomicapp/nulecule/base.py", line 276, in stop
    provider.init()
  File "/opt/atomicapp/atomicapp/providers/openshift.py", line 157, in init
    self._get_config_values()
  File "/opt/atomicapp/atomicapp/providers/openshift.py", line 566, in _get_config_values
    raise ProviderFailedException(msg)
ProviderFailedException: There are conflicting values in /host/root/.kube/config (proj1) and answers.conf (default)
```

dustymabe:
We merged a workaround in #538 but we are leaving this bug open because it is not a pure solution.

PR #538, opened 2016-01-28
If not given, don't populate namespace in answers.conf.gen.

dustymabe:
If the namespace is not provided in the answers.conf, but it is
detected in the providerconfig, then don't put namespace into the
answers.conf.gen file because that will make stopping the application
not work.

A better solution to this would be to populate the right value for
namespace (the detected value) in the answers.conf.gen, but the way
our config data is handled, passing data back from the provider code
would be really messy.

This is a workaround for #486. It is not a fix.


dustymabe:
I'll try to fix this syntax issue later today and update the PR so we can possibly get it in.

dustymabe:
@cdrage - syntax error should be fixed now.

cdrage:
LGTM. Merging.

cdrage:
Tests pass too :+1:

PR #540, opened 2016-02-01
marathon: do not convert types when parsing json artifact

kadel:
If marathon artifact has env variable that has number as value anymarkup converts this to int.
But Marathon api requires all env to be string:string.

`force_types=None` turns off type conversions. [1]

[1] https://github.com/bkabrda/anymarkup#notes-on-parsing-basic-types


dustymabe:
might be worth adding a comment to the code that indicates why we are disabling the type conversion. Other than that, looks good.

kadel:
Good point, I'll do that.

cdrage:
LGTM :+1:

Issue #390, opened 2015-11-10
What's the need for the docker daemon / dependency?


charliedrage:
Been messing around with implementing an alpinelinux Dockerfile for reducing the atomicapp image size.

What's the need for the docker daemon running in our container / the entire package. Ideally, we only need the Docker binary for communication with the docker socket on the host.

My proposal is changing installing via yum/apt-get/etc. in favor of simply grabbing the binary.

```Dockerfile
RUN yum -y install epel-release && \                                                                                                                                                                                                                           
    yum install -y --setopt=tsflags=nodocs python-pip python-setuptools gcc curl && \                                                                                                                                                                          
    python setup.py install && \                                                                                                                                                                                                                               
    curl -sSl https://get.docker.com/builds/Linux/x86_64/docker-1.8.3 > /usr/bin/docker && \                                                                                                                                                                   
    chmod a+x /usr/bin/docker && \                                                                                                                                                                                                                             
    yum remove -y gcc cpp glibc-devel glibc-headers kernel-headers libmpc mpfr python-pip && \                                                                                                                                                                 
    yum clean all 
```

Instead of:

```Dockerfile
RUN yum -y install epel-release && \
    yum install -y --setopt=tsflags=nodocs python-pip python-setuptools docker gcc && \
    python setup.py install && \
    yum remove -y gcc cpp glibc-devel glibc-headers kernel-headers libmpc mpfr python-pip && \
    yum clean all
```

Initial tests show that this reduces the `projectatomic/atomicapp` image from __254mb__ to __212mb__ on the `centos:centos7` based Dockerfile.

And an alpine linux Dockerfile managed to be only __73.42mb__ in size!
```bash
projectatomic/alpine                 latest              058cf216dda3        5 seconds ago       73.42 MB
```

dustymabe:
So you are right in that we don't need all of the junk that the docker rpm brings in. Basically it brings in everything you need to run docker, but we only need to make client calls. If we want to continue having the docker client in the atomicapp base container then we would need to not install the deps.

- For now we can install the rpm with no deps (pending testing)
- In the future we can use docker-client rpm (see https://bugzilla.redhat.com/show_bug.cgi?id=1280471)

An alternative is that we actually just assume the host has the docker daemon installed and do a chroot to the host when we call docker.

dustymabe:
Either way, though, pulling the binary from something other than an rpm isn't going to be acceptable.

charliedrage:
[...]

PR #546, opened 2016-02-09
[WIP] Add Alpine Linux Dockerfile

### LINKED ISSUE #544: Consider rebasing Atomic App image on a stripped-down OS
The atomicapp example container, based on CentOS7, is 289MB in size.  That's very large for an empty container.   If we want people to actually build their apps on top of it, then it needs to be smaller ... in the neighborhood of 30 to 50MB.  

Would it be possible for us to obtain or produce a stripped-down OS for a smaller image?

cdrage:
Yay < 80Mb Docker image! :sparkles:

Uses the Docker binary at the momen.

In the future, the Docker provider will be replaced by API commands to the Docker daemon rather than cmd line running, making this image even smaller, reducing it to 60Mb. 

Fixes #544

Referencing #390 due to previous discussion as we initially said no. Since we have a lot more images now, may be nice to include this as "experimental"?

From initial tests seems that this builds correctly / works well.


cdrage:
Perhaps we can make this the default in the future due to our increasing size (300MB+ :-1: )

cdrage:
Hey @jberkus was just wondering if you happened to test this as well? It works my end from the limited testing I've done. We can add this is as an experimental feature until it's 100% robust.

jberkus:
@cdrage no, I haven't had time to build anything since DevConf.  I'll be doing some this week, though.

However ... strategy for Atomic Host seems to be that we want systemd inside the containers, which means looking at a stripped-down version of Fedora instead of Alpine, in the long run.  There's some real advantages to this provided that the issues can be worked out.

So I'm not sure that making Alpine default for Atomic.app is the way to go, given that.

dustymabe:
@cdrage - I am getting an error when trying to build and run with this (see below) - perhaps rebase and make sure it still works for you.
```
2016-03-01 22:17:02,666 - kubernetes - INFO - Deploying to Kubernetes
2016-03-01 22:17:02,666 - kubernetes - DEBUG - ./artifacts/kubernetes/.mariadb-pod.yaml
2016-03-01 22:17:02,673 - kubernetes - DEBUG - ./artifacts/kubernetes/.mariadb-service.yaml
2016-03-01 22:17:02,681 - __main__ - ERROR - [Errno 2] No such file or directory
```

cdrage:
@dustymabe 
hmmm so it looks like it works normally via actual `atomicapp`. but not via `atomic`

cdrage:
woops wrong button, reopened

dustymabe:
@cdrage I run atomicapp via the alias.. so similar to what atomic cli does, but not the same. I guess you need to test using that case. No rush, just probably need to make sure that works

cdrage:
Ahhhhh, I'm encountering an issue with the Docker binary within Alpine. I'm going to test this a bit more.

cdrage:
I'm going to close this for now as I'm working on getting a Python binary file working (it's 8mb total) and use FROM SCRATCH instead.

Issue #502, opened 2016-01-13
Atomicapps fails when there is not metadata in Nulecule


kadel:
AtomicApp fails when there is no `metadata` field in Nulecule.
```
 __main__ - ERROR - __init__() takes at least 6 arguments (7 given)
Traceback (most recent call last):
  File "/opt/atomicapp/atomicapp/cli/main.py", line 98, in cli_run
    nm.run(**argdict)
  File "/opt/atomicapp/atomicapp/nulecule/main.py", line 235, in run
    self.nulecule = self.unpack(dryrun=dryrun, config=self.answers)
  File "/opt/atomicapp/atomicapp/nulecule/main.py", line 139, in unpack
    self.app_path, dryrun=dryrun, config=config)
  File "/opt/atomicapp/atomicapp/nulecule/base.py", line 123, in load_from_path
    namespace=namespace, **nulecule_data)
TypeError: __init__() takes at least 6 arguments (7 given)
```

According to [Nulecule spec](https://github.com/projectatomic/nulecule/tree/master/spec#container-application-object) `metadata` field should be optional.

surajssd:
I would like to work on this issue.

dustymabe:
Suraj, thanks. I assigned the issue to you

rtnpro:
Go @surajssd, go :+1:

surajssd:
The [Nulecule spec](https://github.com/projectatomic/nulecule/tree/master/spec#container-application-object) defines that the metadata should be optional but looking at the code [here](https://github.com/projectatomic/atomicapp/blob/master/atomicapp/nulecule/base.py#L42) it feels like the implementation is not done in such a way. So do we need to change the way Nulecule object is made by making `metadata` parameter a keyword parameter or should we make a workaround by keeping the definition as it is?(I do have a working, workaround though)

rtnpro:
@surajssd The Nulecule SPEC is correct. We need to fix the implementation of ``metadata`` in atomicapp. So, [here](https://github.com/projectatomic/atomicapp/blob/master/atomicapp/nulecule/base.py#L42), we can move ``metadata`` as a keyword argument with ``None`` as default. And in the ``__init__`` method, we do something like:
```
self.metadata = metadata or {}
```
so that we have a dictionary for any subsequent use of ``metadata``.

I hope that this helps.

surajssd:
[...]

PR #547, opened 2016-02-10
made `metadata` an optional argument in Nulecule file

surajssd:
Atomicapp would fail if `metadata` was not specified in Nulecule file.
While according to Nulecule specification `metadata` is an optional
argument. This would fail because while creating `Nulecule` object
`metadata` was positional argument. So fixed this by making it a keyword
argument. Also tests have been modified accordingly.

This closes issue #502


cdrage:
LGTM :+1:  !

PR #516, opened 2016-01-18
Add man page for atomicapp

ingvagabund:
Content collected from atomicapp [command] -h and from docs.

PR #549, opened 2016-02-10
Add option in makefile to copy over manpage during install

cdrage:
Relies on #516. Please merge that first :)

Adds a manpage to atomicapp, THANKS @ingvagabund !!!


cdrage:
Closing this. We're going to find a way to automatically generate a man page for atomicapp rather than doing it manually each time.

PR #550, opened 2016-02-10
Makes Makefile faster using .PHONY && add default python location

cdrage:
See https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html as well as the improvements on Atomic CLI's Makefile https://github.com/projectatomic/atomic/blob/master/Makefile

We use .PHONY to avoid any conflicting names as well as speed up installation.

A default dir for Python has also been added for any future changes (ex. if we want to build with python3 we can do PYTHON=/usr/bin/python3 sudo make install)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 # Travis CI checks for atomicapp
 
 language: python
+sudo: required
 
 python:
   - "2.7"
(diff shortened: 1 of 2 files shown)


cdrage:
Wut. Looks like travis borked. Going to try it again. May have to do with :sudo.

cdrage:
#dotests

cdrage:
Rebased.

Ping @kadel @dustymabe 

This is right up your alley too @goern :)

cdrage:
Ping @kadel @dustymabe

This is right up your alley too @goern :)

kadel:
LGTM

dustymabe:
unittests failing.. LGTM if you can get tests passing.

cdrage:
Seems to be a false positive with centos-ci.

Redoing again

cdrage:
#dotests

cdrage:
#dotests

cdrage:
#dotests

cdrage:
All tests pass :100: Mergin'

PR #551, opened 2016-02-10
cli: Print helpful error if no app_spec provided.

dustymabe:
If a user doesn't provide an app_spec and it can't be derived from
the $IMAGE environment variable, then print an error and exit.


cdrage:
Tests pass, all LGTM :+1:

dustymabe:
rebased on master.. will merge soon - running some tests of my own

dustymabe:
Tests pass. merging.

PR #552, opened 2016-02-11
providerconfig: support someone specifying a relative path

dustymabe:
This will allow someone to support a relative path for the
providerconfig. The providerconfig must be under the users
current working directory and can't be outside of it.


cdrage:
LGTM :+1:

dustymabe:
Tests in ci.centos.org pass. woot!

PR #555, opened 2016-02-11
Fix reliability of defaulting to provider

### LINKED ISSUE #515: Defaulting to provider is unreliable
`Utils.getSupportedProviders()` doesn't properly handle that artifacts could be inherited artifacts in the Nulecule file (an example is [here](https://github.com/projectatomic/nulecule-library/blob/feb892e0d7e7aa7c218334e9fa3028cb1db137c9/mariadb-centos7-atomicapp/Nulecule#L27)). I hit this issue when trying to add providers openshift support to the guestbook example. After adding the 'inherited' openshift I then wanted to use openshift but the code defaulted to use kubernetes because there is no `/artifacts/openshift` directory (added in b76acee).

We should probably get rid of this functionality once we can properly support setting a default provider within the global params like so:

```
params:
    - name: provider
      description: The provider that is used to deploy the application
      default: openshift
```

This doesn't currently work right now because our config handling is convoluted :(

dustymabe:
Fixes #515

### DIFF
--- a/atomicapp/nulecule/main.py
+++ b/atomicapp/nulecule/main.py
@@ -223,12 +223,6 @@ class NuleculeManager(object):
         if not self.answers_file:
             self._process_answers()
 
-        # Unless otherwise specified with CLI arguments we will
-        # default to the first provider available
-        providers = Utils.getSupportedProviders(self.app_path)
-        if cli_provider is None and len(providers) == 1:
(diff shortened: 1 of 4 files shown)


cdrage:
lgtm :+1: I see why we're switching back to specifying it through params. makes more sense.

dustymabe:
tests passed in ci.centos.org too (same tests as Charlie's, but it is nice to confirm they pass in both places). Merging.

PR #532, opened 2016-01-26
Fail nicely if user does not provide artifact files within Nulecule.

cdrage:
Instead of erroring out:
```
2015-12-03 18:26:31,590 - atomicapp.cli.main - ERROR - argument of type 'NoneType' is not iterable
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.2.3-py2.7.egg/atomicapp/cli/main.py", line 74, in cli_run
    nm.run(**argdict)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.2.3-py2.7.egg/atomicapp/nulecule/main.py", line 187, in run
    self.nulecule.render(cli_provider, dryrun)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.2.3-py2.7.egg/atomicapp/nulecule/base.py", line 223, in render
    component.render(provider_key=provider_key, dryrun=dryrun)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.2.3-py2.7.egg/atomicapp/nulecule/base.py", line 348, in render
    if provider_key and provider_key not in self.artifacts:
TypeError: argument of type 'NoneType' is not iterable
```

It will error out with:
```
2016-01-26 11:30:27,838 - atomicapp.cli.main - ERROR - Unable to find artifact. Is the source path correct? 
```

Suggestions on wording is welcome.

PR #556, opened 2016-02-12
Fail if unable to find artifact

cdrage:
2nd half of #532, raises NuleculeException if unable to find a file.

Another PR ahead for failing when using a wrong inherited provider, (ex. providing inherit: foobar).

### DIFF
--- a/atomicapp/nulecule/base.py
+++ b/atomicapp/nulecule/base.py
@@ -548,6 +548,8 @@ class NuleculeComponent(NuleculeBase):
         immediate children, i.e., we do not deal with nested artifact
         directories at this moment.
 
+        If a file or directory is not found, raise an exception.
+
         Args:
             path (str): Local path
 
@@ -558,9 +560,14 @@ class NuleculeComponent(NuleculeBase):
         if os.path.isfile(path):
             artifact_paths.append(path)
         elif os.path.isdir(path):
+            if os.listdir(path) == []:
+                raise NuleculeException("Artifact directory %s is empty" % path)
             for dir_child in os.listdir(path):
                 dir_child_path = os.path.join(path, dir_child)
                 if dir_child.startswith('.') or os.path.isdir(dir_child_path):
                     continue
                 artifact_paths.append(dir_child_path)
+        else:
+            raise NuleculeException("Unable to find artifact %s" % path)
+
         return artifact_paths


atomicapp/nulecule/base.py
```diff
@@ -548,4 +550,7 @@ def _get_artifact_paths_for_path(self, path):
                 if dir_child.startswith('.') or os.path.isdir(dir_child_path):
                     continue
                 artifact_paths.append(dir_child_path)
+        else:
+            raise NuleculeException("Unable to find artifact %s" % path)
+
```
dustymabe: so right now we only raise an exception if the file or dir doesn't exist. what happens if they pass us a directory but the directory is empty? Should that be an error?

cdrage: @dustymabe Updated to include this too :)

cdrage:
PR updated to include if directory is empty.

atomicapp/nulecule/base.py
```diff
@@ -548,4 +550,9 @@ def _get_artifact_paths_for_path(self, path):
                 if dir_child.startswith('.') or os.path.isdir(dir_child_path):
                     continue
                 artifact_paths.append(dir_child_path)
+        elif os.listdir(path) == []:
+            raise NuleculeException("Artifact directory %s is empty" % path)
```
dustymabe: this is dead code. If it is a directory the it will pass the 2nd if and won't ever get to the 3rd one right?

atomicapp/nulecule/base.py
```diff
@@ -548,4 +550,9 @@ def _get_artifact_paths_for_path(self, path):
...
+        else:
+            raise NuleculeException("Unable to find artifact %s" % path)
+
```
dustymabe: Instead of putting our logic in the if/else statements maybe we should just check to see if artifact_paths is empty before we return and raise an exception there?

cdrage:
Hey @dustymabe I updated it, don't know if it's to your satisfaction, hopefully it's better :+1:

dustymabe:
LGTM. rebase and merge please

dustymabe:
@cdrage ^^

cdrage:
Rebased and unit/integration tests pass. Mergin'

PR #557, opened 2016-02-12
Fix xpathing tests on missing files

cdrage:
Must be merged before #556

Fixes tests on xpathing files due to missing artifacts. Making #556 error out on tests due to missing files.

### DIFF
--- a/tests/units/nulecule/test_xpathing.py
+++ b/tests/units/nulecule/test_xpathing.py
@@ -35,10 +35,10 @@ class TestNuleculeXpathing(unittest.TestCase):
 
     # Create a temporary directory for our setup as well as load the required NuleculeComponent
     def setUp(self):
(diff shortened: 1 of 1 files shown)


dustymabe:
LGTM. rebase and merge please

dustymabe:
Actually I don't know if this is right. First off, it doesn't actually make the tests pass when testing on top of the commit in #556. In the test file for this we create a temporary directory but never do anything with it.

I don't see any other tests that use a tmp dir, they use the examples dir directly. Should we not be doing the same thing? If I update the test to point to the example dir (`os.path.dirname(__file__) + '/artifact_xpath_test/'`) then the tests pass.

cdrage:
Yeah, you're right. Let's switch from tmp files to use an explicit example (pref helloapache). I'll fix this. Do not merge for now.

cdrage:
@dustymabe Updated :+1:

dustymabe:
LGTM merging

Issue #562, opened 2016-02-16
Handle wrong provider name in answers.conf


surajssd:
If the `provider` name for the atomicapp in `answers.conf` is mistaken(other than `docker`, `kubernetes`, `openshift`, `marathon`) then it fails giving all the stack trace without any sane error. Re-creating the errorsome condition:

```bash
(atomicapp) [vagrant@centos7-adb apache-centos7-atomicapp]$ pwd
/home/vagrant/nulecule-library/apache-centos7-atomicapp
(atomicapp) [vagrant@centos7-adb apache-centos7-atomicapp]$ tree
.
├── answers.conf
├── answers.conf.orig
├── artifacts
│   ├── docker
│   │   └── apache-centos7-atomicapp_run
│   ├── kubernetes
│   │   └── apache-centos7-atomicapp-pod.json
│   └── marathon
│       └── apache-centos7-atomicapp.json
├── Dockerfile
├── Nulecule
└── README.md
 
4 directories, 8 files
(atomicapp) [vagrant@centos7-adb apache-centos7-atomicapp]$ cat answers.conf
[apache-centos7-atomicapp]
image = centos/httpd
hostport = 80
[general]
namespace = default
provider = something
(atomicapp) [vagrant@centos7-adb apache-centos7-atomicapp]$ sudo atomicapp run . -v
2016-02-16 07:00:14,739 - atomicapp.cli.main - INFO - Action/Mode Selected is: run
2016-02-16 07:00:14,739 - atomicapp.cli.main - DEBUG - Final parsed cmdline: run . -v
2016-02-16 07:00:14,739 - atomicapp.nulecule.main - DEBUG - NuleculeManager init app_path: .
2016-02-16 07:00:14,739 - atomicapp.nulecule.main - DEBUG - NuleculeManager init    image: None
2016-02-16 07:00:14,739 - atomicapp.utils - DEBUG - Loading answers from file: ./answers.conf
2016-02-16 07:00:14,740 - atomicapp.nulecule.main - DEBUG - Request to unpack to None to .
2016-02-16 07:00:14,744 - atomicapp.plugin - DEBUG - Loading providers from /home/vagrant/atomicapp/atomicapp/providers
2016-02-16 07:00:14,744 - atomicapp.plugin - DEBUG - Loading providers from /home/vagrant/atomicapp/atomicapp/providers
2016-02-16 07:00:14,746 - atomicapp.cli.main - ERROR - 'NoneType' object is not callable
[...]

PR #563, opened 2016-02-17
Wrong provider name in answers.conf, exits AtomicApp with readable error

surajssd:
When `provider` name in `answers.conf` is wrong then AtomicApp fails with stack-trace and no error output which is user understandable. This happened because there was no check for the condition if
given `provider_key` does not match with any of the providers that are supported for now.

Fixes Issue #562

### DIFF
--- a/atomicapp/nulecule/lib.py
+++ b/atomicapp/nulecule/lib.py
@@ -5,9 +5,11 @@ from atomicapp.constants import (GLOBAL_CONF,
                                  LOGGER_COCKPIT,
                                  NAME_KEY,
                                  DEFAULTNAME_KEY,
-                                 PROVIDER_KEY)
+                                 PROVIDER_KEY,
+                                 PROVIDERS)
 from atomicapp.utils import Utils
 from atomicapp.plugin import Plugin
+from atomicapp.nulecule.exceptions import NuleculeException
 
 cockpit_logger = logging.getLogger(LOGGER_COCKPIT)
 
@@ -101,6 +103,11 @@ class NuleculeBase(object):
         if provider_key is None:
             provider_key = self.config.get(GLOBAL_CONF)[PROVIDER_KEY]
         provider_class = self.plugin.getProvider(provider_key)
+        if provider_class is None:
+            raise NuleculeException("Invalid Provider - '{}', provided in "
+                                    "answers.conf (choose from {})"
+                                    .format(provider_key, ', '
+                                                          .join(PROVIDERS)))
         return provider_key, provider_class(
             self.get_context(), self.basepath, dry)
 
--- a/tests/units/nulecule/test_lib.py
+++ b/tests/units/nulecule/test_lib.py
@@ -1,0 +1,40 @@
+import mock
+import unittest
+
+from atomicapp.nulecule.lib import NuleculeBase
+from atomicapp.nulecule.exceptions import NuleculeException
+
+
+class TestNuleculeBaseGetProvider(unittest.TestCase):
+    """ Test NuleculeBase get_provider"""
+    def test_get_provider_success(self):
+        """
+        Test if get_provider method when passed a particular valid key returns
+        the corresponding class.
+        """
+        nb = NuleculeBase(params = [], basepath = '', namespace = '')
+        provider_key = u'openshift'
+        # method `get_provider` will read from this config, we give it here
+        # since we have neither provided it before nor it is auto-generated
+        nb.config = {u'general': {u'provider': provider_key}}
+
+        return_provider = mock.Mock()
+        # mocking return value of method plugin.getProvider,because it returns
+        # provider class and that class gets called with values
+        nb.plugin.getProvider = mock.Mock(return_value=return_provider)
+        ret_provider_key, ret_provider = nb.get_provider()
+        self.assertEqual(provider_key, ret_provider_key)
+        return_provider.assert_called_with({u'provider': provider_key}, 
+                                            '', False)
+
+    def test_get_provider_failure(self):
+        """
+        Test if get_provider method when passed an invalid key raises an
+        exception.
+        """
+        nb = NuleculeBase(params = [], basepath = '', namespace = '')
+        # purposefully give the wrong provider key
+        provider_key = u'mesos'
+        nb.config = {u'general': {u'provider': provider_key}}
+        with self.assertRaises(NuleculeException):
+            nb.get_provider() 
--- a/tests/units/test_plugin.py
+++ b/tests/units/test_plugin.py
@@ -1,0 +1,27 @@
+import mock
+import unittest
+
+from atomicapp.plugin import Plugin
+ 
+class TestPluginGetProvider(unittest.TestCase):
+ 
+    """Test Plugin getProvider"""
+    def test_getProvider(self):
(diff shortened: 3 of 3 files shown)


atomicapp/nulecule/lib.py
```diff
@@ -1,11 +1,18 @@
 # -*- coding: utf-8 -*-
+import sys
+
+import logging
+
+from atomicapp import set_logging
```
dustymabe: where is this used?

atomicapp/plugin.py
```diff
@@ -164,3 +166,6 @@ def getProvider(self, provider_key):
             if key == provider_key:
                 logger.debug("Found provider %s", provider)
                 return provider
+        # if 'provider_key' matches to None of the supported providers
+        raise NuleculeException("Invalid Provider - '{}', provided in answers.conf (choose from "
+                                "'docker', 'kubernetes', 'openshift', 'marathon')".format(provider_key))
```
dustymabe: try to keep lines to below 80 characters.

atomicapp/nulecule/lib.py
```diff
@@ -94,7 +101,11 @@ def get_provider(self, provider_key=None, dry=False):
         # If provider_key isn't provided via CLI, let's grab it the configuration
         if provider_key is None:
             provider_key = self.config.get(GLOBAL_CONF)[PROVIDER_KEY]
-        provider_class = self.plugin.getProvider(provider_key)
+        try:
+            provider_class = self.plugin.getProvider(provider_key)
+        except NuleculeException as e:
+            logger.error(e)
+            sys.exit(1)
```
dustymabe: I don't really know of another place we exit directly from the code like this rather than raising an exception. I'll have to ask the other guys for review.

dustymabe:
@rtnpro @cdrage can you review this? Some questions to answer:

Should this issue be fixed in this way? Is there a better way? Should we sys.exit() straight from the code as is done here?

dustymabe:
@rtnpro should we also write a test for this?

atomicapp/plugin.py
surajssd: Acknowledged

atomicapp/nulecule/lib.py
surajssd: Exiting here won't make much harm because no container has been pulled yet or no actual processing has been started, its just rendering and parsing of configuration files.

Correct me if I am wrong.

atomicapp/nulecule/lib.py
```diff
@@ -1,11 +1,18 @@
 # -*- coding: utf-8 -*-
+import sys
+
```
rtnpro: Unnecessary new line

atomicapp/nulecule/lib.py
```diff
@@ -1,11 +1,18 @@
...
+import logging
+
```
rtnpro: We can have a new line between stdlib imports and atomicapp, for readability :+1:

atomicapp/nulecule/lib.py
rtnpro: @dustymabe You're right! We do most exception handling in ``atomicapp/cli/main.py``. We don't need to put a ``try-except`` block around ``provider_class = self.plugin.getProvider(provider_key)`` because the ``NuleculeException`` raised by ``getProvider`` will be automatically handled in ``cli/main.py``.

@surajssd can you remove the ``try...except`` block here?

atomicapp/nulecule/lib.py
rtnpro: @surajssd remove unused import :)

rtnpro:
@surajssd could you have a look at https://travis-ci.org/projectatomic/atomicapp/builds/109825603? I think with the suggested refactor, it should be OK.

atomicapp/plugin.py
dustymabe: Also if anything we should be using the `PROVIDERS` constant from `constants.py` to show them options if we choose to do so.

rtnpro:
@surajssd Mind adding unittests for ``getProvider``?

atomicapp/plugin.py
```diff
@@ -164,3 +166,8 @@ def getProvider(self, provider_key):
             if key == provider_key:
                 logger.debug("Found provider %s", provider)
                 return provider
+        # if 'provider_key' matches to None of the supported providers
+        raise NuleculeException("Invalid Provider - '{}', provided in "
+                                "answers.conf (choose from 'docker', "
+                                "'kubernetes', 'openshift', 'marathon')"
+                                .format(provider_key))
```
dustymabe: If we choose to tell them what options they can use then we should be using the PROVIDERS constant from constants.py and not a hardcoded list.

surajssd: yes can do that!

atomicapp/plugin.py
```diff
@@ -166,3 +167,10 @@ def getProvider(self, provider_key):
             if key == provider_key:
                 logger.debug("Found provider %s", provider)
                 return provider
+
+        # importing this on top of file leads to circular import
+        from atomicapp.nulecule.exceptions import NuleculeException
```
rtnpro: local imports are kind of hacks, and I prefer to avoid them. Let's not reference ``Nulecule*`` things in ``plugin.py`` at all, because, ``plugin.py`` is a dependency for ``Nulecule*`` and other things. Does it make sense?

surajssd: Yes makes sense, but then what exception do I raise here?

rtnpro: So, as I had said earlier, let's raise an exception in ``nulecule/main.py`` when ``provider is None``.

rtnpro: No need to raise any exception in ``plugin.py``. It returns ``None`` because it's not able to find a provider. Simple, isn't it?

surajssd: yes makes sense!! :)

rtnpro:
@surajssd One last thing, could you add a unittest for ``get_provider()``,  https://github.com/projectatomic/atomicapp/pull/563/files#diff-093821a6e6f49d13d49fdcb4f911695dR91, in ``tests/units/nulecule/test_lib.py``?

rtnpro:
LGTM :+1: @dustymabe shall we merge it in to ``master``?

dustymabe:
@rtnpro I trust you. :) - merge away.

dustymabe:
I think we need to do a better job of commenting the tests. Without knowing mock I should be able to read through the code and get a high level understanding of what it is testing.

surajssd:
Yes sure, its harder for somebody not knowing mock to understand what is actually going on!

dustymabe:
@surajssd - rebase on upstream master so we can merge.. 

bonus points if you comment the tests :)

surajssd:
@dustymabe I have added comments, was little confused about what all to include into them, share your views.

dustymabe:
@cdrage can you merge this if tests pass and you are satisfied with everything?

cdrage:
Tests pass and code looks good. Thank you for the extensives tests by the way! :dancer:

surajssd:
:relaxed: thanks for the complete explanation at each point.

Issue #571, opened 2016-02-23
Extended paths in atomicapp --verbose ... ?


containscafeine:
Currently, in the third column, only the file name is shown. Would it not be easier to debug with at least one directory up shown there? Something like, `nulecule/main.py` or `cli/main.py`?
>
[vagrant@centos7-adb atomicapp]$ sudo atomicapp --verbose  run foo --provider docker
2016-02-22 09:24:58,734 - [INFO] - main.py - Action/Mode Selected is: run
2016-02-22 09:24:58,734 - [DEBUG] - main.py - Final parsed cmdline: run --verbose foo --provider docker
2016-02-22 09:24:58,735 - [DEBUG] - main.py - NuleculeManager init app_path: /var/lib/atomicapp/foo-3530e98ba97c
2016-02-22 09:24:58,735 - [DEBUG] - main.py - NuleculeManager init    image: foo
2016-02-22 09:24:58,735 - [DEBUG] - main.py - Request to unpack to foo to /var/lib/atomicapp/foo-3530e98ba97c
2016-02-22 09:24:58,735 - [INFO] - base.py - Unpacking image: foo to /var/lib/atomicapp/foo-3530e98ba97c
2016-02-22 09:24:58,862 - [INFO] - container.py - Pulling Docker image: foo
2016-02-22 09:24:58,862 - [DEBUG] - container.py - /usr/bin/docker pull foo
...
...

containscafeine:
I could come up with 2 ways to do this.

1. Something like what's done in applogging.py [here](https://github.com/projectatomic/atomicapp/blob/master/atomicapp/applogging.py#L27)
Say,  logging.LogRecord handles the 'pathname', so inherit it and change it there to display the desired output.

2. Do it using a [LoggerAdapter](https://docs.python.org/2/library/logging.html#loggeradapter-objects), using which, in the future we can also specify custom application specific logs too. But how to go about it? If we make a LoggerAdapter in applogging.py, then how to make other files access it for logging?

@rtnpro @dustymabe @cdrage: any pointers? Maybe there is something much simpler and I do not happen to think about it.

dustymabe:
To be honest I don't really want to add any more output to the line at risk of it being too verbose. even if a filename is ambiguous (there is more than one file with that name) it is still pretty easy to find which of the 2 or 3 possibilities it is.

containscafeine:
@dustymabe How about adding it only to DEBUG logs?

dustymabe:
[...]

PR #572, opened 2016-02-24
Implement LoggerAdapter to provide extended paths in the output. Improve #571

containscafeine:
screenshot from 2016-02-24 22-23-17

A logger adapter has been created in applogging.py, and the logging instance created by it is the same which is being used by other files, unlike before when each file created their own instance.

This is done so that additional contextual output can be outputted in the logging messages, which is currently only the filename which is making the logging calls.

In the other files, only one import (`from atomicapp.applogging import Logging`) and one initialization (`logger = Logging.global_logger(__file__)`), and it's done.

The code is poor, but it would be great if I could get some feedback on the approach and where can it be improved.

Also, I have not touched the cockpit logger and not included the check for --verbose, will do once this is reviewed.

P.S. Sorry for the changes around imports, the IDE rearranged them according to PEP8.

### DIFF
--- a/atomicapp/applogging.py
+++ b/atomicapp/applogging.py
@@ -71,8 +71,15 @@ class colorizeOutputFormatter(logging.Formatter):
         return input
 
 
-class Logging:
+class AtomicappLoggingAdapter(logging.LoggerAdapter):
+    """
+    A class to pass contextual information to logs.
+    """
+    def process(self, msg, kwargs):
+        return('{} : {}'.format(self.extra['atomicapp_extra'], msg), kwargs)
 
+
+class Logging:
     @staticmethod
     def setup_logging(verbose=None, quiet=None, logtype=None):
         """
@@ -101,20 +108,20 @@ class Logging:
 
         # Get the loggers and clear out the handlers (allows this function
         # to be ran more than once)
-        logger = logging.getLogger(LOGGER_DEFAULT)
-        logger.handlers = []
+        logger_raw = logging.getLogger(LOGGER_DEFAULT)
+        logger_raw.handlers = []
         cockpit_logger = logging.getLogger(LOGGER_COCKPIT)
         cockpit_logger.handlers = []
 
         if logtype == 'none':
             # blank out both loggers
-            logger.addHandler(logging.NullHandler())
+            logger_raw.addHandler(logging.NullHandler())
             cockpit_logger.addHandler(logging.NullHandler())
             return
 
         if logtype == 'cockpit':
             # blank out normal log messages
-            logger.addHandler(logging.NullHandler())
+            logger_raw.addHandler(logging.NullHandler())
 
             # configure cockpit logger
             handler = logging.StreamHandler(stream=sys.stdout)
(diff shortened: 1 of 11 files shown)


cdrage:
I know it's a pain, but is it possible to revert the changes made from the IDE? We can make the PEP8 changes after in a separate commit.

This will make the commit history a lot cleaner. 

Also, any reason why it's reporting the compiled files (.pyc) rather than .py?

Otherwise, unit + integration tests pass. I'll comment on the code after the commit is updated :)

containscafeine:
@cdrage
Removed the PEP-8 changes.

And as for .pyc files, I'm not sure. It sometimes shows .pyc and the other times .py (see image above), but should it matter since these are debug messages and we get the file names just fine?

atomicapp/applogging.py
```diff
@@ -71,8 +71,15 @@ def _make_unicode(self, input):
         return input
 
 
-class Logging:
+class AtomicappLoggingAdapter(logging.LoggerAdapter):
+    """
+    A class to pass contextual information to logs.
```
cdrage: Instead of "A class". Maybe "A logging adapter?" Bit more descriptive :)

atomicapp/applogging.py
```diff
@@ -142,11 +149,26 @@ def setup_logging(verbose=None, quiet=None, logtype=None):
...
             return
 
         # If we made it here then there is an error
         raise Exception("Invalid logging output type: {}".format(logtype))
+
+    @staticmethod
+    def global_logger(filename):
+        """
+        This function returns a logging instance which will output logging event information
+        along with what the LoggerAdapter tells it to output
```
cdrage: paraphrase, "along with LoggerAdapter output" instead of "along with what the LoggerAdapter tells it to output"

containscafeine:
@cdrage Thanks for the pointers. I'll work on the comments.
Could you review the code once you get time, so that I know whether to proceed with this approach or not.

cdrage:
@containscafeine this all looks good to me :+1: feel free to work on the other stuff.

I'll await for @dustymabe for secondary review

dustymabe:
@containscafeine - can you look at https://github.com/dustymabe/atomicapp/commit/e2c57cc8f81507a45f1a129c239238f687e0aaa2 and give me your input? I think solving this in a way that uses the facilities of the logging library (i.e. creating a custom formatter) is a better way.

containscafeine:
@dustymabe I knew I was unnecessarily complicating stuff, but on another note, at the moment we have to modify just the pathname, which we can deal with using customOutputFormatter, but what if we have to pass in some contextual information to the logs, say, the atomicapp name, we can modify the LoggerAdapter then.
For this particular issue, it's not required, but does it make sense later or do we just modify the logging messages?

dustymabe:
@containscafeine I would like to not modify our logging structure unless we have a need to do so. If you think having the longer filename is something you want in right now then I say we go with something like https://github.com/dustymabe/atomicapp/commit/e2c57cc8f81507a45f1a129c239238f687e0aaa2 for now.

What do you think?

containscafeine:
+1 for https://github.com/dustymabe/atomicapp/commit/e2c57cc8f81507a45f1a129c239238f687e0aaa2

PR #575, opened 2016-02-25
Install requirements before make test

surajssd:
In a clean environment where nothing is installed and if I run `make test` then the required modules should be installed first. So before `make test` added pip install from requirements.txt and test-requirements.txt.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,8 @@ install:
 	python setup.py install
 
 test:
+	pip install -qr requirements.txt
+	pip install -qr test-requirements.txt
 	python -m pytest -vv
 
 image:


rtnpro:
LGTM :+1: @surajssd Please update your commit message to include ``Fixed #issue-no`` so that the issue gets closed automatically when we merge this.

surajssd:
@rtnpro yes let @cdrage put his points.

dustymabe:
The only case where this would be bad is a setting where all of the requirements were installed via system packages (rpms) and not via pip.

As long as we are ok with that..  LGTM.

dustymabe:
@surajssd @rtnpro - please comment on previous message and then we can choose to merge or not.

surajssd:
The usage we are propagating is via `make` utility. And we are not encouraging the usage of rpm based python modules, because sometimes those are old and not updated very often.

dustymabe:
@surajssd - rebase to latest master and then we can merge.

dustymabe:
@surajssd - next time you comply with my request you might want to ping me :) - I don't get emails when new files are pushed up, but I do when comments are added. Just say "hey, I did that thing that you asked me to do".

surajssd:
@dustymabe thanks and will take care of it now on!

PR #583, opened 2016-03-02
logging: add more context to filename in verbose mode

dustymabe:
Previously where the filename would have been 'main.py' in the
log messages, it would now be 'cli/main.py' for verbose mode.


dustymabe:
This is the successor to #572

cdrage:
Commit message = good
Code = good
Tests = super good passing

:+1: and LGTM! Merging.

Issue #591, opened 2016-03-04
No yum update in Dockerfile


surajssd:
Why we don't have a `yum update` in the default [`Dockerfile`](https://github.com/projectatomic/atomicapp/blob/master/Dockerfile#L19)? What I percieve is that since these docker images are built and pushed everyday, that could be the one reason.

PR #592, opened 2016-03-04
Added yum update to Dockerfile

surajssd:
Added `yum update` in Dockerfile of project root directory.

Fixes issue #591

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,7 @@ ADD requirements.txt ./
 
 # Install needed requirements
 RUN yum install -y epel-release && \
+    yum update -y && \
     yum install -y --setopt=tsflags=nodocs docker && \
     yum install -y --setopt=tsflags=nodocs $(sed s/^/python-/ requirements.txt) && \
     yum clean all


goern:
shouldnt that be the last command before `yum clean all`?

surajssd:
@goern why is that so?

goern:
maybe you are right and it doesnt matter, and is just a matter of habit...

cdrage:
As long as there is a `yum clean all` at the very end i don't see an issue..

surajssd:
I am still not very clear about what is correct way to go?

cdrage:
yum update at the top or the bottom shouldn't matter since the image size will still remain the same :+1: this LGTM

surajssd:
okay now its clear to me :)

dustymabe:
I think some of the consensus recently around the yum update vs not is that if you are basing your image on a source image (centos/rhel/debian/etc) that gets properly updated regularly then you shouldn't need to do an update yourself. 

Either way, it is fine if we want to go that approach

dustymabe:
@kbsingh - do you have an guidance on this issue (and more specifically on the previous comment). Do you have any strong opinions on whether we should be using yum update -y in our Dockerfiles vs trusting the container we are using as a base?

bexelbie:
@surajssd Consider the comment about updates here:

https://github.com/projectatomic/container-best-practices/blob/9ac2688de11d5576090b6d97d04ba2454e423bf5/creating/creating_index.adoc

I believe that we should just figure out what the stated update pattern is for the CentOS base and follow through with that.  I believe we are going to find it something we can rely on.

surajssd:
@bexelbie I could not find any written guidelines, but looking at the [Fedora-Dockerfiles](https://github.com/fedora-cloud/Fedora-Dockerfiles) I feel its widely accepted convention to update.

bexelbie:
@surajssd "Updates - Not all community images are necessarily rebuilt on a regular basis or when security vulnerabilities are addressed. You should therefore consider using base images from "official repositories" on Docker Hub, and confirm their update policy in advance."

surajssd:
@bexelbie so that means its a safe thing to update right?

bexelbie:
@surajssd as I understand it, reproducibility and in some ways thinking about what works comes down to not running `yum update`.  If Fedora has an acceptable update policy, I think we should not add `yum update`.  If they are not updating their container we should both do a `yum update` and push on them :)

surajssd:
Who is the right guy from CentOS community to comment on this? Because I was not able to find anything documented in this regard.

jperrin:
The base centos containers are rebuilt monthly, and that cadence may increase. Putting a ```yum -y update``` into the Dockerfile will in many cases be a no-op. If the containers are being properly refreshed and rebuilt I don't see a need to include the update.

cdrage:
Thanks for the inisight @jperrin ! With your comment I shall close this issue. We'll rely on CentOS Docker images being up-to-date instead of `yum update`

Issue #581, opened 2016-03-02
Need better error output for YAML formatting issues


jberkus:
So I had an issue with some of my YAML formatting, and got this:

[josh@redhat atomic-site-master]$ sudo atomicapp run Nulecule
[sudo] password for josh:
2016-03-02 11:58:42,488 - [INFO] - main.py - Action/Mode Selected is: run
2016-03-02 11:58:42,494 - [ERROR] - main.py - AnyMarkupError: caught <class 'yaml.scanner.ScannerError'>: while scanning a simple key
  in "<file>", line 17, column 9
could not found expected ':'

  in "<file>", line 18, column 7
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/atomicapp-0.4.2-py2.7.egg/atomicapp/cli/main.py", line 101, in cli_run
    nm.run(**argdict)
  File "/usr/lib/python2.7/site-packages/atomicapp-0.4.2-py2.7.egg/atomicapp/nulecule/main.py", line 224, in run
    self.nulecule = self.unpack(dryrun=dryrun, config=self.answers)
  File "/usr/lib/python2.7/site-packages/atomicapp-0.4.2-py2.7.egg/atomicapp/nulecule/main.py", line 132, in unpack
    self.app_path, dryrun=dryrun, config=config)
  File "/usr/lib/python2.7/site-packages/atomicapp-0.4.2-py2.7.egg/atomicapp/nulecule/base.py", line 134, in load_from_path
    nulecule_data = anymarkup.parse_file(nulecule_path)
  File "/usr/lib/python2.7/site-packages/anymarkup/__init__.py", line 101, in parse_file
    return parse(f, format, encoding, force_types)
  File "/usr/lib/python2.7/site-packages/anymarkup/__init__.py", line 75, in parse
    raise AnyMarkupError(e)
AnyMarkupError: AnyMarkupError: caught <class 'yaml.scanner.ScannerError'>: while scanning a simple key
  in "<file>", line 17, column 9
could not found expected ':'
  in "<file>", line 18, column 7
[josh@redhat atomic-site-master]$

Issues:

1. That's not a user-friendly error, it's a stack trace.

2. It doesn't tell me which file errored, making it hard to debug where I typo'd.

dustymabe:
Looks like #553 - is that correct?

jberkus:
Yes, same issue.  We need the error to be trapped, and just the relevant portions forwarded to the user (file, line no).

[...]

Issue #553, opened 2016-02-11
yaml / json / ini checker of Nulecule. Beautify erroring out.


cdrage:
So when there's a JSON / YAML / INI error within AnyMarkup, we get this ugly error:
```
2016-01-19 14:45:43,159 - atomicapp.cli.main - ERROR - AnyMarkupError: caught <class 'yaml.scanner.ScannerError'>: mapping values are not allowed here
  in "<file>", line 17, column 13
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.3.1-py2.7.egg/atomicapp/cli/main.py", line 100, in cli_run
    nm.run(**argdict)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.3.1-py2.7.egg/atomicapp/nulecule/main.py", line 221, in run
    self.nulecule = self.unpack(dryrun=dryrun, config=self.answers)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.3.1-py2.7.egg/atomicapp/nulecule/main.py", line 126, in unpack
    nodeps=nodeps, dryrun=dryrun, update=update)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.3.1-py2.7.egg/atomicapp/nulecule/base.py", line 105, in unpack
    dryrun=dryrun, update=update)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.3.1-py2.7.egg/atomicapp/nulecule/base.py", line 130, in load_from_path
    nulecule_data = anymarkup.parse_file(nulecule_path)
  File "build/bdist.linux-x86_64/egg/anymarkup_core/__init__.py", line 114, in parse_file
    return parse(f, format, encoding, force_types)
  File "build/bdist.linux-x86_64/egg/anymarkup_core/__init__.py", line 88, in parse
    raise AnyMarkupError(e)
AnyMarkupError: AnyMarkupError: caught <class 'yaml.scanner.ScannerError'>: mapping values are not allowed here
  in "<file>", line 17, column 13
```

Let's beautify it up a bit so it's a bit less ambiguous.

PR #593, opened 2016-03-04
Error cleanly on missing Nulecule or invalid formatted Nulecule

cdrage:
This commit checks to see if the actual Nulecule exists before loading. If
the Nulecule file is missing it will cleanly error out.

If a Nulecule is wrongly formatted, atomicapp will exit cleanly with a line
number, column and the output of the invalid line.

ex.
``` 
2016-03-04 13:03:59,617 - [ERROR] - main.py - Failure parsing Nulecule
file. Validation error on line 12, column 23:
            {
```

Fixes issues:
#581 
#553

### DIFF
--- a/atomicapp/nulecule/base.py
+++ b/atomicapp/nulecule/base.py
@@ -133,7 +133,8 @@ class Nulecule(NuleculeBase):
         nulecule_path = os.path.join(src, MAIN_FILE)
 
         if os.path.exists(nulecule_path):
-            nulecule_data = open(nulecule_path, 'r').read()
+            with open(nulecule_path, 'r') as f:
+                nulecule_data = f.read()
         else:
             raise NuleculeException("No Nulecule file exists in directory: %s" % src)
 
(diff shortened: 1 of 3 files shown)


cdrage:
Tests pass :+1:

atomicapp/nulecule/base.py
```diff
@@ -128,10 +131,27 @@ def load_from_path(cls, src, config=None, namespace=GLOBAL_CONF,
             an image).
         """
         nulecule_path = os.path.join(src, MAIN_FILE)
+
+        if os.path.exists(nulecule_path):
+            nulecule_data = open(nulecule_path, 'r').read()
```
dustymabe: should we use the contextlib stuff here:

```
with open() as f:
  nulecule_data = f.read()
```

I think that would be good practice.

dustymabe:
@cdrage. nice work. I have a few questions:

- Do we think this is something that would actually be better to be implemented in Anymarkup itself, rather than in Atomic App? If so we could merge this like is, but open a PR on Anymarkup and let the functionality live there once it is merged.
- Would it be useful to give the full path to the file in the output? 
- Would it be useful to show more than one line of context and the line numbers on the left hand side?

cdrage:
@dustymabe 

Yes, this should be put upstream, I'll send in a patch.

Full path :+1: 

I'll try that and see how it looks like.

cdrage:
@dustymabe 
this looks better! :) updated the PR for another review.
```sh
▶ atomicapp run .
2016-03-09 09:11:46,115 - [INFO] - main.py - Action/Mode Selected is: run
2016-03-09 09:11:46,118 - [ERROR] - main.py - Failure parsing ./Nulecule file. Validation error on line 10, column 5:
9     - name: helloapache-app
10     params:
11       - name: image
```

dustymabe:
#dotests

dustymabe:
Nice work Charlie. The only concern I have is: is it possible that an error on the first or last line of the yaml file would throw off the output somehow. If that works just fine (and the tests pass that are in progress now) then LGTM.. merge away.

cdrage:
Awesome :D I'll merge this in.

PR #595, opened 2016-03-04
Remove extra newline on main.py

cdrage:
I get two newlines when using `atomicapp`. I think it should just be
one!


cdrage:
Actually, nevermind this, going to open up a different PR for this.

PR #596, opened 2016-03-04
Use logger for print output in main.py

cdrage:
Instead of print, use logger.info(). Using print creates uneeded
logging for cockpit and makes the overall logging more cleaner.

Although may need quick discussion since it's less obvious to the user
when this is outputted (before it was a new-line print with no [INFO]
beside it). Maybe output via logger but without the data / info /
cli.mainpy information?)

### DIFF
--- a/atomicapp/cli/main.py
+++ b/atomicapp/cli/main.py
@@ -46,8 +46,8 @@ logger = logging.getLogger(LOGGER_DEFAULT)
 def print_app_location(app_path):
     if app_path.startswith(HOST_DIR):
         app_path = app_path[len(HOST_DIR):]
-    print("\nYour application resides in %s" % app_path)
-    print("Please use this directory for managing your application\n")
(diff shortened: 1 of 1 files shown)


dustymabe:
I honestly don't know how I feel about this. Part of the point of using the print statement was to separate the output from the rest so that the user would pay attention to it. I like the fact that it is separated now, but would like for the team to agree or disagree with me and we can choose which way forward is best.

@rtnpro @kadel what do you think?

kadel:
Hmm....
I don't know. Information about directory where application resides is more important for user than other INFO level messages, I'm afraid that it is going to be little lost in other messages.

Outputting this without level information might make this better.

cdrage:
@dustymabe @kadel 

Looks like the decision has been made! I just needed some insight into this. But I agree. This differentiates from the logs and tells the user where the files are, etc.

PR #361, opened 2015-10-27
Adds nice error message for Docker client/server out of sync. Closes #174

dustymabe:

Issue #174, opened 2015-07-15
Atomicapp image fails with 1.7.0 : client and server don't have same version (client : 1.19, server: 1.18)


sub-mod:
If you have created a atomicapp image recently .
It will download docker 1.7.0 and it gives the below error.
I have tried to overcome that issue by putting docker-1.6.0 in yum command  for now.
@goern @aweiteka @vpavlin 

 /home/development/atomicapp-run/atomicapp.sh -v --answers=/home/development/atomic_jsonbrooks/nulecule/examples/tmp/answers.conf run submod/guestbookphp-app
2015-07-15 14:19:19,496 - atomicapp.install - INFO - App name is submod/guestbookphp-app, will be populated to /atomicapp
2015-07-15 14:19:19,496 - atomicapp.utils - INFO - atomicapp.status.info.message=Loading app submod/guestbookphp-app .
2015-07-15 14:19:19,497 - atomicapp.utils - INFO - atomicapp.status.answer.message={"general": {"namespace": "default", "provider": "kubernetes"}}
2015-07-15 14:19:19,497 - atomicapp.nulecule_base - DEBUG - {'namespace': 'default', 'provider': 'kubernetes'}
Error response from daemon: client and server don't have same version (client : 1.19, server: 1.18)
Traceback (most recent call last):
  File "/usr/bin/atomicapp", line 9, in <module>
    load_entry_point('atomicapp==0.1.9', 'console_scripts', 'atomicapp')()
  File "/usr/lib/python2.7/site-packages/atomicapp-0.1.9-py2.7.egg/atomicapp/cli/main.py", line 101, in main
    cli.run()
  File "/usr/lib/python2.7/site-packages/atomicapp-0.1.9-py2.7.egg/atomicapp/cli/main.py", line 82, in run
    args.func(args)
  File "/usr/lib/python2.7/site-packages/atomicapp-0.1.9-py2.7.egg/atomicapp/cli/main.py", line 22, in cli_run
    ae = Run(**vars(args))
  File "/usr/lib/python2.7/site-packages/atomicapp-0.1.9-py2.7.egg/atomicapp/run.py", line 63, in __init__
    install.install()
  File "/usr/lib/python2.7/site-packages/atomicapp-0.1.9-py2.7.egg/atomicapp/install.py", line 99, in install
    self.nulecule_base.pullApp()
[...]

PR #600, opened 2016-03-07
Handle docker error, 'raise' instead of 'print', fix #570

### LINKED ISSUE #570: Error not handled - docker daemon not running
This is how `atomicapp` errors out when the docker daemon is not running.

> [vagrant@centos7-adb atomicapp]$ sudo atomicapp --verbose  run foo --provider docker
2016-02-22 09:32:57,763 - [INFO] - main.py - Action/Mode Selected is: run
2016-02-22 09:32:57,763 - [DEBUG] - main.py - Final parsed cmdline: run --verbose foo --provider docker
2016-02-22 09:32:57,763 - [DEBUG] - main.py - NuleculeManager init app_path: /var/lib/atomicapp/foo-a45442b82021
2016-02-22 09:32:57,763 - [DEBUG] - main.py - NuleculeManager init    image: foo
2016-02-22 09:32:57,764 - [DEBUG] - main.py - Request to unpack to foo to /var/lib/atomicapp/foo-a45442b82021
2016-02-22 09:32:57,764 - [INFO] - base.py - Unpacking image: foo to /var/lib/atomicapp/foo-a45442b82021
2016-02-22 09:32:57,800 - [ERROR] - main.py - Command '['/usr/bin/docker', 'version']' returned non-zero exit status 1
  File "/home/vagrant/atomicapp/atomicapp/cli/main.py", line 101, in cli_run
    nm.run(**argdict)
  File "/home/vagrant/atomicapp/atomicapp/nulecule/main.py", line 224, in run
    self.nulecule = self.unpack(dryrun=dryrun, config=self.answers)
  File "/home/vagrant/atomicapp/atomicapp/nulecule/main.py", line 129, in unpack
    nodeps=nodeps, dryrun=dryrun, update=update)
  File "/home/vagrant/atomicapp/atomicapp/nulecule/base.py", line 103, in unpack
    docker_handler = DockerHandler(dryrun=dryrun)
  File "/home/vagrant/atomicapp/atomicapp/nulecule/container.py", line 37, in \__init\__
    raise e
CalledProcessError: Command '['/usr/bin/docker', 'version']' returned non-zero exit status 1

[Here](https://github.com/projectatomic/atomicapp/blob/master/atomicapp/nulecule/container.py#L27-#L37) is the culprit, but we have already handled `subprocess.CalledProcessError` for a different error.
The traceback can be handled with #569, but how do we distinguish between different errors? How do we make sense out of the docker CLI output and use it? I just tried the switch `--format`, but it does not seem to format errors.

Please correct if I'm talking gibberish.

containscafeine:


atomicapp/nulecule/container.py
```diff
@@ -31,10 +31,14 @@ def __init__(self, dryrun=False, docker_cli='/usr/bin/docker'):
...
-                          "Atomic App version for this host.\n")
-                raise e
+                    raise DockerException("\nThe docker version in this Atomic App "
+                                          "differs greatly from the host version.\n"
+                                          "Please use a different Atomic App version "
+                                          "for this host.\n")
+                elif "Is your docker daemon up and running" in e.output or \
+                     "Are you trying to connect to a TLS-enabled daemon without TLS" \
+                     in e.output:
+                    raise DockerException("Could not connect to the docker daemon.")
```
cdrage: Should we interpret the two above sentences to a summary of: "Count not connect to the docker daemon?" May be better just to output `e`?

containscafeine: @cdrage +1, it makes the code a lot cleaner and also more verbose to the end user.
Do you mean something like:
```python
        if not dryrun:
            try:
                subprocess.check_output([docker_cli, 'version'],
                                        stderr=subprocess.STDOUT)
            except subprocess.CalledProcessError as e:
                raise DockerException(e.output)
```
The output for this is,
```bash
...
...
2016-03-08 01:08:16,441 - [DEBUG] - nulecule/main.py - NuleculeManager init image: foo
2016-03-08 01:08:16,441 - [DEBUG] - nulecule/main.py - Request to unpack to foo to /var/lib/atomicapp/foo-d48b86bf8730
2016-03-08 01:08:16,441 - [INFO] - nulecule/base.py - Unpacking image: foo to /var/lib/atomicapp/foo-d48b86bf8730
2016-03-08 01:08:16,486 - [ERROR] - cli/main.py - Client:
 Version:      1.8.2-el7.centos
 API version:  1.20
 Package Version: docker-1.8.2-10.el7.centos.x86_64
 Go version:   go1.4.2
 Git commit:   a01dc02/1.8.2
 Built:        
 OS/Arch:      linux/amd64
Get http:///var/run/docker.sock/v1.20/version: dial unix /var/run/docker.sock: no such file or directory.
* Are you trying to connect to a TLS-enabled daemon without TLS?
* Is your docker daemon up and running?
```

cdrage: @containscafeine 
yes! that's much better. since Docker may change those sentences / texts in the future and screw up the logging output as well as give the user a bit more information as to why Docker isn't working :)

dustymabe: So there is a bit of history behind this. See https://github.com/projectatomic/atomicapp/issues/174 and then the PR https://github.com/projectatomic/atomicapp/pull/361. I think the idea was to give the user of atomicapp a message that helped them understand that the version of docker within the atomicapp container differed from the host and what they needed to do to resolve that.

Just spitting out e.output doesn't really help them with that particular problem. If there are problems that could be specific to atomic apps interaction with docker then it would do our users some good to have a more helpful message delivered to them.

containscafeine:
@dustymabe agreed, but till the API/docker-py is implemented, there could be a hundred other errors while interacting with docker besides version mismatch. Instead of sending out only an atomicapp specific error message for each one of those, is it not better to display e.output for all the errors plus atomicapp specific output for the issues we know about, like the version mismatch.
Something like `e.output` + `docker version in atomicapp is different ...`
Because @cdrage has a point that we don't know when the error output changes breaks the code.

cdrage:
@containscafeine @dustymabe 

If it's good with you, I'm okay with @containscafeine commit for raising the exception.

Once we eventually switch to the Docker API we can parse each error code from the API and output the respective message.

dustymabe:
@containscafeine 

I think if we know the problem then we shouldn't add more output that the user has to parse through. We should certainly print out e.output when we don't properly identify a problem, but otherwise we should give the user a helpful message. 

If the output from the docker daemon changed and our string matching didn't work anymore then the user would still get e.output right?

containscafeine:
@dustymabe +1

So,
if we know the error:
    custom message
else if we don't know the error:
    e.output

Do I move ahead with this @cdrage @dustymabe?

dustymabe:
@containscafeine +1

cdrage:
@containscafeine That seems good to me.

containscafeine:
Hey, thanks @centos-ci!
@cdrage @dustymabe is this better now?

dustymabe:
@containscafeine looks better to me. I'm running tests now. If @cdrage approves then we'll merge

dustymabe:
#dotests

dustymabe:
ok to test

dustymabe:
#dotests

containscafeine:
Oh @centos-ci, I'll miss you once you're fixed! :)

cdrage:
Tests pass, LGTM :+1:

PR #602, opened 2016-03-07
Multiple problems with issuestats.com

cdrage:
For the past two weeks, issuestats have been having multiple problems
with uptime as well as API calls.

(even right now at this moment: 03/07/2016 18:13 UTC each page load is
more than 3000ms.

Do we still want these on our README?


cdrage:
See https://github.com/hstove/issue_stats/issues/41

It also looks bad when the README.md load and some artifacts are broken :(

dustymabe:
@cdrage I'm +1. I trust your judgement.

PR #603, opened 2016-03-07
Added 'init' command to initialize a new atomic app.

rtnpro:
Implement ``init`` command in ``atomicapp`` CLI to initialize a new atomic app project.


cdrage:
When I was thinking of `atomicapp init` I thought it was intended to be an actual full-on example. (ex. using helloapache or something in there). Similar to how `nomad init` does it with their examples (github.com/hashicorp/nomad)

If it's just a templated file, it's going to still be confusing for the initial user.

rtnpro:
@cdrage I agree! However, I like this approach which is similar to ``django-admin.py createproject``. This creates boilerplate code that you can start working on. However, I do agree that we need to add boilerplate examples, which I plan to implement as follows:
``atomicapp init --example helloapache ./helloapache``

Does it make sense?

cdrage:
Maybe @dustymabe can chime in. 

But in my opinion, by __default__ `atomicapp init` should output an example. `atomicapp init --template foobar` should output scaffolding.

dustymabe:
> But in my opinion, by default atomicapp init should output an example. atomicapp init --template foobar should output scaffolding.

@cdrage +1. So we still get both, but the very first interaction a user has will probably be for them to demo creating an application. We should start them with a working example. After that they can use init --template to create a template of an application they want to build.

@rtnpro do you agree that making init output a working application would be desirable as a first interaction for the user?

rtnpro:
@dustymabe @cdrage I agree :+1:

rtnpro:
#dotests unittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotests

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

dustymabe:
#dotests

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
#dotestunittest

rtnpro:
@dustymabe @cdrage 

I see 2 problems:
1. the ``global`` params of ``atomicapp`` command does not make sense for ``init`` command.
1. Also, there's no point to attaining the ``lock`` which requires ``superuser`` access for running the init command.

Problem ``2`` can be easily solved. However, I am not sure how to best resolve the problem ``1``.

cdrage:
@rtnpro 

If it doesn't make sense using global params then don't implement them for now, since they *just* got merged into the Nulecule spec.

I'd just focus on adding the example (helloapache) to atomicapp init as well as create a good template for people to use.

dustymabe:
> 1. the global params of atomicapp command does not make sense for init command.

I assume you mean "global options", like --provider-api, etc. I say we either just ignore them if someone provides them. The only other option is to restructure our parsers such that we meticulously define what options applies to what subcommand. This would work best if we define reused options in a common place and then just call out to them for each subparser. 

Is that what you meant? 

> 1. Also, there's no point to attaining the lock which requires superuser access for running the init command.

Maybe we should just remove locking for now until it becomes a pain point for us. Locking has caused quite a bit of trouble for other things in the past.

rtnpro:
On Mon, Mar 28, 2016 at 7:27 PM, Dusty Mabe <notifications@github.com> wrote:
> the global params of atomicapp command does not make sense for init command.
>
> I assume you mean "global options", like --provider-api, etc. I say we
> either just ignore them if someone provides them. The only other option is
> to restructure our parsers such that we meticulously define what options
> applies to what subcommand. This would work best if we define reused options
> in a common place and then just call out to them for each subparser.
>
> Is that what you meant?
Yes!


> Also, there's no point to attaining the lock which requires superuser access
> for running the init command.
>
> Maybe we should just remove locking for now until it becomes a pain point
> for us. Locking has caused quite a bit of trouble for other things in the
> past.
OK


-- 
Ratnadeep Debnath,
https://www.waartaa.com
GPG Fingerprint: 033C 8041 A0E9 CDBA 2E02  B785 2119 5486 F245 DFD6

atomicapp/nulecule/main.py
```diff
@@ -102,6 +103,40 @@ def __init__(self, app_spec, destination=None,
         self.answers_file = answers_file
         self._process_answers()
 
+    @staticmethod
```
dustymabe: A few things:
- Definitely needs comments as to what is going on
- Should we break some of this function up into smaller functions?

atomicapp/templates/nulecule/answers.conf.sample.tpl
```diff
@@ -0,0 +1,7 @@
+[general]
+namespace = default
+provider = kubernetes
+
+[${app_name}]
+image = centos/httpd
+
```
dustymabe: Do we need a template answers.conf.sample file? Since this should be generated by atomicapp from the Nulecule I see no reason why we would need one.

dustymabe:
@rtnpro, just looked at the PR. I thought we agreed to have `init` output a working example and to output scaffolding if we give something different. So `atomicapp init` could output a helloapache example, but `atomicapp init myapp` would output scaffolding or something like that?

cdrage:
@rtnpro @dustymabe 
Yeah, I thought that too.

should we do `atomicapp init myapp` or `atomicapp init --template myapp`?

atomicapp/templates/nulecule/answers.conf.sample.tpl
rtnpro: :+1:

dustymabe:
@rtnpro and I had a discussion about this. for now we think the following is what we will target for this PR:

```
atomicapp init myapp
- creates 'myapp' directory, alternatively use directory provided by --destination
- gives you scaffolding
- gives you working application (uses helloapache container)
```

so init of myapp will give you scaffolding as well as a runnable helloapache example. I think this solves our concerns @cdrage.. Agree?

rtnpro:
We need to find a better way to re organize the CLI options, I'd suggest that we do it in another PR.

cdrage:
@dustymabe @rtnpro 

I still believe "atomicapp init" outputs a working example. Similar to how hashicorp/nomad does it. Gives the user a working example that they can atomicapp init and then atomicapp run.

The way you describe it sounds good. But if you just say `atomicapp init` what does that do?

dustymabe:
> So what you describe sounds good. But if you just say atomicapp init what does that do? Is the "myapp" part just the name of the dir it will extract to?

It will create an application named myapp, inside a folder named myapp, with a server named myapp, but it will use the apache image so that it is runnable out of the box.

cdrage:
@dustymabe alright cool, that sounds good to me. then people can do `atomicapp init helloworld` and have an app named helloworld.  :+1:

rtnpro:
Cool, I think that this PR is merge ready.

cdrage:
Is --help suppose to add this much data / noise ?
```
▶ atomicapp init --help
usage: atomicapp init [-h] [-V] [-v] [-q] [--mode {fetch,run,stop,genanswers}]
                      [--dry-run] [--answers-format {ini,json,xml,yaml}]
                      [--namespace NAMESPACE]
                      [--providertlsverify {True,False}]
                      [--providerconfig PROVIDERCONFIG]
                      [--providercafile PROVIDERCAFILE]
                      [--providerapi PROVIDERAPI]
                      [--logtype {cockpit,color,nocolor,none}]
                      [--destination DESTINATION]
                      app_name

positional arguments:
  app_name              App name.

optional arguments:
  -h, --help            show this help message and exit
  -V, --version         Show the version and exit.
  -v, --verbose         Verbose output mode.
  -q, --quiet           Quiet output mode.
  --mode {fetch,run,stop,genanswers}
                        The mode Atomic App is run in. This option has the
                        effect of switching the 'verb' that was passed by the
                        user as the first positional argument. This is useful
                        in cases where a user is not using the Atomic App cli
                        directly, but through another interface such as the
                        Atomic CLI. EX: `atomic run <IMAGE> --mode=genanswers`
  --dry-run             Don't actually call provider. The commands that should
                        be run will be logged but not run.
  --answers-format {ini,json,xml,yaml}
                        The format for the answers.conf.sample file. Default:
                        ini
```

atomicapp/nulecule/main.py
```diff
@@ -102,6 +103,48 @@ def __init__(self, app_spec, destination=None,
...
+            app_version=app_version,
+            app_desc=app_desc)
+        tmpdir = tempfile.mkdtemp(prefix='nulecule-new-app-')
+        template_dir = os.path.join(os.path.dirname(__file__),
+                                    '../templates/nulecule')
+        if destination is None:
+            destination = os.path.join('.', app_name)
+
+        if os.path.exists(destination) and os.path.isdir(destination) and os.listdir(destination):
+            value = raw_input('Destination directory is not empty! Do you still want to proceed? (y/n): ')
```
cdrage: Y should be default, no? Simlar to other apps?

Ex: `[Y]/n`

cdrage:
I can't seem to use this PR:
```
▶ atomicapp init test    
[INFO] - main.py - Action/Mode Selected is: init
[ERROR] - main.py - cannot copy tree '/usr/local/lib/python2.7/dist-packages/atomicapp-0.4.5-py2.7.egg/atomicapp/nulecule/../templates/nulecule': not a directory
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.4.5-py2.7.egg/atomicapp/cli/main.py", line 107, in cli_init
    NuleculeManager.init(argdict['app_name'], argdict['destination'])
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.4.5-py2.7.egg/atomicapp/nulecule/main.py", line 143, in init
    distutils.dir_util.copy_tree(template_dir, tmpdir)
  File "/usr/lib/python2.7/distutils/dir_util.py", line 128, in copy_tree
    "cannot copy tree '%s': not a directory" % src
DistutilsFileError: cannot copy tree '/usr/local/lib/python2.7/dist-packages/atomicapp-0.4.5-py2.7.egg/atomicapp/nulecule/../templates/nulecule': not a directory
```

atomicapp/nulecule/main.py
rtnpro: yes

rtnpro:
@cdrage I think you are using an older codebase, the current ``143`` line is https://github.com/projectatomic/atomicapp/pull/603/files#diff-75b9e7f171687ed710472e8b5767cd36R143.

I could not reproduce the same.

atomicapp/nulecule/main.py
rtnpro: Fixed :+1:

cdrage:
@rtnpro Can you rebase against master? I believe that this is the issue.

cdrage:
Still getting the same issue even after a clean pull:
```
▶ atomicapp init hello
[INFO] - main.py - Action/Mode Selected is: init
Destination directory is not empty! Do you still want to proceed? [Y]/n: Y
[ERROR] - main.py - cannot copy tree '/usr/local/lib/python2.7/dist-packages/atomicapp-0.4.5-py2.7.egg/atomicapp/nulecule/../templates/nulecule': not a directory
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.4.5-py2.7.egg/atomicapp/cli/main.py", line 106, in cli_init
    NuleculeManager.init(argdict['app_name'], argdict['destination'])
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.4.5-py2.7.egg/atomicapp/nulecule/main.py", line 126, in init
    distutils.dir_util.copy_tree(template_dir, tmpdir)
  File "/usr/lib/python2.7/distutils/dir_util.py", line 128, in copy_tree
    "cannot copy tree '%s': not a directory" % src
DistutilsFileError: cannot copy tree '/usr/local/lib/python2.7/dist-packages/atomicapp-0.4.5-py2.7.egg/atomicapp/nulecule/../templates/nulecule': not a directory
```

rtnpro:
@cdrage could you try cleaning your ``*pyc`` files?

rtnpro:
@cdrage did you setup the code again after pull?

cdrage:
@rtnpro 

Yes, I did a
```
dropbox/dev/atomicapp  pr_603 ✔                                                                                                                                                                                                                         3h9m  
▶ sudo find . -name "*.pyc" -exec rm -rf {} \;
```

Also a 
```
sudo make clean
sudo make install
```

An no changes :( I'll try again with dist/lib files removed manually.

atomicapp/nulecule/main.py
```diff
@@ -102,6 +103,49 @@ def __init__(self, app_spec, destination=None,
...
+        if destination is None:
+            destination = os.path.join('.', app_name)
+
+        if os.path.exists(destination) and os.path.isdir(destination) and os.listdir(destination):
+            value = raw_input('Destination directory is not empty! Do you still want to proceed? [Y]/n: ')
+            value = value or 'y'
+            if value.lower() != 'y':
+                return
+
+        distutils.dir_util.copy_tree(template_dir, tmpdir)
```
cdrage: def an issue here within Debian. See other comment

cdrage:
@rtnpro 

Issue is most likely `python setup.py install` not including the template files.

Please see setup.py, particularly:
``
package_data={'atomicapp': ['providers/external/kubernetes/*.yaml']},
```

In order for you to include the .tpl files.

**Comment by rtnpro** [comment]
@cdrage , good catch! I never noticed, as I setup my development env using ``python setup.py develop``.

**Comment by dustymabe** [comment]
So just trying running `atomicapp init dustytest && cd dustytest && atomicapp run .` - I get something running but I can't access it because no service is created. @rtnpro I think we should add a service to this example app so that it is a fully functioning app. What do you think?

**Comment by rtnpro** [comment]
@dustymabe I added a k8s service for the newly created app.

@cdrage Does https://github.com/projectatomic/atomicapp/pull/603/commits/44530e13b3e1b452d5b3a53d2845360778f18418 fix the issue you reported?

**Comment by cdrage** [comment]
@rtnpro Yup. That'd do it.

Do you mind setting the folders as: `external/templates/nulecule/*.tpl` instead? In a later commit I will fix the `providers/external/kubernetes/*.yaml` so everything is in one folder.

**Comment by rtnpro** [comment]
@cdrage could we merge it?

File: atomicapp/external/templates/nulecule/Dockerfile.tpl [hunk 5]
```diff
@@ -0,0 +1,10 @@
+FROM projectatomic/atomicapp:0.4.1
```
cdrage: This is bad as when a new release comes out (0.4.5 as a prime example) it still uses 0.4.5. We should be replacing this variable with the most current version of Atomic App as listed in `constants.py`

atomicapp/external/templates/nulecule/Dockerfile.tpl
```diff
@@ -0,0 +1,10 @@
...
+
+MAINTAINER Your Name <you@email.com>
+
+LABEL io.projectatomic.nulecule.specversion="0.0.2" \
```
cdrage: Same goes here, should use whatever is in `constants.py`

atomicapp/external/templates/nulecule/Dockerfile.tpl
```diff
@@ -0,0 +1,10 @@
...
+      io.projectatomic.nulecule.providers="kubernetes, docker" \
+      Build="docker build --rm --tag test/$app_name-atomicapp ."
```
cdrage: what is this? i've never seen this before. this shouldn't be here as it's not common convention to have it.

cdrage: i don't see any examples on nulecule-library that use this.

atomicapp/external/templates/nulecule/Nulecule.tpl
```diff
@@ -0,0 +1,21 @@
+---
+specversion: 0.0.2
```
cdrage: pull from constants.py

atomicapp/external/templates/nulecule/answers.conf.sample.tpl
```diff
@@ -0,0 +1,7 @@
+[general]
+namespace = default
+provider = kubernetes
+
+[${app_name}]
+image = centos/httpd
+
```
cdrage: as what @dustymabe said, why do we have a answers.conf.sample.tpl template? we should be auto-generating this, not using a template.

cdrage: for ex. if our test example changes in the future.

cdrage:
@rtnpro 

Is there no notification for when "atomicapp init" is created? The only output I get is:

```
▶ atomicapp init test
[INFO] - main.py - Action/Mode Selected is: init
```

Otherwise, please see other comments as I was able to now test this on my local machine.

atomicapp/external/templates/nulecule/Dockerfile.tpl
dustymabe: @cdrage.. its not common but I have seen it before. @rtnpro I say let's not include it as it may confuse the user more than help.

atomicapp/external/templates/nulecule/answers.conf.sample.tpl
dustymabe: @cdrage.. @rtnpro and I had a discussion about this. I forget his exact argument but he was able to convince me that this is fine for now. The template is rather small and if we change the example in the future then we can easily change the template. 

@rtnpro do you remember our discussion? You may be able to reproduce some of it here.

atomicapp/nulecule/main.py
dustymabe: @rtnpro - I don't believe this has been done. Can we add comments and break the function up?

dustymabe:
It works for me now!!! 

> Is there no notification for when "atomicapp init" is created?

@cdrage I agree, it would be nice to give the user some output to let them know what happened.

@rtnpro can you rebase this on latest master? Also please see comments about breaking the init function up and adding comments.

dustymabe:
@rtnpro - depending on how much this progresses in the next 18 hours we might be able to merge for tomorrows release.

atomicapp/external/templates/nulecule/Dockerfile.tpl
rtnpro: :+1:

atomicapp/external/templates/nulecule/Dockerfile.tpl
rtnpro: :+1:

atomicapp/external/templates/nulecule/answers.conf.sample.tpl
rtnpro: @dustymabe @cdrage The point was that ``atomicapp init`` should be dead simple and not do Nulecule introspection and other complicated stuffs.

atomicapp/cli/main.py
```diff
@@ -365,6 +375,21 @@ def create_parser(self):
             help='The name of a container image containing an Atomic App.')
         gena_subparser.set_defaults(func=cli_genanswers)
 
+        # === "init" SUBPARSER ===
+        init_subparser = toplevel_subparsers.add_parser(
+            "init", parents=[globals_parser])
```
cdrage: why do we need global_parsers? for ex, we won't be using --provider-config for init

atomicapp/nulecule/main.py
```diff
@@ -120,6 +123,56 @@ def __init__(self, app_spec, destination=None,
...
+                with open(templ_path) as f:
+                    s = f.read()
+                t = Template(s)
+                with open(file_path, 'w') as f:
+                    f.write(t.safe_substitute(**context))
+                os.remove(templ_path)
+
+        distutils.dir_util.copy_tree(tmpdir, destination, True)
+        distutils.dir_util.remove_tree(tmpdir)
+        print('\nInitialized atomic app: %s at %s' % (app_name, destination))
```
dustymabe: Might be better to do this in `cli/main.py` since that is where run/stop/fetch print out information to the user about what happened during execution.

rtnpro: Agreed!

atomicapp/nulecule/main.py
```diff
@@ -120,6 +123,56 @@ def __init__(self, app_spec, destination=None,
...
+                                    '../external/templates/nulecule')
+        if destination is None:
+            destination = os.path.join('.', app_name)
+
+        if os.path.exists(destination) and os.path.isdir(destination) and os.listdir(destination):
+            value = raw_input('Destination directory is not empty! Do you still want to proceed? [Y]/n: ')
+            value = value or 'y'
+            if value.lower() != 'y':
+                print(
+                    '\nAtomic app: %s was not initialized because destination '
```
cdrage: Should be `Atomic App`

atomicapp/nulecule/main.py
cdrage: Should be `Atomic App`

atomicapp/nulecule/main.py
rtnpro: @dustymabe Although I am tempted to break the function into units, but I don't see much value in it as I don't see much reusability for the units.

dustymabe: That is fine. please add comments then.

rtnpro: ok

atomicapp/nulecule/main.py
rtnpro: :+1:

atomicapp/cli/main.py
cdrage: okay, removed parents=[global_parser] and got this:
```
dropbox/dev/atomicapp  pr_603 ✗                                                                                        22m ⚑  
▶ sudo atomicapp init --help
usage: atomicapp init [-h] [--destination DESTINATION] app_name

positional arguments:
  app_name              App name.

optional arguments:
  -h, --help            show this help message and exit
  --destination DESTINATION
                        Path to the directory where the Atomic App is to be
                        initialized.
```

which is much better :+1:

cdrage: although now getting error:
```
▶ atomicapp init hello
  File "/usr/local/bin/atomicapp", line 9, in <module>
    load_entry_point('atomicapp==0.4.5', 'console_scripts', 'atomicapp')()
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.4.5-py2.7.egg/atomicapp/cli/main.py", line 501, in main
    cli.run()
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.4.5-py2.7.egg/atomicapp/cli/main.py", line 440, in run
    if args.mode:
AttributeError: 'Namespace' object has no attribute 'mode'

```

rtnpro: Otherwise, https://github.com/rtnpro/atomicapp/blob/atomicapp_init/atomicapp/cli/main.py#L440 breaks.
```
  File "/home/rtnpro/.virtualenvs/atomicapp/bin/atomicapp", line 9, in <module>
    load_entry_point('atomicapp==0.4.5', 'console_scripts', 'atomicapp')()
  File "/home/rtnpro/workspace/projectatomic/atomicapp/atomicapp/cli/main.py", line 501, in main
    cli.run()
  File "/home/rtnpro/workspace/projectatomic/atomicapp/atomicapp/cli/main.py", line 440, in run
    if args.mode:
AttributeError: 'Namespace' object has no attribute 'mode'
```

That's why I had suggested to clean up the CLI in another PR.

dustymabe: yeah leave it like it is for now. We have a lot to do around handling of both config and CLI :(.

cdrage: @dustymabe :( okay, new PR in the future then!

cdrage:
LGTM :+1: Awesome work!

dustymabe:
#dotests

dustymabe:
@rtnpro I have a few comments but I say we merge this and I'll open up a PR and let you review it later. @cdrage let's merge.

cdrage:
Merged! Awesome work @rtnpro

PR #610, opened 2016-03-09
Change from ReadWrite to ReadWriteOnce in persistent storage tests

cdrage:
In 1.0.0 of Kubernetes it is no longer ReadWrite but now ReadWriteOnce.

http://kubernetes.io/v1.1/docs/user-guide/persistent-volumes.html

This modifies the tests to ReadWriteOnce instead of ReadWrite.


kadel:
LGTM

Issue #617, opened 2016-03-10
Support multi-line run command in docker artifact file


surajssd:
In docker artifact if I provide the docker run command as a multi-line then Atomic App fails.

```bash
(atomic) [vagrant@centos7-adb etherpad-centos7-atomicapp]$ cat artifacts/docker/etherpad-centos7-atomicapp-run 
docker run --link=mariadb-atomicapp-app:db \
           -d \
           --name etherpad \
           -p $hostport:9001 \
           $image
(atomic) [vagrant@centos7-adb etherpad-centos7-atomicapp]$ 
(atomic) [vagrant@centos7-adb etherpad-centos7-atomicapp]$ sudo atomicapp run . -v --provider=docker
2016-03-10 12:55:59,785 - [INFO] - cli/main.py - Action/Mode Selected is: run
2016-03-10 12:55:59,785 - [DEBUG] - cli/main.py - Final parsed cmdline: run . -v --provider=docker
2016-03-10 12:55:59,786 - [DEBUG] - nulecule/main.py - NuleculeManager init app_path: .
2016-03-10 12:55:59,786 - [DEBUG] - nulecule/main.py - NuleculeManager init image: None
2016-03-10 12:55:59,786 - [DEBUG] - nulecule/main.py - Request to unpack to None to .
2016-03-10 12:55:59,792 - [DEBUG] - atomicapp/plugin.py - Loading providers from /home/vagrant/atomicapp/atomicapp/providers
2016-03-10 12:55:59,794 - [DEBUG] - atomicapp/plugin.py - Loading providers from /home/vagrant/atomicapp/atomicapp/providers
2016-03-10 12:55:59,794 - [INFO] - nulecule/base.py - Pulling external application for mariadb-centos7-atomicapp.
2016-03-10 12:55:59,794 - [INFO] - nulecule/base.py - Unpacking image: projectatomic/mariadb-centos7-atomicapp to ./external/mariadb-centos7-atomicapp
2016-03-10 12:55:59,884 - [INFO] - nulecule/container.py - Pulling Docker image: projectatomic/mariadb-centos7-atomicapp
2016-03-10 12:55:59,884 - [DEBUG] - nulecule/container.py - /usr/bin/docker pull projectatomic/mariadb-centos7-atomicapp
Using default tag: latest
Trying to pull repository docker.io/projectatomic/mariadb-centos7-atomicapp ... latest: Pulling from projectatomic/mariadb-centos7-atomicapp

[...]

PR #624, opened 2016-03-14
Docker-run file with multi-line command supported

surajssd:
If user has provided a docker-run artifact with multi-line command then it can be parsed now, by removing the backslashes(`\`) and `\n`.

Fixes issue #617

### DIFF
--- a/atomicapp/providers/docker.py
+++ b/atomicapp/providers/docker.py
@@ -90,6 +90,8 @@ class DockerProvider(Provider):
             label_run = None
             with open(artifact_path, "r") as fp:
                 label_run = fp.read().strip()
+                # if docker-run provided as multiline command
+                label_run = ' '.join(label_run.split('\\\n'))
             run_args = label_run.split()
 
             # If --name is provided, do not re-name due to potential linking of containers. Warn user instead.
--- a/tests/units/providers/docker_artifact_test/run-with-backslashes
+++ b/tests/units/providers/docker_artifact_test/run-with-backslashes
@@ -1,0 +1,6 @@
+docker run \
+-d \
+-p \
+80:80 \
+--name centos7 \
+centos7
--- a/tests/units/providers/test_docker_provider.py
+++ b/tests/units/providers/test_docker_provider.py
@@ -78,3 +78,15 @@ class TestDockerProviderBase(unittest.TestCase):
         provider.artifacts = [self.artifact_dir + 'hello-world-one']
         with pytest.raises(ProviderFailedException):
             provider.run()
+
+    def test_docker_run_with_backslashes(self):
+        data = {'namespace': 'test', 'provider': 'docker'}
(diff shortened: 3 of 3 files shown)


surajssd:
@centos-ci putting a single comment is enough! @dustymabe Where can I file a bug for @centos-ci ?

cdrage:
#dotests

atomicapp/providers/docker.py
```diff
@@ -90,6 +90,8 @@ def run(self):
             label_run = None
             with open(artifact_path, "r") as fp:
                 label_run = fp.read().strip()
+                # if docker-run provided as multiline command
+                label_run = ' '.join(label_run.split('\\\n'))
```
cdrage: This looks fine to me. 

Although couldn't we change the top one to: ".rstrip()" instead of "strip()" (since that'll remove newlines) and then remove only backslashes instead?

surajssd: @cdrage there are no multiple lines its a single line, backslashes make it single line

dustymabe:
> @centos-ci putting a single comment is enough! @dustymabe Where can I file a bug for @centos-ci ?

Hey @surajssd. It is because there are three different tests (one for each provider) that is configured to kick off. The comment isn't ideal so I'll try to find an answer to it in the future (might be next week)

dustymabe:
If we go with something like this can we add tests for it?

surajssd:
@dustymabe yep will send tests soon!

atomicapp/providers/docker.py
cdrage: What I mean is that wouldn't it be more pythonesque to use:
```py
label_run = fp.read().rstrip()
                # if docker-run provided as multiline command
                 label_run = ' '.join(label_run.split('\\'))
```

surajssd: @cdrage 

```bash
[hummer@hummer tmp] $ cat run 
docker run \
-d \
-p \
80:80 \
centos7
[hummer@hummer tmp] $ ipython
In [4]: f = open('run')

In [5]: text = f.read().rstrip()

In [6]: text
Out[6]: 'docker run \\\n-d \\\n-p \\\n80:80 \\\ncentos7'

In [7]: ' '.join(text.split('\\'))
Out[7]: 'docker run  \n-d  \n-p  \n80:80  \ncentos7'
```

cdrage: Ahhhh, that's the reason why you have it like that. I thought rstrip would actually remove \n as well, but nope.

This code LGTM :+1: thank you for explaining!

surajssd: :) :+1:

surajssd:
@dustymabe @cdrage added tests!

cdrage:
#dotests

cdrage:
Tests pass, LGTM :+1:

PR #633, opened 2016-03-15
Update docs / more clarity on cli comparison

cdrage:
Create more spacing / update markdown clarity on CLI comparisons


kadel:
What if we put that comparison to table?
It might get even clearer.

something like this:

|            | atomicapp | atomic |
|----------|-----------|--------|
| fetch    |  `atomicapp fetch ...` | `atomic run ...` |
| run      |
| stop     | 
|genanswers|

cdrage:
@kadel woah, much better, I'm an idiot >.>

I'll update the PR with that change.

cdrage:
@kadel updated based on your comments, thanks! :)

kadel:
:+1:  LGTM :wink:

Issue #588, opened 2016-03-03
genanswers should drop in target directory, not current directory


jberkus:
If I do:

    atomicapp genanswers atomic-site/Nulecule

It will drop an answers.conf.sample in my *current* directory, not in the target directory of atomic-site/Nulecule.  Which is both messy, and a problem because 'atomicapp run' will look for answers.conf in the target directory when it runs.

dustymabe:
Well it depends on your take on this. Most of the time `genanswers` is used is when you want to grab answers from a remote docker image, modify them, and then run the app like so:

```
atomicapp genanswers projectatomic/helloapache
vi answers.conf
atomicapp run projectatomic/helloapache -a answers.conf
```

In this case what you are saying is `atomicapp genanswers atomic-site/Nulecule`  where `atomic-site/Nulecule` is a directory that already exists?

With `genanswers`, up until this point there has been no concept of a target directory.

jberkus:
What I'm talking about is sensible defaults. That is, the default for one command should be the same as the default for another command.  If `genanswers` drops answers.conf in the current directory, then `run` should look in the current directory first, without requiring a parameter from the user.

Having the default be one thing for `genanswers` and something different for `run` is very unfriendly, and makes it harder to learn the tool.

dustymabe:
I'll be honest I've never used genanswers in a context where I wasn't grabbing it from a remote image. So there are two cases:

1. generating from nulecule that is embedded in docker image
2. generating from local nulecule

For 1. is it ok to have it populated in the local directory. I don't really see another option without forcing the user to tell us where to put it.

For 2. we can detect this and put it in the directory given (my thoughts are that most of the time the user would be in the directory and just pass ./).

dustymabe:
The only problem with the above approach is that it's inconsistent based on where the nulecule currently resides, which I don't like.

jberkus:
well, per above, my concern is that `run` and `genanswers` should have the same behavior.  So if it's going to make sense in a lot of cases to drop it in the current directory, then `run` should also look in the current directory for answers by default.

dustymabe:
I definitely don't think having `run` look in the local dir is a solution so let's search elsewhere.

`genanswers` is the only case that exists where there won't be a context directory (i.e. a local directory where a nulecule is extracted to) that results from an action. In the case of `fetch`, which is almost the same as `genanswers` but actually does leave a directory on the local system, the sample answers file gets put in the context directory.

[...]

PR #638, opened 2016-03-17
Adds -a option to genanswers to designate where to put answers.conf

cdrage:
Fixues issue:
https://github.com/projectatomic/atomicapp/issues/588

Adds the feature to specify the destination of your answers.conf file
generation.


cdrage:
#dotests

cdrage:
openshift tests = false positive? the timeout issue occured

cdrage:
Tests pass locally :+1:

atomicapp/cli/main.py
```diff
@@ -53,9 +53,10 @@ def print_app_location(app_path):
 def cli_genanswers(args):
     try:
         argdict = args.__dict__
+        location = argdict['answers']
```
dustymabe: I don't know if I like this. The reason you are converting from `answers` to `location` is because "answers" doesn't really make sense as a variable name for this I don't think.

How about we use `--destination` instead and the user could provide a path (`./path/to/myanswers.conf`) or a directory (`./path/to/`) in which the program will tack on the default filename to the directory. 

@jberkus, do you think `--answers` makes sense, or should we change it?

cdrage: Yeah, I was thinking that too, using `--destination` to coincide with how we do `fetch`. :)

jberkus:
The reason to use -a is symmetry:

    genanswers -a demo/answers.conf Nulecule/

    atomicapp run -a demo/answers.conf Nulecule/

I understand the arguments for other parameters, but I don't find them persuasive.  Anything other than -a is going to require me to look it up.

dustymabe:
> I understand the arguments for other parameters, but I don't find them persuasive. Anything other than -a is going to require me to look it up.

This is partly why I decided to have the only behavior to be to put the file in your cwd. If it has one behavior then it is certainly easy to remember. :-P

atomicapp/cli/main.py
dustymabe: If we are to leave it as `-a` then I'd rather not do this here but rather leave it as `answers` and comment in the function in `main.py` what it is being used for.

jberkus:
Except that using the CWD is asymmetrical. That is, "run" requires -a if you want to use an answers.conf, it won't use CWD.

cdrage:
@dustymabe Should we change this to --destination or stick to -a ? What's your opinion?

jberkus:
@cdrage here's the instructions I want to be able to give:

1. generate answers with `atomicapp genanswers -a answers.conf Nulecule/`
2. edit answers
3. run the app `atomicapp run -a answers.conf Nulecule/'

See the advantage in usability from being symmetric with the use of the `-a` parameter?  The user doesn't have to constantly look up which switch is usable with which command.

jberkus:
Relevant to this, the PostgreSQL project used to have different parameters for different commands when it came to databases and the data directory.  After years of user bug reports, we refactored and made them consistent; all CLI commands use -D for the data directory and -d for the database, or they don't support those parameters.  And the letter 'd' is used for nothing else. 

Let's start out by being consistent instead of having to refactor later.

dustymabe:
@cdrage `-a` is fine.

> generate answers with atomicapp genanswers -a answers.conf Nulecule/

@jberkus just so we are clear, my understanding is that command will create an answers.conf file in your current directory. Is that your understanding?

jberkus:
if a directory is not supplied, yes.  but if you do this:

`atomicapp genanswers -a Nulecule/dev/answers.conf Nulecule/`

... then it would create it in `Nulecule/dev/` relative to the current directory.  Or any other user-supplied directory.  Just to keep things simple, if the directory does not exist, it should error.

cdrage:
#dotests

cdrage:
@jberkus @dustymabe 

Updated this PR to include checking if the directory exists + rebased against mastah.

atomicapp/cli/main.py
```diff
@@ -52,9 +52,10 @@ def print_app_location(app_path):
 
 def cli_genanswers(args):
     argdict = args.__dict__
+    location = argdict['answers']
```
dustymabe: I still don't think we need to do this here. What do you think about just passing it in with the rest of the args and do it inside the function in main.py ?

atomicapp/nulecule/main.py
```diff
@@ -147,10 +147,19 @@ def genanswers(self, dryrun=False, answers_format=None, **kwargs):
...
-        answers_file = os.path.join(os.getcwd(), ANSWERS_FILE)
+        if location is None:
+            answers_file = os.path.join(os.getcwd(), ANSWERS_FILE)
+        else:
+            answers_file = os.path.join(location, ANSWERS_FILE)
+
         if os.path.exists(answers_file):
             raise NuleculeException(
-                "Can't generate answers.conf over existing file")
+                "Can't generate %s over existing file" % answers_file)
```
dustymabe: Might be better to have the file at the end after a colon:

`"Can't generate answers. File exists: %s" % answers_file`

atomicapp/nulecule/main.py
```diff
@@ -147,10 +147,19 @@ def genanswers(self, dryrun=False, answers_format=None, **kwargs):
         self.answers_format = answers_format or ANSWERS_FILE_SAMPLE_FORMAT
 
         # Check to make sure an answers.conf file doesn't exist already
-        answers_file = os.path.join(os.getcwd(), ANSWERS_FILE)
+        if location is None:
+            answers_file = os.path.join(os.getcwd(), ANSWERS_FILE)
+        else:
+            answers_file = os.path.join(location, ANSWERS_FILE)
```
dustymabe: Double sigh.. problems like this make me want to retire. Ok we need to handle this for both the "in container" and "native" cases. That means if we are inside a container and the path is absolute then we need to translate the path like is done here: https://github.com/projectatomic/atomicapp/blob/master/atomicapp/nulecule/main.py#L72

dustymabe:
Ok a few comments in the code. We also need to decide exactly "what" happens when. Let's create a few cases:

- User provides path to existing directory
    - aa genanswers -a /path/to/dir
    - => creates /path/to/dir/answers.conf
- User provides path to a file (not existing) in an existing directory
    - aa genanswers -a /path/to/dir/somefile
    - => creates /path/to/dir/somefile
- User provides path to a directory that does not exist
    - aa genanswers -a /path/to/no/dir
    - => error - directory does not exist

The hard part about this is number 2. How do we know if the user provided the path to a file that they wanted us to create or they provided us a path to a directory that doesn't exist? 

I think we should just make them give us a full file name when providing -a. That way it is simple:

if dirname(file) doesn't exist => error
if file exists => error
else create file

atomicapp/cli/main.py
cdrage: I was following similar to how we do:
```
destination = argdict['destination'] 
```
Should we not follow this convention / convert to do this later in main.py?

dustymabe: I see what you were trying to do now. this seems reasonable. +1

cdrage:
Will continue on this once #658 is merged so we can use the "are we in a container" util

dustymabe:
> Will continue on this once #658 is merged so we can use the "are we in a container" util

No need to wait as we already have the Utils.inContainer() function.

atomicapp/nulecule/main.py
cdrage: I'm a bit confused in regards to self.app_path.

So if inContainer() comes back as true, we'd use the /host/ path, correct?

dustymabe: sigh.. another reason to retire :) 

The link i used was based on master.. which moved since I pasted the link.. lets try using a link where the content won't move:
https://github.com/projectatomic/atomicapp/blob/fdc8a0bfb691d9117c5d1bd78549a37a84e39b74/atomicapp/nulecule/main.py#L76

basically if the path is absolute then we want to convert to the equivalent path inside the container. we are using the `getRoot()` function for this in other places.

cdrage:
Update: I'm going to pick back up on this soon. ^^

PR #639, opened 2016-03-17
Remove legal information from each file

cdrage:
ping @jberkus @dustymabe

Do we really need ALL this legal information in each and every file (some files we don't even had it in)?

We already list the license as LGPL3 within setup.py, isn't that enough?


jberkus:
I don't believe so, but I'm not an attorney.

I can see if we can get an answer from our legal help.

goern:
"Put all the copyright notices together, right near the top of each file." at http://www.gnu.org/licenses/gpl-howto.en.html

cdrage:
@goern maybe instead I should make a PR to add more legal information? :D there's a bunch of files missing the copyright notices at the top.

cdrage:
Closing this for now, I believe @Preeticp is working on this.

PR #643, opened 2016-03-18
Move kubeconfig parsing to a separate file

cdrage:
First few PR's in the conversion of cmd to API for Kubernetes.

This moves the kubeconfig parsing to a separate file so both providers (and perhaps future ones too) can use the same file.


cdrage:
ping @kadel 

I'm pretty sure that:
```py
op = self.get_oc_provider()
```

doesn't do anything to the kubeconfig parsing :)

cdrage:
#dotests

dustymabe:
@cdrage yay! - will look at this more today and do some testing and comment.

kadel:
> ping @kadel
> 
> I'm pretty sure that:
> 
> op = self.get_oc_provider()
> 
> doesn't do anything to the kubeconfig parsing :)

Yes, you are right :wink: 
It definitely makes sense to get it out of the openshift provider to somewhere else.

atomicapp/providers/openshift.py
```diff
@@ -749,7 +648,7 @@ def _set_config_values(self):
 
         # get values from providerconfig
         if self.config_file:
-            providerconfig = self._parse_kubeconf(self.config_file)
+            providerconfig = KubeConfig._parse_kubeconf(self.config_file)
```
dustymabe: broken -- no leading underscore needed: `[ERROR] - main.py - type object 'KubeConfig' has no attribute '_parse_kubeconf'`

atomicapp/kubeconfig.py
```diff
@@ -0,0 +1,117 @@
...
+        logger.debug("Parsing %s", filename)
+
+        with open(filename, 'r') as fp:
+            kubecfg = anymarkup.parse(fp.read())
+
+        try:
+            return KubeConfig()._parse_kubeconf_data(kubecfg)
+        except ProviderFailedException:
+            raise ProviderFailedException('Invalid %s' % filename)
+
```
dustymabe: can we make this a @staticmethod too?

dustymabe:
See attached patch for a few small suggestions. [patch.txt](https://github.com/projectatomic/atomicapp/files/183242/patch.txt)

dustymabe:
@cdrage - also why shouldn't `kubeconfig.py` still be in the providers folder? Maybe we should think about the hierarchy of what we want this thing to look like.

atomicapp/providers/openshift.py
cdrage: crap you're right, that underscore shouldn't be there.

atomicapp/kubeconfig.py
cdrage: yup, will do

cdrage:
@dustymabe the reason being was the old documentation for atomicapp where you could add your own provider by throwing it into `/providers`, but we're not doing it that way, so having it in kubeconfig should be fine :)

cdrage:
Updated!

cdrage:
#dotests

cdrage:
Kubernetes failed due to the centos-ci bug with default/default, otherwise, tests pass

dustymabe:
#dotestkubernetes

dustymabe:
@cdrage rebase please.

cdrage:
@dustymabe done!

atomicapp/providers/kubeconfig.py
```diff
@@ -0,0 +1,118 @@
...
+        with open(filename, 'r') as fp:
+            kubecfg = anymarkup.parse(fp.read())
+
+        try:
+            return KubeConfig.parse_kubeconf_data(kubecfg)
+        except ProviderFailedException:
+            raise ProviderFailedException('Invalid %s' % filename)
+
+    @staticmethod
+    def parse_kubeconf_data(kubecfg):
```
dustymabe: Why did we remove the _ ? It's still a method that should only be called from this file right?

dustymabe:
> @dustymabe the reason being was the old documentation for atomicapp where you could add your own provider by throwing it into /providers, but we're not doing it that way, so having it in kubeconfig should be fine :)

Hey @cdrage. You have a point. Why don't we create a subdirectory under providers and put kubeconfig (and possibly future common items) in there: `atomicapp/providers/lib/kubeconfig.py` ?

atomicapp/providers/kubeconfig.py
cdrage: sicne you wanted it to be a static method instead :) makes sense in case someone wants to pass in an object rather than a yaml of the config

cdrage:
@dustymabe Yeah, that sounds like a better idea. i'll move it to lib

cdrage:
@dustymabe updated based on your comments, rebased as well.

cdrage:
#dotests

dustymabe:
LGTM.. merging

Issue #501, opened 2016-01-12
Check status of target environment before deploy


dustymabe:
We should check the status of the target environment before we start deploying artifacts. If we have 4 artifacts to deploy:

etherpad rc
etherpad svc
mariadb rc
mariadb svc

Then we should check to see what exists in the target environment before we start to deploy. If any of them exist before then we shouldn't attempt to deploy any of them.

dustymabe:
Anyone who picks up this issue should not worry about doing it on the kubernetes provider until we migrate kubernetes to use the API. (i.e. do kubernetes last)

PR #653, opened 2016-03-28
[WIP] k8s cmd to api

cdrage:
This works (and mostly well), although we're blocked a few things:

- [ ] Get https://pypi.python.org/pypi/pykube/0.9.0 on CentOS (since we're using that to build)

Other stuff to do:
- [ ] Get https://pypi.python.org/pypi/pykube/0.9.0 on other distros too preferably
- [ ] Update documentation, since now you can pass --provider-auth to k8s provider
- [ ] Make sure all config differences work correctly (having no user, having no auth token in .kube/config, etc.) and error / warn.

To test this PR:
`pip install pykube`

### DIFF
--- a/atomicapp/providers/kubernetes.py
+++ b/atomicapp/providers/kubernetes.py
@@ -20,16 +20,58 @@
 import anymarkup
 import logging
 import os
-from string import Template
 
-from atomicapp.constants import (LOGGER_COCKPIT,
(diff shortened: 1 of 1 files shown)


dustymabe:
Hey @cdrage, while you do this work if you keep in mind #501 and #421 - @rtnpro made a lot of progress on #421 in #456 so you might take a look at some of the work he did.

cdrage:
Closing in favour of #712

Issue #656, opened 2016-03-29
Add default config value ~/.kube/config for both OpenShift and Kubernetes.


cdrage:
ping @kadel @dustymabe @rtnpro 

So currently we use `kubectl` got Kubernetes. This automatically picks up `.kube/config` for usage.

We do NOT do this with OpenShift as we have to manually specify --providerconfig or  token + ca.

Thus you can easily run `atomicapp run --provider=kubernetes foobar`. But with OpenShift you have to provider `atomicapp run --provider=openshift foobar --providerconfig ~/.kube/config`.

With the conversion of the k8s provider from cmd to api, we have to require a config file location.

Allso, after looking at the OpenShift code we should provide a **DEFAULT** value of `~/.kube/config` of the current user for both the OpenShift and k8s providers. So we no longer have to provider `--providerconfig ~/.kube/config` each time.

Or else we'll break a lot of current blog posts and examples. Also the fact that `~/.kube/config` is conventionally where the user stores their config.

cdrage:
Added the label blocker due to the nature of all these examples / blog posts we currently have.

I don't like the feeling of having to add `--providerconfig ~/.kube/config` to each `atomicapp run projectatomic/helloapache`.

dustymabe:
@cdrage - I agree that picking up ~/.kube/config is what we should do - see my comment https://github.com/projectatomic/atomicapp/pull/420#issuecomment-163036234

One problem I think is that `~` doesn't work inside the container so what we need to do is actually find out what the user's home directory is and then use the full path right?

dustymabe:
@kadel your input on this would would be appreciated.

kadel:
I agree with defaulting to `~/kube/config`, but as @dustymabe  mentioned finding users home directory might be tricky.

I don't know if there is a easy way how to find out what who started container :-( 
There is `$SUDO_UID` in atomic cli (https://github.com/projectatomic/atomic/blob/master/docs/atomic-run.1.md) but I'm not sure if it is going to be helpful, or how to do this without atomic cli.

cdrage:
@dustymabe @kadel 

We can use the env variables SUDO_USER and USER within atomicapp.

For ex.

If root is ran these are the env variables:
USER=root
SUDO_USER=foobar

If user ran without root:
USER=foobar
SUDO_USER=None

Should be easy to parse :)

dustymabe:
[...]

PR #658, opened 2016-03-31
Util to gather what user is running Atomic App and which home dir it should use

cdrage:
This call finds out what user is running the Atomic App.

Whether they are:
  sudo
  root
  user

If it is running under sudo, this util will find the user running the
sudo command.

Once the user is found, os.path.expanduser(user) is utilized to find the
home directory. If a home directory is not found atomic app will error
out.

Fixes https://github.com/projectatomic/atomicapp/issues/656

This util will be used to implement default config locations for OpenShift and k8s providers.

### DIFF
--- a/atomicapp/utils.py
+++ b/atomicapp/utils.py
@@ -342,7 +342,7 @@ class Utils(object):
         Returns:
             (bool): True == we are in a container
         """
-        if os.path.isdir(HOST_DIR):
+        if os.path.isfile('/.dockerenv') and os.path.isfile('/.dockerinit') and os.path.isdir(HOST_DIR):
             return True
         else:
             return False
--- a/atomicapp/utils.py
+++ b/atomicapp/utils.py
@@ -335,7 +335,9 @@ class Utils(object):
     @staticmethod
     def inContainer():
         """
-        Determine if we are running inside a container or not.
+        Determine if we are running inside a container or not. This is done by
+        checking to see if /host has been passed as well as if .dockerenv and
+        .dockerinit files exist
 
         Returns:
             (bool): True == we are in a container
@@ -376,6 +378,58 @@ class Utils(object):
     def rm_dir(directory):
         logger.debug('Recursively removing directory: %s' % directory)
         distutils.dir_util.remove_tree(directory)
+
+    @staticmethod
+    def getUserName():
+        """
+        Finds the username of the user running the application. Uses the
+        SUDO_USER and USER environment variables. If runnning within a
+        container, SUDO_USER and USER varibles must be passed for proper
+        detection.
+        Ex. docker run -v /:/host -e SUDO_USER -e USER foobar
+        """
+        sudo_user = os.environ.get('SUDO_USER')
+
+        if os.getegid() == 0 and sudo_user is None:
+            user = 'root'
+        elif sudo_user is not None:
+            user = sudo_user
+        else:
+            user = os.environ.get('USER')
+        return user
+
+    @staticmethod
+    def getUserHome():
+        """
+        Finds the home directory of the user running the application.
+        If runnning within a container, the root dir must be passed as
+        a volume.
+        Ex. docker run -v /:/host -e SUDO_USER -e USER foobar
+        """
+        logger.debug("Finding the users home directory")
+        user = Utils.getUserName()
+        incontainer = Utils.inContainer()
+
+        # Check to see if we are running in a container. If we are we
+        # will chroot into the /host path before calling os.path.expanduser
+        if incontainer:
+            os.chroot(HOST_DIR)
+
+        # Call os.path.expanduser to determine the user's home dir.
+        # See https://docs.python.org/2/library/os.path.html#os.path.expanduser
+        # Warn if none is detected, don't error as not having a home
+        # dir doesn't mean we fail.
+        home = os.path.expanduser("~%s" % user)
+        if home == ("~%s" % user):
+            logger.error("No home directory exists for user %s" % user)
+
+        # Back out of chroot if necessary
+        if incontainer:
+            os.chroot("../..")
+
+        logger.debug("Running as user %s. Using home directory %s for configuration data"
+                     % (user, home))
+        return home
 
     @staticmethod
     def make_rest_request(method, url, verify=True, data=None):
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,8 +7,8 @@ ENV ATOMICAPPVERSION="0.4.5"
 LABEL io.projectatomic.nulecule.atomicappversion=${ATOMICAPPVERSION} \
       io.openshift.generate.job=true \
       io.openshift.generate.token.as=env:TOKEN_ENV_VAR \
-      RUN="docker run -it --rm \${OPT1} --privileged -v \${PWD}:/atomicapp -v /run:/run -v /:/host --net=host --name \${NAME} -e NAME=\${NAME} -e IMAGE=\${IMAGE} \${IMAGE} \${OPT2} run \${OPT3}" \
-      STOP="docker run -it --rm \${OPT1} --privileged -v \${PWD}:/atomicapp -v /run:/run -v /:/host --net=host --name \${NAME} -e NAME=\${NAME} -e IMAGE=\${IMAGE} \${IMAGE} \${OPT2} stop \${OPT3}"
+      RUN="docker run -it --rm \${OPT1} --privileged -v \${PWD}:/atomicapp -v /run:/run -v /:/host --net=host --name \${NAME} -e USER -e SUDO_USER -e NAME=\${NAME} -e IMAGE=\${IMAGE} \${IMAGE} \${OPT2} run \${OPT3}" \
+      STOP="docker run -it --rm \${OPT1} --privileged -v \${PWD}:/atomicapp -v /run:/run -v /:/host --net=host --name \${NAME} -e USER -e SUDO_USER -e NAME=\${NAME} -e IMAGE=\${IMAGE} \${IMAGE} \${OPT2} stop \${OPT3}"
 
 WORKDIR /opt/atomicapp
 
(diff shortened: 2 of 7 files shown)


cdrage:
@dustymabe @kadel :)

dustymabe:
@cdrage - this doesn't work when running inside a container.  I modified it to print out the results of a call to this function and here is what you see when running in a container:

```
[vagrant@f23 wordpress-centos7-atomicapp]$ sudo docker run -it --rm  --privileged -v $(pwd):/atomicapp -v /run:/run -v /:/host --net=host --name atomicapp projectatomic/atomicapp:local run ./      
[INFO] - main.py - Action/Mode Selected is: run
XXX /root
```

here is what you get when not running inside a container:

```
[vagrant@f23 wordpress-centos7-atomicapp]$ sudo PYTHONPATH=/sharedfolder/atomicapp /usr/bin/python -B /sharedfolder/atomicapp/atomicapp/cli/main.py run ./                                           
[INFO] - main.py - Action/Mode Selected is: run
XXX /home/vagrant
```

cdrage:
@dustymabe 
One more thing to add then! Now it's determining whether we are in a container or not.

cdrage:
http://stackoverflow.com/questions/20010199/determining-if-a-process-runs-inside-lxc-docker seems like the best way, i'll modify this a bit.

@dustymabe 
So what should it output when it's in a container? How do you think we should be able to detect the host dir of the user?

cdrage:
@dustymabe 

Okay, so essentially it's two things that we push for AtomicApp. Running it via Atomic CLI or AtomicApp CLI.

Thus the only change we need to add is -e SUDO and -e SUDO_USER to the RUN label within our Dockerfile.

cdrage:
@dustymabe 

Updated this in order to detect if we are running within a container. Thus when we use atomic CLI we can still detect what user is running it as from a usage perspective the only two options we give is to run via `atomicapp` or `atomic`.

Tested this with `atomic` CLI. By building this container (tagging it as `atomicapp`) then building the helloapache example with `FROM atomicapp` and running it via the Atomic CLI with the Utils.getHomeDir() somewhere in the code in order to see the output. Everything works correctly, woo!


__Remember:__ We'll only use this function when need be, not putting this randomly in the `base.py` code. Thus when we run Atomic App as a container within OpenShift (that doesn't need to find the configdata) it won't break anything. 

This function is intended to:
 - Fix the permission issues when using `sudo fetch, run, etc.` so it doesn't extract as a root user
 - Be able to find configdata by default for k8s (when the api version is implemented) and openshift / marathon providers.

atomicapp/utils.py
```diff
@@ -378,6 +378,60 @@ def rm_dir(directory):
...
+
+        logger.debug("Running as user %s. Using home directory %s for configuration data"
+                     % (user, home))
+        return home
+
+    # This ONLY checks to see if we are running in a Docker container
+    # Atomic App does not support any other container platform (yet).
+    # Within a Docker container, the .dockerenv and .dockerinit files ALWAYS exist.
+    @staticmethod
+    def areWeContainerized():
```
dustymabe: what about the inContainer() function that we have in this same file?

atomicapp/utils.py
```diff
@@ -378,6 +378,60 @@ def rm_dir(directory):
...
+        # os.path.expanduser due to the passed volume
+        if Utils.areWeContainerized():
+            logger.debug("Atomic App is running in a container")
+            if not os.path.isdir("/host"):
+                raise AtomicAppUtilsException(
+                    "/ host directory must be passed to the Atomic App container.")
+            if user is "root":
+                home = "/host/root"
+            else:
+                home = "/host/home/%s" % user
```
dustymabe: This is statically setting the home directory based on the username. This makes an assumption that is unsafe.

cdrage: @dustymabe So parse `/host/etc/passwd` instead?

atomicapp/utils.py
cdrage: Oh yeah, I see that now! Adding .dockerenv and .docketinit to it would be viable too.

atomicapp/utils.py
dustymabe: I'm investigation a possible alternative.. stand by

dustymabe: Ok here is an alternative:

```
diff --git a/atomicapp/utils.py b/atomicapp/utils.py
index 63c3428..f31a740 100644
--- a/atomicapp/utils.py
+++ b/atomicapp/utils.py
@@ -404,10 +404,11 @@ class Utils(object):
             if not os.path.isdir("/host"):
                 raise AtomicAppUtilsException(
                     "/ host directory must be passed to the Atomic App container.")
-            if user is "root":
-                home = "/host/root"
-            else:
-                home = "/host/home/%s" % user
+            os.chroot("/host")
+            home = os.path.expanduser("~%s" % user)
+            if home == ("~%s" % user):
+                logger.error("No home directory exists for user %s" % user)
+            os.chroot("../..")
 
         # See https://docs.python.org/2/library/os.path.html#os.path.expanduser
         # on how it finds the users home directory and why "~foobar" is valid for
```

dustymabe:
Also it would be cool to have the part that figures out what "user" we are as part of a separate function that is called from getUserHome(). I can see that being used in other places.

cdrage:
@dustymabe The part of the function that gets sudo + sudo_user env variables?

Otherwise, I've updated with your patch as well as removed the container function I had to inContainer!

atomicapp/utils.py
```diff
@@ -378,6 +380,51 @@ def rm_dir(directory):
...
+            user = sudo_user
+        else:
+            user = os.environ.get('USER')
+
+        # Check to see if we are running in a container. If we are we will use
+        # the /host path for finding the user directory. We are unable to use
+        # os.path.expanduser due to the passed volume
+        if Utils.inContainer():
+            logger.debug("Atomic App is running in a container")
+            if not os.path.isdir("/host"):
```
dustymabe: since inContainer checks this I don't think we need it anymore.

atomicapp/utils.py
```diff
@@ -378,6 +380,51 @@ def rm_dir(directory):
...
+                raise AtomicAppUtilsException(
+                    "/ host directory must be passed to the Atomic App container.")
+            os.chroot("/host")
```
dustymabe: should probably use `HOST_DIR` - sorry for the bad patch

dustymabe:
> @dustymabe The part of the function that gets sudo + sudo_user env variables?

yes

atomicapp/utils.py
```diff
@@ -335,12 +335,14 @@ def getDockerCli(dryrun=False):
...
-        Determine if we are running inside a container or not.
+        Determine if we are running inside a container or not. This is done by
+        checking to see if /host has been passed as well as if .dockerenv and
+        .dockerinit files exist
 
         Returns:
             (bool): True == we are in a container
         """
-        if os.path.isdir(HOST_DIR):
+        if os.path.isfile('/.dockerenv') and os.path.isfile('/.dockerinit') and os.path.isdir(HOST_DIR):
```
dustymabe: are we sure we want to make this change here? if we do this should probably be in a separate commit

dustymabe:
Also we need to update all dockerfiles.. don't want them getting out of sync

atomicapp/utils.py
cdrage: I'll do this a separate commit then )

cdrage:
@dustymabe 
Updated as well as the Dockerfiles. Put the inContainer() edit in a separate commit.

cdrage:
#dotests

dustymabe:
> Also it would be cool to have the part that figures out what "user" we are as part of a separate function that is called from getUserHome(). I can see that being used in other places.

So I refactored the code a bit.. there is a new getUserName() function as well as refactored the getUserHome() function to not repeat code. What do you think? 

```python
    @staticmethod
    def getUserName():
        """
        Finds the username of the user running the application. Uses the 
        SUDO_USER and USER environment variables. If runnning within a 
        container, SUDO_USER and USER varibles must be passed for proper
        detection.
        Ex. docker run -v /:/host -e SUDO_USER -e USER foobar
        """
        sudo_user = os.environ.get('SUDO_USER')

        if os.getegid() == 0 and sudo_user is None:
            user = 'root'
        elif sudo_user is not None:
            user = sudo_user
        else:
            user = os.environ.get('USER')
        return user

    @staticmethod
    def getUserHome():
        """
        Finds the home directory of the user running the application.
        If runnning within a container, the root dir must be passed as
        a volume.
        Ex. docker run -v /:/host -e SUDO_USER -e USER foobar
        """
        logger.debug("Finding the users home directory")
        user = Utils.getUserName()
        incontainer = Utils.inContainer()

        # Check to see if we are running in a container. If we are we
        # will chroot into the /host path before calling os.path.expanduser
        if incontainer: os.chroot(HOST_DIR)

        # Call os.path.expanduser to determine the user's home dir.
        # See https://docs.python.org/2/library/os.path.html#os.path.expanduser
        # Warn if none is detected, don't error as not having a home
        # dir doesn't mean we fail.
        home = os.path.expanduser("~%s" % user)
        if home == ("~%s" % user):
            logger.error("No home directory exists for user %s" % user)

        # Back out of chroot if necessary
        if incontainer: os.chroot("../..")

        logger.debug("Running as user %s. Using home directory %s for configuration data"
                     % (user, home))
        return home
```

cdrage:
@dustymabe 
WOAH! Thanks :dancers: That's wicked.

I've updated the PR with your changes as well as re-based against master.

dustymabe:
#dotests

dustymabe:
That darn pesky kubernetes test needs some love :( 

cdrage most all looks good to me. as part of this do you mind bringing the `Dockerfiles.pkgs/` Dockerfiles up to date. Let's keep the version #, run labels, etc.. synced across all Dockerfiles. Do you agree?

cdrage:
@dustymabe 

Agreed. Let's bring them back up-to-date and let's get this merged. WOOHOO!

cdrage:
@dustymabe 
Updated the Dockerfile.pkgs folder to an up-to-date-version.

dustymabe:
LGTM.

dustymabe:
#dotests

cdrage:
Woo! :confetti_ball:

PR #664, opened 2016-04-01
Fix persistent storage

### LINKED ISSUE #660: persistent storage not working - mutiple entries in nulecule
I tried adding a persistent volume to the wordpress example to test #609. I can't for the life of me get it to even create the persistent volume claim.

Added this to the bottom of the Nulecule:
```
    requirements:
        - persistentVolume:
            name: "foobar"
            accessMode: "ReadWriteMany"
            size: 10
```

And then I added this to do some debug:
```
diff --git a/atomicapp/nulecule/base.py b/atomicapp/nulecule/base.py
index 17c19a5..b7dee4b 100644
--- a/atomicapp/nulecule/base.py
+++ b/atomicapp/nulecule/base.py
@@ -154,11 +154,15 @@ class Nulecule(NuleculeBase):
         # Process preliminary requirements
         # Pass configuration, path of the app, graph, provider as well as dry-run
         # for provider init()
+        print("XXX here")
+        import pprint
+        pprint.pprint(self.graph)
         if REQUIREMENTS_KEY in self.graph[0]:
             logger.debug("Requirements key detected. Running action.")
             r = Requirements(self.config, self.basepath, self.graph[0][REQUIREMENTS_KEY],
                              provider_key, dryrun)
             r.run()
+        print("XXX there")
 
         # Process components
         for component in self.components:
```

The graph that gets to this function in NuleculeBase doesn't even see the requirements:
```
XXX here
[{u'name': u'mariadb-centos7-atomicapp',
  u'source': u'docker://projectatomic/mariadb-centos7-atomicapp'},
 {u'artifacts': {u'docker': [u'file://artifacts/docker/wordpress-run'],
                 u'kubernetes': [u'file://artifacts/kubernetes/wordpress-pod.yaml',
                                 u'file://artifacts/kubernetes/wordpress-service.yaml'],
                 u'openshift': [u'file://artifacts/openshift/volume.yaml',
                                u'file://artifacts/openshift/service.yaml',
                                u'file://artifacts/openshift/pod.yaml']},
  u'name': u'wordpress',
  u'params': [{u'default': u'wordpress',
               u'description': u'wordpress image',
               u'name': u'image'},
              {u'description': u'username for database',
               u'name': u'db_user'},
              {u'description': u'password for db_user', u'name': u'db_pass'},
              {u'description': u'database name', u'name': u'db_name'},
              {u'default': u'mariadb:3306',
               u'description': u'database host',
               u'name': u'db_host'},
              {u'default': 8080,
               u'description': u'port for docker port mapping (used only with docker provider)',
               u'name': u'port'}],
  u'requirements': [{u'persistentVolume': {u'accessMode': u'ReadWriteMany',
                                           u'name': u'foobar',
                                           u'size': 10}}]}]
XXX there
```

This basically shows that there are two items in the array (mariadb first and then wordpress 2nd) so since we are only looking at self.graph[0] there is a bug here.

cdrage:
__Fix detection of no persistentVolumes__

  Kubernetes recently changed that if you use `kubectl get pv` with no pv
  present, the columns should not appear.

  Before 1.0.0:
```
  ▶ kubectl get pvc 
NAME      STATUS    VOLUME    CAPACITY   ACCESSMODES  AGE
```

  After:
```
  ▶ kubectl get pvc
```

  Hence the need to add <= 2 instead of == 2

 __Requirements should retrieve from Nulecule object not graph__

  Requirements previously retrieved from graph:
```
 graph:
     ...
     requirements:
       - persistentVolume:
           name: "var-log-httpd"
           accessMode: "ReadWrite"
           size: 4
```

  This is an undesired affect and should retrieve from it's own object:
```
graph:
     ... 
requirements:
   - persistentVolume:
       name: "var-log-httpd"
       accessMode: "ReadWrite"
       size: 4
```

Fixes #660

### DIFF
--- a/atomicapp/nulecule/base.py
+++ b/atomicapp/nulecule/base.py
@@ -38,7 +38,6 @@ from atomicapp.constants import (APP_ENT_PATH,
                                  NAME_KEY,
                                  INHERIT_KEY,
                                  ARTIFACTS_KEY,
-                                 REQUIREMENTS_KEY,
                                  DEFAULT_PROVIDER)
 from atomicapp.utils import Utils
 from atomicapp.requirements import Requirements
@@ -195,14 +194,11 @@ class Nulecule(NuleculeBase):
         """
         provider_key, provider = self.get_provider(provider_key, dryrun)
 
-        # Process preliminary requirements
-        # Pass configuration, path of the app, graph, provider as well as dry-run
-        # for provider init()
-        if REQUIREMENTS_KEY in self.graph[0]:
-            logger.debug("Requirements key detected. Running action.")
-            r = Requirements(self.config, self.basepath, self.graph[0][REQUIREMENTS_KEY],
-                             provider_key, dryrun)
-            r.run()
+        # Process preliminary requirements before componenets
+        if self.requirements:
+            logger.debug("Requirements detected. Running action.")
+            Requirements(self.config, self.basepath, self.requirements,
+                         provider_key, dryrun).run()
 
         # Process components
         for component in self.components:


dustymabe:
In the commit message for 1410001 we should mention that 'according to the nulecule spec' we should be retrieving from Nulecule object/ not graph. 

Also I think we need to do effect instead of affect, but I always get those wrong so don't shoot me if I'm leading you astray.

cdrage:
@dustymabe updated!

dustymabe:
LGTM - rebase+merge please

cdrage:
Rebased and mergin!

Issue #636, opened 2016-03-16
cli auto-complete for atomicapp


surajssd:
atomicapp is missing auto-complete of commands, this would be nice feature to have.

containscafeine:
+1 for bash completion. @surajssd if you want, we can work on this together.

surajssd:
@containscafeine I feel this can be a **beginner friendly** issue which new people who wanna contribute can work on.

dustymabe:
@surajssd this could be an interesting thing to have, but if we mainly run from an embedded container then we might not ever see the benefits of something like this

shiminsh:
Hi, I'm beginner to atomic project and wanna work on this issue, please guide me further.

cdrage:
@shiminsh How docker-machine does their bash autocompletion could possibly help: https://github.com/docker/machine/blob/master/contrib/completion/bash/docker-machine.bash

abhi1004:
Hi, I was just browsing through the issue and gave it a shot here is the poc https://asciinema.org/a/3ygvn2uxnstvgsn3h02e4kng2 
and the code which I wrote for it https://github.com/abhi1004/atomicapp_cli_completion
Could you please review if it is the step in the right direction.

Regards,
Abhishek

dustymabe:
@abhi1004 this looks pretty awesome! I am fairly new to bash completion development. do you know if it is possible to autogenerate the competion file https://github.com/abhi1004/atomicapp_cli_completion ?

abhi1004:
@dustymabe thanks for the review, I'll try to figure it out if it is possible to autogenerate the completion file.

shiminsh:
@abhi1004 Yeah, mine is also in same direction but you should have talked here before working on any issue, as I'm working on same.

cdrage:
@abhi1004 that'd be awesome :) even if it's a bash script that will auto generate from atomicapp's output that'd be great.

I've been trying to google for a solution but none so far, if it's too hard then there's no worries, even large projects still have to update their autocompletion manually since there's not a solution to autogenerating it.

dustymabe:
> @abhi1004 Yeah, mine is also in same direction but you should have talked here before working on any issue, as I'm working on same.

@shiminsh  I agree. @abhi1004 @shiminsh I would like to thank you both for your interests. Do you mind working on this together? 

Also feel free to try to pick up other issues that look like low hanging fruit and thanks again!

abhi1004:
[...]

PR #665, opened 2016-04-02
Added code for cli auto-complete for atomicapp

shiminsh:


rtnpro:
@shiminsh Thanks for the pull request. The initial work looks good. However, we can do some improvements:
- let's keep the shell script out of the python package, let's follow what ``atomic`` CLI is doing, https://github.com/projectatomic/atomic/blob/master/bash/atomic
- we should 4 spaces for a ``TAB``
- we need to support the commands: ``run``, ``stop``, etc and their options as well

Let us know if you need some help.

shiminsh:
Thanks, I'm looking into it for improvements. I'll ask for help if needed.

shiminsh:
@rtnpro Please explain a little bit third point which other options as I have mention run and stop command in my script.

surajssd:
@shiminsh have you looked into https://github.com/projectatomic/atomicapp/issues/636#issuecomment-206433211 ?

cdrage:
Hi @shiminsh thanks for the contribution. We're going to close this for now until we find an alternative way to generate the auto-completion scripts.

shiminsh:
Okay.

PR #676, opened 2016-04-06
Revert back to pip packaging

cdrage:
Go back to using pip packages with hard-coded versions rather than relying
on waiting upstream for OS Python related packages.

This opens up opportunity for adding third-party packages not yet included on their respective operating system.

(Note: adding `pykube` and a future OpenShift library to the mix)

Also we will use hard-coded versions in order to prevent breakage in the future.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,8 +17,9 @@ ADD requirements.txt ./
 
 # Install needed requirements
 RUN yum install -y epel-release && \
-    yum install -y --setopt=tsflags=nodocs docker && \
-    yum install -y --setopt=tsflags=nodocs $(sed s/^/python-/ requirements.txt) && \
+    yum install -y --setopt=tsflags=nodocs docker python-pip && \
+    pip install -r requirements.txt && \
+    yum remove -y python-pip && \
     yum clean all
 
 WORKDIR /atomicapp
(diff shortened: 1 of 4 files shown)


cdrage:
#dotests

cdrage:
#dotests

dustymabe:
Hey @cdrage. I still don't think this is a good idea to do. The primary reason is because we really shouldn't do a release if that release can't be built downstream. If we do that then we just make it harder on the guys doing the builds because they then have to track down the software and create packages for them, etc.. 

I think it is reasonable to create a new Dockerfile that can be used that pulls from pip - but I don't think we should be pushing anything to Docker Hub (i.e. a real release) unless the packages can be installed as rpms.

cdrage:
@dustymabe 

My problem is the fact that Atomic App is so big (200mb+ now) yet if every provider is through the API it's viable to just use an Alpine Linux container with pip. Would be much smaller and more portable (how we want it to be). 

On the swing side of things, if this was in `Go`, it would just be the binary and using a scratch Dockerfile.

cdrage:
So @dustymabe and @LalatenduMohanty convinced me as to why we should be using rpm's instead of through pip.

Hell, look at what hapenned with left-pad in the npm space. Let's not get to that where if a package goes rouge we can't even build anymore.

PR #677, opened 2016-04-06
Replace inspect with ps, fix #672

### LINKED ISSUE #672: docker-inspect error when no containers running on system
There is a docker inspect error when there are not yet any containers running on the system:

```
[INFO] - docker.py - Deploying to provider: Docker
docker: "inspect" requires a minimum of 1 argument.
See 'docker inspect --help'.

Usage:  docker inspect [OPTIONS] CONTAINER|IMAGE [CONTAINER|IMAGE...]

Return low-level information on a container or image
```

This comes from the following line in docker.py: 
```
docker_cmd = 'docker inspect --format="{{ .Name }}" $(docker ps -aq --no-trunc) | sed "s,/,,g"'
```

We need to run the docker ps command first and then only if there are existing containers running will we call docker-inspect.

containscafeine:


cdrage:
#dotests

dustymabe:
LGTM and tests pass.

cdrage:
Tests pass, dry-run works, tested it locally and it was fine.

Merging.

PR #679, opened 2016-04-06
Added elif to consecutive exclusive if statements

surajssd:
The for loop had conditions and at any given point of time only one can be true not all so added elif to the only if statements in a for loop.


dustymabe:
LGTM

dustymabe:
#dotests

kadel:
LGTM

cdrage:
Tests pass. Mergin'

PR #680, opened 2016-04-06
Added defaultdict to reduce checks on dict and errors

surajssd:
When we know that the dict will have values as list, it is beneficial to use the defaultdict datatype,
we don't have to put an extra code of checks and validations thereby reducing the code and errors.

### DIFF
--- a/atomicapp/providers/openshift.py
+++ b/atomicapp/providers/openshift.py
@@ -25,7 +25,6 @@ import tarfile
 import time
 from urlparse import urljoin
 from urllib import urlencode
-from collections import OrderedDict
 import websocket
 
 from atomicapp.utils import Utils
@@ -40,6 +39,7 @@ from atomicapp.constants import (PROVIDER_AUTH_KEY,
                                  PROVIDER_CA_KEY,
                                  OPENSHIFT_POD_CA_FILE)
 from atomicapp.providers.lib.kubeconfig import KubeConfig
+from atomicapp.applibs.appcollections import OrderedDefaultDict
 from requests.exceptions import SSLError
 import logging
 logger = logging.getLogger(LOGGER_DEFAULT)
@@ -332,7 +332,7 @@ class OpenShiftProvider(Provider):
 
     def init(self):
         # Parsed artifacts. Key is kind of artifacts. Value is list of artifacts.
-        self.openshift_artifacts = OrderedDict()
+        self.openshift_artifacts = OrderedDefaultDict(list)
 
         self._set_config_values()
 
@@ -504,14 +504,10 @@ class OpenShiftProvider(Provider):
             # add all processed object to artifacts dict
             for obj in processed_objects:
                 obj_kind = obj["kind"].lower()
-                if obj_kind not in self.openshift_artifacts.keys():
-                    self.openshift_artifacts[obj_kind] = []
                 self.openshift_artifacts[obj_kind].append(obj)
             return
 
         # add parsed artifact to dict
-        if kind not in self.openshift_artifacts.keys():
-            self.openshift_artifacts[kind] = []
         self.openshift_artifacts[kind].append(data)
 
     def _process_template(self, template):
--- a/atomicapp/applibs/appcollections.py
+++ b/atomicapp/applibs/appcollections.py
@@ -1,0 +1,12 @@
+from collections import defaultdict
+from collections import OrderedDict
+
+
+class OrderedDefaultDict(OrderedDict, defaultdict):
+    """
+    Combines the power of two dictionaries i.e. OrderedDict and defaultdict
+    from collections module, to make a OrderedDefaultDict.
+    """
+    def __init__(self, default_factory=None, *args, **kwargs):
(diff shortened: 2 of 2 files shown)


atomicapp/providers/openshift.py
```diff
@@ -332,10 +333,10 @@ class OpenShiftProvider(Provider):
     provider_ca = None
 
     # Parsed artifacts. Key is kind of artifacts. Value is list of artifacts.
-    openshift_artifacts = {}
+    openshift_artifacts = defaultdict(list)
 
     def init(self):
-        self.openshift_artifacts = {}
+        self.openshift_artifacts = defaultdict(list)
```
dustymabe: Does it make sense why we are setting this in the class variable and the instance variable? @kadel?

atomicapp/providers/openshift.py
```diff
@@ -513,8 +514,6 @@ def _process_artifact_data(self, artifact, data):
             return
 
         # add parsed artifact to dict
-        if kind not in self.openshift_artifacts.keys():
-            self.openshift_artifacts[kind] = []
```
dustymabe: can we git rid of the same check on line 511?

dustymabe:
#dotestsopenshift

dustymabe:
#dotestopenshift

atomicapp/providers/openshift.py
kadel: That looks strange I agree,  but it seems that I had reason for this. https://github.com/projectatomic/atomicapp/commit/c5e6f27c76823343351a3c048f5efde13e86f44d.  But to be honest I don't remember more about this, It is almost 6 months ago

surajssd: https://github.com/projectatomic/atomicapp/commit/c5e6f27c76823343351a3c048f5efde13e86f44d is fine, this is needed, @dustymabe is also asking about https://github.com/projectatomic/atomicapp/pull/680/files#diff-b41fe9caf01c4b4c9724d8a1fd6325baR336

surajssd:
@dustymabe all changes done :)

dustymabe:
#dotests

dustymabe:
@cdrage can you review this and give it a full run through of tests before you merge? perhaps we should wait until after the next release?

surajssd:
OpenShift tests pass!

kadel:
@surajssd  I noticed that you removed `openshift_artifacts = {}` from class. Have you tested this with nested Nulecule?

kadel:
I did couple of manual tests everything works fine.

`defaultdict` is new to me, I've never used it myself,  but it looks nice and clean.

cdrage:
Okay so I've ran it through the tests and everything passes. Code looks good to me.

But let's wait until after Tuesday's release. I'm not too familiar with `defaultdict` as well so I'm hesitant to merge it.

Since Tuesday will be the release that'll (most likely) be in GA, let's wait until after and then merge so we can test against master during that time with this PR.

surajssd:
@kadel I didn't try with the nested Nulecule

dustymabe:
> I did couple of manual tests everything works fine.

@kadel.. thanks so much. @surajssd once the next release comes out we'll have you rebase and then we'll merge.

surajssd:
@dustymabe sure.

dustymabe:
rebase this please

surajssd:
@dustymabe Actually there is a conflict here, we have `OrderedDict` in [code ATM](https://github.com/projectatomic/atomicapp/blob/master/atomicapp/providers/openshift.py#L335), which was addded in https://github.com/projectatomic/atomicapp/commit/1c438d2d6c0416a0cfc6e7b184e409240096e26f

So this PR was generated before https://github.com/projectatomic/atomicapp/commit/1c438d2d6c0416a0cfc6e7b184e409240096e26f and used `defaultdict` and by default `collections` has no mechanism to combine `OrderedDict` and `defaultdict`, but if we want to have an OrderedDefaultDict we can do something like this:

```python
from collections import OrderedDict, defaultdict

class OrderedDefaultDict(OrderedDict, defaultdict):
    def __init__(self, default_factory=None, *args, **kwargs):
        super(OrderedDefaultDict, self).__init__(*args, **kwargs)
        self.default_factory = default_factory
```
Source: http://stackoverflow.com/a/35968897

surajssd:
Rebased and added changes to the PR as per above comment https://github.com/projectatomic/atomicapp/pull/680#issuecomment-212752009

dustymabe:
@surajssd.. considering the conflict between ordered dict and default dict I suggest that we don't implement this PR. I don't think adding the complexity is worth the benefit.

As a side note I wonder if anyone has proposed OrderedDefaultDict upstream.

surajssd:
if we come up with mechanism of ordering artifacts on our own, we can remove the `OrderedDict` and add `defaultdict` and reduce code, as this PR originally was. Until then, close this PR if you feel appropriate.

Considering the upstream, I will also look into it :)

dustymabe:
Closing this for now.

Issue #641, opened 2016-03-18
Atomic App deploys to provider even if provider artifact file not provided


surajssd:
Here is a `Nulecule` file with default provider as `docker` whereas `Nulecule` file has no artifact for docker.
```bash
(atomic) ✔ ~/nulecule-library/skydns-atomicapp [master L|✚ 14…97] 
06:08 $ pwd
/home/vagrant/nulecule-library/skydns-atomicapp
(atomic) ✔ ~/nulecule-library/skydns-atomicapp [master L|✚ 14…97] 
06:08 $ cat Nulecule
---
specversion: 0.0.2
id: skydns-atomicapp
metadata:
  name: SkyDNS
  appversion: 2015-10-21-001
  description: >
    This is a nulecule that will get you the container components you need
    to dev on nulecule apps
params:
    - name: provider
      description: The provider used to deploy application
      default: docker
graph:
  - name: skydns
    params:
      - name: dns_domain
        description: >
          Internal DNS domain name.
          This domain must not be used in your network. Services will be discoverable
          under <service-name>.<namespace>.<domainname>, e.g.
          myservice.default.cluster.local
        default: cluster.local
      - name: dns_server
        description: >
          IP address of the DNS server.
          Kubernetes will create a pod with several containers, serving as the DNS
          server and expose it under this IP address. The IP address must be from
          the range specified as kube_service_addresses above.
          And this is the IP address you should use as address of the DNS server
          in your containers.
        default: 10.254.0.10
    artifacts:
      kubernetes:
        - file://artifacts/kubernetes/skydns-replicationcontroller.yaml
        - file://artifacts/kubernetes/skydns-service.yaml
```
Deploying the app
```bash
(atomic) ✔ ~/nulecule-library/skydns-atomicapp [master L|✚ 14…97] 
06:08 $ sudo atomicapp run . -v
1458295693 - [INFO] - cli/main.py - Action/Mode Selected is: run
1458295693 - [DEBUG] - cli/main.py - Final parsed cmdline: run . -v
1458295693 - [DEBUG] - nulecule/main.py - NuleculeManager init app_path: .
1458295693 - [DEBUG] - nulecule/main.py - NuleculeManager init image: None
1458295693 - [DEBUG] - nulecule/main.py - Request to unpack to None to .
1458295693 - [DEBUG] - atomicapp/plugin.py - Loading providers from /home/vagrant/atomicapp/atomicapp/providers
[...]

PR #682, opened 2016-04-07
Artifacts should be specified for the provider

surajssd:
For a provider mentioned by the user, artifacts should be present in Nulecule file. Earlier if a provider was mentioned and even if that provider's artifacts were not specified in Nulecule file, Atomic App would go ahead. Now it fails error out if artifacts for that provider are not mentioned in Nulecule file.

Fixes issue #641

### DIFF
--- a/atomicapp/nulecule/lib.py
+++ b/atomicapp/nulecule/lib.py
@@ -120,6 +120,17 @@ class NuleculeBase(object):
         # If provider_key isn't provided via CLI, let's grab it the configuration
         if provider_key is None:
             provider_key = self.config.get(GLOBAL_CONF)[PROVIDER_KEY]
+
+        # get a list of all the providers whose artifacts are specified in
+        # Nulecule file into 'providers'
+        providers = []
+        for graph in self.graph:
+            providers.extend(graph['artifacts'].keys())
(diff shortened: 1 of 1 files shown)


surajssd:
I think I broke many tests.

cdrage:
To be honest, I don't understand what's going on here.

Are you trying to have Atomic App fail if there is no provider files in the specific artifact folder?

surajssd:
@cdrage no its not about files being in a folder, the bug here is, if you don't mention the `artifacts` in `Nulecule` file for a particular `provider` and give that `provider` while deployment as a parameter, Atomic App goes ahead with the deployment without having the `provider` `artifacts`. 
So as per https://github.com/projectatomic/atomicapp/issues/641#issue-141825784 you see the default `provider` is `docker` and there are no `artifacts` for `docker` in the file. But while running you can see that Atomic App does not fail. Actually it does not deploy but just says it has deployed.
With this patch I am making a list of `providers` whose `artifacts` are available in `graph` and then checking if the `provider` to deploy to is available in that list.

cdrage:
Hey @surajssd any updates to this? could you make the tests pass? :)

Issue #683, opened 2016-04-07
Atomic App creates a random folder


surajssd:
When using `atomicapp stop` it is creating a random folder in `/var/lib/atomicapp`

Before running the command the directory `/var/lib/atomicapp` does not even exists
```bash
(atomic) [vagrant@centos7-adb ~]$ ll /var/lib/atomicapp
ls: cannot access /var/lib/atomicapp: No such file or directory
```

This is how I call the command (I know it is wrong way to call it)
```bash
(atomic) [vagrant@centos7-adb ~]$ sudo atomicapp stop projectatomic/helloapache
[INFO] - main.py - Action/Mode Selected is: stop
[ERROR] - main.py - Provided answers file doesn't exist: /var/lib/atomicapp/projectatomic-helloapache-fc223bf89cd6/answers.conf.gen
```

Here is a directory created, which it should not.
```bash
(atomic) [vagrant@centos7-adb ~]$ ll /var/lib/atomicapp
total 0
drwxr-xr-x. 2 root root 6 Apr  7 08:37 projectatomic-helloapache-fc223bf89cd6
```

Based on the discussions on #630

dustymabe:
You are still using the command wrong, but I agree no folder should be created.

PR #686, opened 2016-04-08
Atomic App will not create redundant folder on `atomicapp stop`

surajssd:
Previously when `atomicapp stop projectatomic/helloapache` was called atomicapp would create a redundant folder in `/var/lib/atomicapp` and fail saying that `answers` file not found. Though it was failing but the folder would still remain there, now there are checks added to see if user has provided
the image name and the action is stop, the at the location where it creates a directory, the new checks will cause it to fail.

Fixes issue #683

### DIFF
--- a/atomicapp/nulecule/main.py
+++ b/atomicapp/nulecule/main.py
@@ -51,7 +51,7 @@ class NuleculeManager(object):
     """
 
     def __init__(self, app_spec, destination=None,
-                 cli_answers=None, answers_file=None):
+                 cli_answers=None, answers_file=None, action=None):
         """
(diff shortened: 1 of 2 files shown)


atomicapp/nulecule/main.py
```diff
@@ -104,7 +104,10 @@ def __init__(self, app_spec, destination=None,
             if destination:
                 self.app_path = destination
             else:
-                self.app_path = Utils.getNewAppCacheDir(self.image)
+                if action == 'stop':
+                    raise NuleculeException("Provide the path where application is deployed.")
```
surajssd: Need help with this message. I want to make it sound better.

Issue #620, opened 2016-03-10
Docker id output should be in logging format


cdrage:
It's annoying to see this:

```
1457647049 - [DEBUG] - atomicapp/utils.py - Recursively removing directory: /tmp/nulecule-152a8ebc-e70b-11e5-acfd-3402868080db
1457647049 - [DEBUG] - nulecule/container.py - Removing Docker container: /usr/bin/docker rm -f 41fb827975d57e57c9e6fbb8e0afbf1c7235ce64a8ca5b6ae915bf5cd8e63836
41fb827975d57e57c9e6fbb8e0afbf1c7235ce64a8ca5b6ae915bf5cd8e63836
```

The container that is created should either be hidden (put in verbose output) or in-line with the logging formatter.

cdrage:
This should be fixed by the second part of #594

dustymabe:
I'm +1 on this. Would be nice to have it only output on debug.. and even then maybe we can chop it off so it is just a short uuid (i.e use the first 12 characters)

abhi1004:
Hi @dustymabe can I with this ?

abhi1004:
Hi @dustymabe can I  help with this?

dustymabe:
@abhi1004 go for it!

surajssd:
This is also spotted else where, which can be re-created as follows:

```bash
(atomic) ✔ ~/nulecule-library/apache-centos7-atomicapp [master ↑·14|…4] 
04:45 $ sudo atomicapp run . -v --provider docker
1460364339 - [INFO] - cli/main.py - Action/Mode Selected is: run
1460364339 - [DEBUG] - cli/main.py - Final parsed cmdline: run . -v --provider docker
1460364339 - [DEBUG] - nulecule/main.py - NuleculeManager init app_path: .
1460364339 - [DEBUG] - nulecule/main.py - NuleculeManager init image: None
1460364339 - [DEBUG] - nulecule/main.py - Request to unpack to None to .
1460364339 - [DEBUG] - atomicapp/plugin.py - Loading providers from /home/vagrant/atomicapp/atomicapp/providers
1460364339 - [DEBUG] - atomicapp/plugin.py - Loading providers from /home/vagrant/atomicapp/atomicapp/providers
1460364339 - [DEBUG] - atomicapp/plugin.py - Found provider <class 'docker.DockerProvider'>
1460364339 - [WARNING] - atomicapp/plugin.py - Configuration option 'providerconfig' not found
[...]

PR #687, opened 2016-04-11
Container

abhi1004:
This pr is for issue #620, Print only first 12 characters of container_id

### DIFF
--- a/atomicapp/nulecule/container.py
+++ b/atomicapp/nulecule/container.py
@@ -160,7 +160,7 @@ class DockerHandler(object):
         Utils.rm_dir(tmpdir)
 
         # Clean up dummy container
-        rm_cmd = [self.docker_cli, 'rm', '-f', container_id]
+        rm_cmd = [self.docker_cli, 'rm', '-f', container_id[:12]]
         logger.debug('Removing Docker container: %s' % ' '.join(rm_cmd))
         subprocess.call(rm_cmd)
 


atomicapp/nulecule/container.py
```diff
@@ -160,7 +160,7 @@ def extract(self, image, source, dest, update=False):
         Utils.rm_dir(tmpdir)
 
         # Clean up dummy container
-        rm_cmd = [self.docker_cli, 'rm', '-f', container_id]
+        rm_cmd = [self.docker_cli, 'rm', '-f', container_id[:12]]
```
surajssd: IMO using a CONSTANT rather than hard-coding the number would be a good thing to do.

surajssd:
@abhi1004 you can try something like:

```python
# Clean up dummy container
rm_cmd = [self.docker_cli, 'rm', '-f', container_id]
logger.debug('Removing Docker container: %s' % ' '.join(rm_cmd))
output = subprocess.check_output(rm_cmd)
logger.debug('Removed container: {}'.format(output.rstrip()[:CONTAINER_ID_LENGTH]))
```

atomicapp/nulecule/container.py
cdrage: Agreed with @surajssd 

In regards to this command, is it possible to have it output via the "logger.info" command?

dustymabe: > In regards to this command, is it possible to have it output via the "logger.info" command?

Actually I think that would probably not be what we want.. We want to hide some of the complexity here right? I'm about to make another comment in the generic comments that should clear up what we want to do with this PR. Please read that and give me your input.

dustymabe:
After looking at this problem closer I think we can actually leave the logger.debug statement as it currently is. The line that we want to get rid of is the uuid that gets output not through the logger at all. So what we need to do for that is capture the output and don't let it get printed to stdout. 

To do this we can easily just convert the `subprocess.call(rm_cmd)` to `
Utils.run_cmd(rm_cmd, checkexitcode=True)` or to `subprocess.check_output(rm_cmd)`. 

This should not print anything to stdout if the command succeeds.

dustymabe:
> Though I am skeptical about having two logger.debug for the same action.

@surajssd right. probably don't need both.

abhi1004:
@dustymabe tried `subprocess.check_output(rm_cmd)` it works.

cdrage:
Hey @abhi1004 I'd love to get this in :) 

Are you able to update your PR with the changes from the comments? `subprocess.check_output(rm_cmd)` and we can go ahead with the merge? Remember to update the section when it's creating the container too! :dancers:

cdrage:
Hi @abhi1004 I've created https://github.com/projectatomic/atomicapp/pull/726 to replace this :)

abhi1004:
@cdrage Sorry I was not available since I was busy with my exams, and thanks for the update. :+1:

PR #690, opened 2016-04-12
init: modify docker template to publish to host port 80

dustymabe:
This should make it easier for a user to demo without having to figure
out what random port was assigned to be used.


dustymabe:
#dotests

cdrage:
LGTM

Issue #693, opened 2016-04-13
Certificate passing param in `make_rest_request`


surajssd:
Currently `make_rest_request` function in `utils` don't support passing local certificate as client side certificate. So adding additional parameter to it, called as `cert`, would help.

dustymabe:
hey @surajssd let's let #533 track this work - don't need this issue too.

Issue #533, opened 2016-01-26
Support authenticating via client certificate for Openshift


dustymabe:
You can authenticate to Openshift either with an OAuth token or client cert/key. Currently we only support the token method. 

We should support client cert/key as well. See the following for an example of how to use the python-requests library to do the auth:

http://docs.python-requests.org/en/latest/user/advanced/#ssl-cert-verification

sub-mod:
+1

surajssd:
Can I work on this? If nobody is already doing so?

cdrage:
@surajssd all yours!

dustymabe:
@surajssd.. so @kadel and I had a discussion about this when I first opened the ticket, but the discussion didn't make it back into the ticket. 

What we have right now is `accesstoken` and that is the way we do auth. We'd like to change this to `providerauth`, or something generic like that, and allow the user to pass in either an Oauth access token or a path to client cert/key via that one config item. The client cert/key would be two files separated by a `;` or something similar. 

The client may also have a single file with cert/key combined so we would need to support that as well. These are the 3 options:

- OAuth access token
- client cert/key in two separate files
- cert/key in combined file (single file)

kadel:
There is also another think we should support regarding `providerconfig`

There are two options how client cert/key can be specified in `providerconfig`
- `client-key-data` and `client-certificate-data`
- `client-key` and `client-certificate`

when first one (`*-data`) is used cert and key is embeded inside configfile as values of those keys
when second one is used it points to file with key/cert

dustymabe:
@surajssd - how's the progress on this coming?

surajssd:
I am not clear on the concept of how to use certificates to authenticate user to OpenShift? From what I understand is for now we support in Atomic App as: 
`$ oc login --token=XZ0UQoKHiMY411Hs4Q3uyfAunCBovRSpquberltvBiM` 
(where this token is taken from `~/.kube/config` and this works fine, I get logged in as the specific user)

Now for using certificates we are gonne use this method authentication:
`$ oc login --certificate-authority='path-to-crt'`
So the doubt is where do I get these certificates from? How to generate them? I have looked already into the documentation of OpenShift and it is not helpful, I followed this doc and it does not lead anywhere https://docs.openshift.org/latest/architecture/additional_concepts/authentication.html#api-authentication

dustymabe:
[...]

PR #694, opened 2016-04-13
Added cert param in `make_rest_request` to pass client side certs

surajssd:
Added cert param so that now we can pass client side certs while making request to API server, this can be passed as path to file that contains cert and key or it can be a tuple which has path ti cert and key.

This is part of the effort to add support to authenticate to server via cert in addition to access_token.

### DIFF
--- a/atomicapp/utils.py
+++ b/atomicapp/utils.py
@@ -432,7 +432,7 @@ class Utils(object):
         return home
 
     @staticmethod
-    def make_rest_request(method, url, verify=True, data=None):
+    def make_rest_request(method, url, verify=True, data=None, cert=None):
         """
         Make HTTP request to url
 
@@ -444,6 +444,10 @@ class Utils(object):
                                   of trusted CAs
             data (dict/list): object to be serialised to json and send as http
                               data (when method=post/put/delete)
+            cert (tuple/string): Path to local cert as client side certificate,
+                                 as a single file (containing the certificate
+                                 and the private key) or as a tuple of both
+                                 file's path
 
         Returns:
             tuple (status_code, return_data): status_code - http status code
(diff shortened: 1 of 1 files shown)


surajssd:
This will fix #693 and add to effort going on with #533

dustymabe:
Hey @surajssd this looks great but let's create a PR that solves #533 and includes this commit. You can open a WIP (work in progress) PR if you'd like show progress and gather feedback. 

Also, you don't need an issue in order to create a PR. Just open the PR  :-P

surajssd:
@dustymabe I and @kadel had a discussion and we decided to work on this together, so we divided the tasks into sub-tasks which can be done in separate PRs. I feel that will introduce less bugs, compared to adding a large chunk of code once. Please suggest.

cdrage:
@surajssd Ideally it would be best if one of you just submit the major PR and if you need to change parts of the code you can send a .patch to @kadel for him to edit the code :) This would prevent any major breakage as we'd have to implement and merge in two separate PR's ideally at the same exact time.

dustymabe:
while I have not done this before I think it is possible to do:

you can allow push access to your fork from Tomas and then you can both push changes to the PR that you submit (by updating commits on the branch in your forked repo). If both of you have push access to your repo then you should be able to do that without any issue.

rtnpro:
On Apr 14, 2016 12:00 AM, "Dusty Mabe" <notifications@github.com> wrote:
>
> while I have not done this before I think it is possible to do:
>
> you can allow push access to your fork from Tomas and then you can both
push changes to the PR that you submit (by updating commits on the branch
in your forked repo). If both of you have push access to your repo then you
should be able to do that without any issue.

:+1 This should just work fine.

surajssd:
Closing this PR in response to the PR that @kadel is sending!

Issue #529, opened 2016-01-25
support relative path for ca specified in kubeconfig file


dustymabe:
Apparently it is possible for the path to the ca within the kubeconfig files to be specified as a path relative to the kubeconfig file itself. It is done this way in the CDK. 

When parsing the kubeconfig file we should try to handle this. If it is a relative path then just find the dirname of the kubeconfig file and tack the path to the ca onto the end of it.

dustymabe:
This really applies to any certificate that is provided in the kubeconfig file, not just the ca.

PR #696, opened 2016-04-18
[WIP] Add support for authenticating via client certificate for Openshift

### LINKED ISSUE #533: Support authenticating via client certificate for Openshift

kadel:
fixes #533


atomicapp/providers/openshift.py
```diff
@@ -593,10 +610,11 @@ def _get_url(self, namespace, kind, name=None, params=None):
...
 
-        if params:
-            params["access_token"] = self.access_token
-        else:
-            params = {"access_token": self.access_token}
+        if not params:
+            params = []
+
+        if self.access_token:
+                params = {"access_token": self.access_token}
```
surajssd: Extra tab given here

atomicapp/providers/openshift.py
```diff
@@ -593,10 +610,11 @@ def _get_url(self, namespace, kind, name=None, params=None):
         if name:
             url = urljoin(url, name)
 
-        if params:
-            params["access_token"] = self.access_token
-        else:
-            params = {"access_token": self.access_token}
+        if not params:
+            params = []
```
surajssd: This should be a dict by default?

kadel: yes, my bad, it should be dict

atomicapp/providers/openshift.py
kadel: that is wrong you are right :-)

dustymabe:
while the [WIP] text is in the name I assume you guys are still working on this. let me know when you want review.

kadel:
#dotests

atomicapp/providers/openshift.py
```diff
@@ -593,10 +610,11 @@ def _get_url(self, namespace, kind, name=None, params=None):
         if name:
             url = urljoin(url, name)
 
-        if params:
-            params["access_token"] = self.access_token
-        else:
-            params = {"access_token": self.access_token}
+        if not params:
+            params = {}
```
dustymabe: just set `params={}` in the function declaration. That way `{}` is the default value given if params isn't provided right? I guess someone could provide params=None though :(

dustymabe:
Is providing a cert+key in a single file (as mentioned [here](https://github.com/projectatomic/atomicapp/issues/533#issuecomment-183360758)) a thing? If so are we supporting it with this work?

dustymabe:
Also, is there any chance we can bundle this bug in as part of this work: https://github.com/projectatomic/atomicapp/issues/529

atomicapp/providers/lib/kubeconfig.py
```diff
@@ -103,16 +106,52 @@ def parse_kubeconf_data(kubecfg):
...
+
+        if tls_verify:
+            ca_data = cluster["cluster"].get("certificate-authority-data")
+            if ca_data:
+                ca = Utils.getTmpFile(b64decode(ca_data))
+            else:
+                if "certificate-authority" in cluster["cluster"]:
+                    # if we are in container translate path to path on host
+                    ca = os.path.join(Utils.getRoot(),
+                                      cluster["cluster"].get("certificate-authority").lstrip('/'))
```
dustymabe: really long line.. can we break it up into two? maybe `cluster["cluster"].get("certificate-authority").lstrip('/')` on the first and the `join` on the 2nd.

dustymabe: "into two" meaning we would need to assignments probably
```
ca = cluster["cluster"].get("certificate-authority").lstrip('/')
ca = os.path.join(Utils.getRoot(), ca)

File: atomicapp/providers/lib/kubeconfig.py [hunk 5]
```diff
@@ -103,16 +106,52 @@ def parse_kubeconf_data(kubecfg):
...
+
+            cert_data = user["user"].get("client-certificate-data")
+            key_data = user["user"].get("client-key-data")
+
+            if cert_data:
+                cert = Utils.getTmpFile(b64decode(cert_data))
+            else:
+                if "client-certificate" in user["user"]:
+                    cert = os.path.join(Utils.getRoot(),
+                                        user["user"].get("client-certificate").lstrip('/'))
```
dustymabe: really long line

atomicapp/providers/lib/kubeconfig.py
```diff
@@ -103,16 +106,52 @@ def parse_kubeconf_data(kubecfg):
...
+
+            if key_data:
+                key = Utils.getTmpFile(b64decode(key_data))
+            else:
+                if "client-key" in user["user"]:
+                    key = os.path.join(Utils.getRoot(),
+                                       user["user"].get("client-key").lstrip('/'))
```
dustymabe: really long line

atomicapp/providers/lib/kubeconfig.py
```diff
@@ -103,16 +106,52 @@ def parse_kubeconf_data(kubecfg):
...
+        if not auth:
+            # If token not specified, check for certificate auth.
+
+            # client-certificate-data and client-key-data options overrides
+            # client-certificate and client-key
+            # https://github.com/kubernetes/kubernetes/blob/v1.2.2/pkg/client/unversioned/clientcmd/api/types.go#L78
+            # `{client-certificate,client-key}-data` keys in Kubernetes config
+            # file are inline base64 encoded certificates, requests library
+            # requires certs in files, this is why we are putting them to tmp
+            # files.
```
dustymabe: can we add a few more users with examples of these cases to the `Example of expected file format:` section of the parse_kubeconf function description? 

you can truncate output for brevity, just would like a code reader to be able to reference something so they can better understand this code.

atomicapp/providers/lib/kubeconfig.py
```diff
@@ -103,16 +106,52 @@ def parse_kubeconf_data(kubecfg):
...
+
+            cert_data = user["user"].get("client-certificate-data")
+            key_data = user["user"].get("client-key-data")
+
+            if cert_data:
+                cert = Utils.getTmpFile(b64decode(cert_data))
```
dustymabe: Do these tmp files ever get cleaned up? If not we should probably try to make sure they do. I know this makes it harder.

atomicapp/providers/lib/kubeconfig.py
```diff
@@ -103,16 +106,52 @@ def parse_kubeconf_data(kubecfg):
...
+        if tls_verify:
+            ca_data = cluster["cluster"].get("certificate-authority-data")
+            if ca_data:
+                ca = Utils.getTmpFile(b64decode(ca_data))
+            else:
+                if "certificate-authority" in cluster["cluster"]:
+                    # if we are in container translate path to path on host
+                    ca = os.path.join(Utils.getRoot(),
+                                      cluster["cluster"].get("certificate-authority").lstrip('/'))
+
```
dustymabe: Should probably add a comment above the if statement below that states something like: "We will give precedence to an access token if it was provided in the kube-config, if not given an access token then check for client certs."

atomicapp/providers/lib/kubeconfig.py
```diff
@@ -103,16 +106,52 @@ def parse_kubeconf_data(kubecfg):
...
+                                        user["user"].get("client-certificate").lstrip('/'))
+
+            if key_data:
+                key = Utils.getTmpFile(b64decode(key_data))
+            else:
+                if "client-key" in user["user"]:
+                    key = os.path.join(Utils.getRoot(),
+                                       user["user"].get("client-key").lstrip('/'))
+
+            auth = "{}:{}".format(cert, key)
```
dustymabe: Add a comment that explains what this format is.. i.e. certfile:keyfile, because that is the format the provider expects it in.

atomicapp/providers/lib/kubeconfig.py
dustymabe: might be able to use atexit to register these to be deleted. https://docs.python.org/2/library/atexit.html

dustymabe:
Great job guys.. Sorry about all the comments :( - we are really close though

atomicapp/providers/lib/kubeconfig.py
kadel: Now when #705 is merged, I'm going to shorted this by using `get_real_abspath`

cdrage:
@kadel @dustymabe @surajssd 

Shout... I forgot to mention that I already implemented this in `kubebase.py` in my PR...

See https://github.com/projectatomic/atomicapp/pull/712/files

Specifically, `kubebase.py` functions: `_connection()` and `_certfile()`...

cdrage:
It also includes your issue of base64 encoding (converts files to base64, etc.) as well as cert requests when specifying a certificate.

Originally, i was implementing this for the kubernetes api, but because of how similar both the openshift + kubernetes api http implementations are, the cert passing works for both.

atomicapp/providers/lib/kubeconfig.py
cdrage: hey @kadel and @dustymabe feel free to look at my implementation `def cert_file` under https://github.com/projectatomic/atomicapp/pull/712/files (specifically, `kubebase.py`) for how I implemented the base64 encoding / files and such.

dustymabe:
@cdrage so you're saying that #533 is handled by #712?

cdrage:
@dustymabe yes sir

surajssd:
@dustymabe @cdrage is it time to close this PR?

containscafeine:
@cdrage since work around this is being done in #743, let's close this? WDYT?

PR #700, opened 2016-04-19
Update README.md with correct installation instructions.

cdrage:
Forgot to update the release version in README.md

Also the commands for installing via tar.gz were wrong (cd into a
.tar.gz)


dustymabe:
LGTM

PR #705, opened 2016-04-20
Abstracted the way we get absolute path

surajssd:
Added a `staticmethod` in `utils.Utils` named `get_abspath` which abstracts the use of `os.path.join(Utils.getRoot, path)`

### DIFF
--- a/atomicapp/cli/main.py
+++ b/atomicapp/cli/main.py
@@ -474,7 +474,7 @@ class CLI():
             if hasattr(args, item) and getattr(args, item) is not None:
                 args.cli_answers[item] = getattr(args, item)
 
-        lock = LockFile(os.path.join(Utils.getRoot(), LOCK_FILE))
+        lock = LockFile(Utils.get_real_abspath(LOCK_FILE))
         try:
             if args.action != 'init':
                 lock.acquire(timeout=-1)
--- a/atomicapp/utils.py
+++ b/atomicapp/utils.py
@@ -354,6 +354,21 @@ class Utils(object):
         else:
             return "/"
 
+    @staticmethod
+    def get_real_abspath(path):
+        """
+        Take the user provided 'path' and return the real path to the resource
+        irrespective of the app running location either inside container or
+        outside.
+
+        Args:
+            path (str): path to a resource
+
+        Returns:
+            str: absolute path to resource in the filesystem.
+        """
+        return os.path.join(Utils.getRoot(), path.lstrip('/'))
+
     # generates a unique 12 character UUID
     @staticmethod
     def getUniqueUUID():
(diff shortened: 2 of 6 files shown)


atomicapp/utils.py
```diff
@@ -354,6 +354,10 @@ def getRoot():
         else:
             return "/"
 
+    @staticmethod
+    def get_abspath(path):
+        return os.path.join(Utils.getRoot(), path)
```
kadel: I think that we should add `.lstrip('/')` to `path` here, as it should be always called with `lstrip('/')`

`getRoot` returns '/host' or '/'
with leading '/' in second argument `os.path.join` cuts first part in path
```
>>> os.path.join("/host","/foo/bar")
'/foo/bar'

>>> os.path.join("/host","foo/bar")
'/host/foo/bar'
```

surajssd:
@kadel done!

kadel:
#dotests

kadel:
#dotestkubernetes

atomicapp/utils.py
```diff
@@ -354,6 +354,10 @@ def getRoot():
         else:
             return "/"
 
+    @staticmethod
+    def get_abspath(path):
+        return os.path.join(Utils.getRoot(), path.lstrip('/'))
```
dustymabe: Add a comment about what this function is doing if you don't mind.

atomicapp/utils.py
```diff
@@ -354,6 +354,10 @@ def getRoot():
         else:
             return "/"
 
+    @staticmethod
+    def get_abspath(path):
```
dustymabe: should we name it something like get_real_abspath() ?

dustymabe:
kube test failed due to yum error reaching repos. Trying again.

dustymabe:
#dotestkubernetes

surajssd:
done as you said also rebased!

dustymabe:
#dotests

dustymabe:
LGTM - @cdrage.. do you mind spot checking this before merge?

atomicapp/cli/main.py
```diff
@@ -474,7 +474,7 @@ def run(self):
             if hasattr(args, item) and getattr(args, item) is not None:
                 args.cli_answers[item] = getattr(args, item)
 
-        lock = LockFile(os.path.join(Utils.getRoot(), LOCK_FILE))
+        lock = LockFile(Utils.get_real_abspath(LOCK_FILE))
```
cdrage: I'll have to rebase this on my PR on removing the lockfile, but that's fine :)

cdrage:
Tests pass locally, code looks a-okay! LGTM :+1: Merge when ready

dustymabe:
#dotestopenshift

PR #706, opened 2016-04-20
Remove locking from Atomic App

cdrage:
Removed due to numerous issues with locking (dev + test environments) as well as
the deprecation of LockFile.

### DIFF
--- a/atomicapp/cli/main.py
+++ b/atomicapp/cli/main.py
@@ -22,8 +22,6 @@ import sys
 
 import argparse
 import logging
-from lockfile import LockFile
-from lockfile import AlreadyLocked
 
 from atomicapp.applogging import Logging
 from atomicapp.constants import (__ATOMICAPPVERSION__,
@@ -33,7 +31,6 @@ from atomicapp.constants import (__ATOMICAPPVERSION__,
                                  APP_ENT_PATH,
                                  CACHE_DIR,
                                  HOST_DIR,
-                                 LOCK_FILE,
                                  LOGGER_DEFAULT,
                                  PROVIDERS)
 from atomicapp.nulecule import NuleculeManager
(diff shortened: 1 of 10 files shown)


atomicapp/cli/main.py
```diff
@@ -474,31 +471,6 @@ def run(self):
             if hasattr(args, item) and getattr(args, item) is not None:
                 args.cli_answers[item] = getattr(args, item)
 
-        lock = LockFile(os.path.join(Utils.getRoot(), LOCK_FILE))
```
cdrage: one sec. i'll fix this

cdrage:
#dotest

cdrage:
#dotests

dustymabe:
unittests fail

cdrage:
#dotests

cdrage:
Re-doing tests

dustymabe:
#dotestsdocker

dustymabe:
#dotestkubernetes

dustymabe:
#dotestdocker

cdrage:
All passed :+1:

dustymabe:
> We're removing locking in favour of each provider handling the way that
> each artifact is deployed.
> 
> For example: k8s checks the namespaces, docker checks the docker name
> before deployment, oc checks either the configuration or namespace

I don't think the intent of locking was really to ever do anything other than to prevent multiple instances of atomicapp from running at the same time (similar to the way yum prevents multiple instances of itself from running at the same time). So your statement doesn't really apply.

If anything we are simply removing locking now because it is causing more problems than it solves. If we get to a point where people are scripting the use of atomicapp and have issues with multiple instances running at the same time then we will add it back.

cdrage:
@dustymabe 

Updated the PR as well as commit message.

dustymabe:
#dotests

dustymabe:
#dotestkubernetes

cdrage:
Looks to be timeout issues / timing issues again. Going to run this locally as well and see what I get.

dustymabe:
@cdrage LGTM.. if you can get the tests to pass then merge away.

cdrage:
#dotests

dustymabe:
#dotestkubernetes

Issue #708, opened 2016-04-20
Missing /usr/bin/kubectl on kubernetes tests


cdrage:
For some odd reason I'm getting this:

```
1461183851 - [DEBUG] - atomicapp/plugin.py - Found provider <class 'kubernetes.KubernetesProvider'>
1461183851 - [WARNING] - atomicapp/plugin.py - Configuration option 'provider-config' not found
1461183851 - [DEBUG] - providers/kubernetes.py - Given config: {u'image': u'centos/httpd', u'hostport': 80, u'namespace': u'default', u'provider': u'kubernetes'}
1461183851 - [INFO] - providers/kubernetes.py - Using namespace default
1461183851 - [INFO] - providers/kubernetes.py - trying kubectl at /usr/bin/kubectl
1461183851 - [INFO] - providers/kubernetes.py - trying kubectl at /usr/local/bin/kubectl
1461183851 - [ERROR] - cli/main.py - No kubectl found in /usr/bin/kubectl:/usr/local/bin/kubectl
```

Despite having k8s on my localhost

```
▶ whereis kubectl
kubectl: /usr/bin/kubectl /usr/local/bin/kubectl
```

cdrage:
Hmmm...
```sh
▶ sudo atomic run projectatomic/helloapache -v
docker run -it --rm  --privileged -v /home/wikus:/atomicapp -v /run:/run -v /:/host --net=host --name helloapache -e USER -e SUDO_USER -e NAME=helloapache -e IMAGE=projectatomic/helloapache projectatomic/helloapache  run  -v
docker run -it --rm  --privileged -v /home/wikus:/atomicapp -v /run:/run -v /:/host --net=host --name helloapache -e USER -e SUDO_USER -e NAME=helloapache -e IMAGE=projectatomic/helloapache projectatomic/helloapache  run  -v
1461184342 - [INFO] - cli/main.py - Action/Mode Selected is: run
1461184342 - [DEBUG] - cli/main.py - Final parsed cmdline: run -v
1461184342 - [DEBUG] - cli/main.py - Setting app_spec based on $IMAGE env var
1461184342 - [DEBUG] - nulecule/main.py - NuleculeManager init app_path: /var/lib/atomicapp/projectatomic-helloapache-7ba5fe96ca32
1461184342 - [DEBUG] - nulecule/main.py - NuleculeManager init image: projectatomic/helloapache
[...]

PR #709, opened 2016-04-20
Remove dockerenv and dockerinit check

cdrage:
Found this out when using Docker 1.11

.dockerenv and .dockerinit files have been removed in 1.10 of Docker due
to not using the LXC built-in exec driver anymore and thus the files no
longer appear within Atomic App and fail to pick up whether or not we
are running in a container or not.

See: https://docs.docker.com/engine/deprecated/
```
LXC built-in exec driver
Deprecated In Release: v1.8

Target For Removal In Release: v1.10

The built-in LXC execution driver is deprecated for an external
implementation. The lxc-conf flag and API fields will also be removed.
```

### DIFF
--- a/atomicapp/utils.py
+++ b/atomicapp/utils.py
@@ -342,7 +342,7 @@ class Utils(object):
         Returns:
             (bool): True == we are in a container
         """
-        if os.path.isfile('/.dockerenv') and os.path.isfile('/.dockerinit') and os.path.isdir(HOST_DIR):
+        if os.path.isdir(HOST_DIR):
             return True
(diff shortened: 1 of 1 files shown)


cdrage:
Fixes #708

cdrage:
#dotests

dustymabe:
too bad we just added this. wtf :) - sucks that it actually made it into a release though. a pretty big one too.

dustymabe:
#dotestkubernetes

dustymabe:
LGTM

PR #712, opened 2016-04-25
Add a new 'library' for k8s/openshift providers.

cdrage:
This is the first commit and first initialization into refactoring the
OpenShift provider in order to create a library that is both compatible
with OpenShift as well as Kubernetes.

With this 'library', the *ONLY* thing that is passed in is an object of
the .kube/config configuration. After passing the configuration, you may
make API calls such as .create(object) and .delete(object).

This commit integrates the library as well as converts the current Kubernetes provider from cmd-line (using kubectl) to using the API.

### DIFF
--- a/atomicapp/providers/kubernetes.py
+++ b/atomicapp/providers/kubernetes.py
@@ -20,14 +20,21 @@
 import anymarkup
 import logging
 import os
-from string import Template
 
-from atomicapp.constants import (LOGGER_COCKPIT,
+from atomicapp.constants import (PROVIDER_AUTH_KEY,
+                                 ANSWERS_FILE,
+                                 DEFAULT_NAMESPACE,
                                  LOGGER_DEFAULT,
-                                 PERSISTENT_STORAGE_FORMAT)
+                                 PROVIDER_API_KEY,
+                                 PROVIDER_CA_KEY,
+                                 PROVIDER_TLS_VERIFY_KEY,
+                                 LOGGER_COCKPIT,
+                                 K8S_DEFAULT_API)
 from atomicapp.plugin import Provider, ProviderFailedException
-from atomicapp.utils import Utils
 
+from atomicapp.providers.lib.kubeshift.kubeconfig import KubeConfig
+from atomicapp.providers.lib.kubeshift.client import Client
+from atomicapp.utils import Utils
 cockpit_logger = logging.getLogger(LOGGER_COCKPIT)
 logger = logging.getLogger(LOGGER_DEFAULT)
 
@@ -38,155 +45,227 @@ class KubernetesProvider(Provider):
     This class implements deploy, stop and undeploy of an atomicapp on
     Kubernetes provider.
     """
+
+    # Class variables
     key = "kubernetes"
+    namespace = DEFAULT_NAMESPACE
+    k8s_artifacts = {}
+
+    # From the provider configuration
     config_file = None
-    kubectl = None
+
+    # Essential provider parameters
+    provider_api = None
+    provider_auth = None
+    provider_tls_verify = None
+    provider_ca = None
 
     def init(self):
-        self.namespace = "default"
-
-        self.k8s_manifests = []
+        self.k8s_artifacts = {}
 
         logger.debug("Given config: %s", self.config)
         if self.config.get("namespace"):
             self.namespace = self.config.get("namespace")
 
         logger.info("Using namespace %s", self.namespace)
-        if self.container:
-            self.kubectl = self._find_kubectl(Utils.getRoot())
-            kube_conf_path = "/etc/kubernetes"
-            host_kube_conf_path = Utils.get_real_abspath(kube_conf_path)
-            if not os.path.exists(kube_conf_path) and os.path.exists(host_kube_conf_path):
-                if self.dryrun:
-                    logger.info("DRY-RUN: link %s from %s" % (kube_conf_path, host_kube_conf_path))
+
+        self._process_artifacts()
+
+        if self.dryrun:
+            return
+
+        '''
+        Config_file:
+            If a config_file has been provided, use the configuration
+            from the file and load the associated generated file.
+            If a config_file exists (--provider-config) use that.
+
+        Params:
+            If any provider specific parameters have been provided,
+            load the configuration through the answers.conf file
+
+        .kube/config:
+            If no config file or params are provided by user then try to find and
+            use a config file at the default location.
+
+        no config at all:
+            If no .kube/config file can be found then try to connect to the default
+            unauthenticated http://localhost:8080/api end-point.
+        '''
+
+        default_config_loc = os.path.join(
+            Utils.getRoot(), Utils.getUserHome().strip('/'), '.kube/config')
+
+        if self.config_file:
+            logger.debug("Provider configuration provided")
+            self.api = Client(KubeConfig.from_file(self.config_file), "kubernetes")
+        elif self._check_required_params():
+            logger.debug("Generating .kube/config from given parameters")
+            self.api = Client(self._from_required_params(), "kubernetes")
+        elif os.path.isfile(default_config_loc):
+            logger.debug(".kube/config exists, using default configuration file")
+            self.api = Client(KubeConfig.from_file(default_config_loc), "kubernetes")
+        else:
+            self.config["provider-api"] = K8S_DEFAULT_API
+            self.api = Client(self._from_required_params(), "kubernetes")
+
+        # Check if the namespace that the app is being deployed to is available
+        self._check_namespaces()
+
+    def _build_param_dict(self):
+        # Initialize the values
+        paramdict = {PROVIDER_API_KEY: self.provider_api,
+                     PROVIDER_AUTH_KEY: self.provider_auth,
+                     PROVIDER_TLS_VERIFY_KEY: self.provider_tls_verify,
+                     PROVIDER_CA_KEY: self.provider_ca}
+
+        # Get values from the loaded answers.conf / passed CLI params
+        for k in paramdict.keys():
+            paramdict[k] = self.config.get(k)
+
+        return paramdict
+
+    def _check_required_params(self, exception=False):
+        '''
+        This checks to see if required parameters associated to the Kubernetes
+        provider are passed. Only PROVIDER_API_KEY is *required*. Token may be blank.
+        '''
+
+        paramdict = self._build_param_dict()
+        logger.debug("List of parameters passed: %s" % paramdict)
+
+        # Check that the required parameters are passed. If not, error out.
+        for k in [PROVIDER_API_KEY]:
+            if paramdict[k] is None:
+                if exception:
+                    msg = "You need to set %s in %s or pass it as a CLI param" % (k, ANSWERS_FILE)
+                    raise ProviderFailedException(msg)
                 else:
-                    os.symlink(host_kube_conf_path, kube_conf_path)
-        else:
-            self.kubectl = self._find_kubectl()
+                    return False
 
-        if not self.dryrun:
-            if not os.access(self.kubectl, os.X_OK):
-                raise ProviderFailedException("Command: " + self.kubectl + " not found")
+        return True
 
-            # Check if Kubernetes config file is accessible, but only
-            # if one was provided by the user; config file is optional.
-            if self.config_file:
-                self.checkConfigFile()
+    def _from_required_params(self):
+        '''
+        Create a default configuration from passed environment parameters.
+        '''
 
-    def _find_kubectl(self, prefix=""):
-        """Determine the path to the kubectl program on the host.
-        1) Check the config for a provider_cli in the general section
-           remember to add /host prefix
-        2) Search /usr/bin:/usr/local/bin
+        self._check_required_params(exception=True)
+        paramdict = self._build_param_dict()
 
-        Use the first valid value found
+        # Generate the configuration from the paramters
+        config = KubeConfig().from_params(api=paramdict[PROVIDER_API_KEY],
+                                          auth=paramdict[PROVIDER_AUTH_KEY],
+                                          ca=paramdict[PROVIDER_CA_KEY],
+                                          verify=paramdict[PROVIDER_TLS_VERIFY_KEY])
+        return config
+
+    def _check_namespaces(self):
+        '''
+        This function checks to see whether or not the namespaces created in the cluster match the
+        namespace that is associated and/or provided in the deployed application
+        '''
+
+        # Get the namespaces and output the currently used ones
+        namespace_list = self.api.namespaces()
+        logger.debug("There are currently %s namespaces in the cluster." % str(len(namespace_list)))
+
+        # Create a namespace list
+        namespaces = []
+        for ns in namespace_list:
+            namespaces.append(ns["metadata"]["name"])
+
+        # Output the namespaces and check to see if the one provided exists
+        logger.debug("Namespaces: %s" % namespaces)
+        if self.namespace not in namespaces:
+            msg = "%s namespace does not exist. Please create the namespace and try again." % self.namespace
+            raise ProviderFailedException(msg)
+
+    def _process_artifacts(self):
         """
-
-        if self.dryrun:
-            # Testing env does not have kubectl in it
-            return "/usr/bin/kubectl"
-
-        test_paths = ['/usr/bin/kubectl', '/usr/local/bin/kubectl']
-        if self.config.get("provider_cli"):
-            logger.info("caller gave provider_cli: " + self.config.get("provider_cli"))
-            test_paths.insert(0, self.config.get("provider_cli"))
-
-        for path in test_paths:
-            test_path = prefix + path
-            logger.info("trying kubectl at " + test_path)
-            kubectl = test_path
-            if os.access(kubectl, os.X_OK):
-                logger.info("found kubectl at " + test_path)
-                return kubectl
-
-        raise ProviderFailedException("No kubectl found in %s" % ":".join(test_paths))
-
-    def _call(self, cmd):
-        """Calls given command
-
-        :arg cmd: Command to be called in a form of list
-        :raises: Exception
-        """
-
-        if self.dryrun:
-            logger.info("DRY-RUN: %s", " ".join(cmd))
-        else:
-            ec, stdout, stderr = Utils.run_cmd(cmd, checkexitcode=True)
-            return stdout
-
-    def process_k8s_artifacts(self):
-        """Processes Kubernetes manifests files and checks if manifest under
-        process is valid.
+        Parse each Kubernetes file and convert said format into an Object for
+        deployment.
         """
         for artifact in self.artifacts:
+            logger.debug("Processing artifact: %s", artifact)
             data = None
+
+            # Open and parse the artifact data
             with open(os.path.join(self.path, artifact), "r") as fp:
-                logger.debug(os.path.join(self.path, artifact))
-                try:
-                    data = anymarkup.parse(fp)
-                except Exception:
-                    msg = "Error processing %s artifcats, Error:" % os.path.join(
-                        self.path, artifact)
-                    cockpit_logger.error(msg)
-                    raise
-            if "kind" in data:
-                self.k8s_manifests.append((data["kind"].lower(), artifact))
-            else:
-                apath = os.path.join(self.path, artifact)
-                raise ProviderFailedException("Malformed kube file: %s" % apath)
+                data = anymarkup.parse(fp, force_types=None)
 
-    def _resource_identity(self, path):
-        """Finds the Kubernetes resource name / identity from resource manifest
-        and raises if manifest is not supported.
+            # Process said artifacts
+            self._process_artifact_data(artifact, data)
 
-        :arg path: Absolute path to Kubernetes resource manifest
-
-        :return: str -- Resource name / identity
-
-        :raises: ProviderFailedException
+    def _process_artifact_data(self, artifact, data):
         """
-        data = anymarkup.parse_file(path)
+        Process the data for an artifact
+
+        Args:
+            artifact (str): Artifact name
+            data (dict): Artifact data
+        """
+
+        # Check if kind exists
+        if "kind" not in data.keys():
+            raise ProviderFailedException(
+                "Error processing %s artifact. There is no kind" % artifact)
+
+        # Change to lower case so it's easier to parse
+        kind = data["kind"].lower()
+
+        if kind not in self.k8s_artifacts.keys():
+            self.k8s_artifacts[kind] = []
+
+        # Fail if there is no metadata
+        if 'metadata' not in data:
+            raise ProviderFailedException(
+                "Error processing %s artifact. There is no metadata object" % artifact)
+
+        # Change to the namespace specified on init()
+        data['metadata']['namespace'] = self.namespace
+
+        if 'labels' not in data['metadata']:
+            data['metadata']['labels'] = {'namespace': self.namespace}
+        else:
+            data['metadata']['labels']['namespace'] = self.namespace
+
+        self.k8s_artifacts[kind].append(data)
+
+    '''
+    This is DEPRECATED and not needed anymore as we check the /resource URL of the kubernetes api against the artifact
+    def _identify_api(self, artifact, data):
+        """
+        Make sure that the artifact is using the correct API
+
+        Args:
+            artifact (str): Artifact name
+            data (dict): Artifact data
+        """
         if data["apiVersion"] == "v1":
-            return data["metadata"]["name"]
+            pass
         elif data["apiVersion"] in ["v1beta3", "v1beta2", "v1beta1"]:
-            msg = ("%s is not supported API version, update Kubernetes "
+            msg = ("%s is not a supported API version, update Kubernetes "
                    "artifacts to v1 API version. Error in processing "
-                   "%s manifest." % (data["apiVersion"], path))
+                   "%s manifest." % (data["apiVersion"], artifact))
             raise ProviderFailedException(msg)
         else:
-            raise ProviderFailedException("Malformed kube file: %s" % path)
-
-    def _scale_replicas(self, path, replicas=0):
-        """Scales replicationController to specified replicas size
-
-        :arg path: Path to replicationController manifest
-        :arg replicas: Replica size to scale to.
-        """
-        rname = self._resource_identity(path)
-        cmd = [self.kubectl, "scale", "rc", rname,
-               "--replicas=%s" % str(replicas),
-               "--namespace=%s" % self.namespace]
-        if self.config_file:
-            cmd.append("--kubeconfig=%s" % self.config_file)
-
-        self._call(cmd)
+            raise ProviderFailedException("Malformed kubernetes artifact: %s" % artifact)
+    '''
 
     def run(self):
-        """Deploys the app by given resource manifests.
+        """
+        Deploys the app by given resource artifacts.
         """
         logger.info("Deploying to Kubernetes")
-        self.process_k8s_artifacts()
 
-        for kind, artifact in self.k8s_manifests:
-            if not artifact:
-                continue
-
-            k8s_file = os.path.join(self.path, artifact)
-
-            cmd = [self.kubectl, "create", "-f", k8s_file, "--namespace=%s" % self.namespace]
-            if self.config_file:
-                cmd.append("--kubeconfig=%s" % self.config_file)
-            self._call(cmd)
+        for kind, objects in self.k8s_artifacts.iteritems():
+            for artifact in objects:
+                if self.dryrun:
+                    logger.info("DRY-RUN: Deploying k8s KIND: %s, ARTIFACT: %s"
+                                % (kind, artifact))
+                else:
+                    self.api.create(artifact, self.namespace)
 
     def stop(self):
         """Undeploys the app by given resource manifests.
@@ -194,71 +273,19 @@ class KubernetesProvider(Provider):
         the resource from cluster.
         """
         logger.info("Undeploying from Kubernetes")
-        self.process_k8s_artifacts()
 
-        for kind, artifact in self.k8s_manifests:
-            if not artifact:
-                continue
+        for kind, objects in self.k8s_artifacts.iteritems():
+            for artifact in objects:
+                if self.dryrun:
+                    logger.info("DRY-RUN: Deploying k8s KIND: %s, ARTIFACT: %s"
+                                % (kind, artifact))
+                else:
+                    self.api.delete(artifact, self.namespace)
 
-            path = os.path.join(self.path, artifact)
-
-            if kind in ["ReplicationController", "rc", "replicationcontroller"]:
-                self._scale_replicas(path, replicas=0)
-
-            cmd = [self.kubectl, "delete", "-f", path, "--namespace=%s" % self.namespace]
-            if self.config_file:
-                cmd.append("--kubeconfig=%s" % self.config_file)
-            self._call(cmd)
-
+    # TODO
     def persistent_storage(self, graph, action):
-        """
-        Actions are either: run, stop or uninstall as per the Requirements class
-        Curently run is the only function implemented for k8s persistent storage
-        """
+        pass
 
-        logger.debug("Persistent storage enabled! Running action: %s" % action)
-
-        if graph["accessMode"] not in PERSISTENT_STORAGE_FORMAT:
-            raise ProviderFailedException("{} is an invalid storage format "
-                                          "(choose from {})"
-                                          .format(graph["accessMode"],
-                                                  ', '.join(PERSISTENT_STORAGE_FORMAT)))
-
-        if action not in ['run']:
-            logger.warning(
-                "%s action is not available for provider %s. Doing nothing." %
-                (action, self.key))
-            return
-
-        self._check_persistent_volumes()
-
-        # Get the path of the persistent storage yaml file includes in /external
-        # Plug the information from the graph into the persistent storage file
-        base_path = os.path.dirname(os.path.realpath(__file__))
-        template_path = os.path.join(base_path,
-                                     'external/kubernetes/persistent_storage.yaml')
-        with open(template_path, 'r') as f:
-            content = f.read()
-        template = Template(content)
-        rendered_template = template.safe_substitute(graph)
-
-        tmp_file = Utils.getTmpFile(rendered_template, '.yaml')
-
-        # Pass the .yaml file and execute
-        if action is "run":
-            cmd = [self.kubectl, "create", "-f", tmp_file, "--namespace=%s" % self.namespace]
-            if self.config_file:
-                cmd.append("--kubeconfig=%s" % self.config_file)
-            self._call(cmd)
-            os.unlink(tmp_file)
-
-    def _check_persistent_volumes(self):
-            cmd = [self.kubectl, "get", "pv"]
-            if self.config_file:
-                cmd.append("--kubeconfig=%s" % self.config_file)
-            lines = self._call(cmd)
-
-            # If there are no persistent volumes to claim, warn the user
-            if not self.dryrun and len(lines.split("\n")) == 2:
-                logger.warning("No persistent volumes detected in Kubernetes. Volume claim will not "
-                               "initialize unless persistent volumes exist.")
+    # TODO
+    def _check_persistent_volumes(self, graph, action):
+        pass
--- a/atomicapp/providers/lib/kubeshift/client.py
+++ b/atomicapp/providers/lib/kubeshift/client.py
@@ -1,0 +1,57 @@
+"""
+ Copyright 2014-2016 Red Hat, Inc.
+
+ This file is part of Atomic App.
+
+ Atomic App is free software: you can redistribute it and/or modify
+ it under the terms of the GNU Lesser General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ Atomic App is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public License
+ along with Atomic App. If not, see <http://www.gnu.org/licenses/>.
+"""
+
+from atomicapp.providers.lib.kubeshift.kubernetes import KubeKubernetesClient
+from atomicapp.providers.lib.kubeshift.exceptions import KubeClientError
+from atomicapp.constants import LOGGER_DEFAULT
+import logging
+logger = logging.getLogger(LOGGER_DEFAULT)
+
+
+class Client(object):
+
+    def __init__(self, config, provider):
+        '''
+
+        Args:
+            config (obj): Object of the configuration data
+            provider (str): String value of the provider that is being used
+
+        '''
+        self.config = config
+        self.provider = provider
+
+        # Choose the type of provider that is being used. Error out if it is not available
+        if provider is "kubernetes":
+            self.connection = KubeKubernetesClient(config)
+            logger.debug("Using Kubernetes Provider KubeClient library")
+        else:
+            raise KubeClientError("No provider by that name.")
+
+    # Create an object using its respective API
+    def create(self, obj, namespace="default"):
+        self.connection.create(obj, namespace)
+
+    # Delete an object using its respective API
+    def delete(self, obj, namespace="default"):
+        self.connection.delete(obj, namespace)
+
+    # Current support: kubernetes only
+    def namespaces(self):
+        return self.connection.namespaces()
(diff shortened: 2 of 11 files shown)


cdrage:
#dotests

cdrage:
#dotests

cdrage:
#dotests

cdrage:
#dotests

cdrage:
#dotests

cdrage:
#dotests

cdrage:
Functional tests failing due to: https://github.com/projectatomic/adb-tests/pull/32/files

cdrage:
So I've gotten this to work with the Kubernetes provider.

The only thing that's left is the conversion of the OpenShift provider to this new refactor. I've left the `lib/openshift.py` code in there in order to pick-up where I left off when we convert said OpenShift to the new 'library' (with certs and all). I can remove that code if you wish.

Otherwise, this PR is ready for review!

Feel free to give it a spin!

dustymabe:
@kadel can we get you to review this?

dustymabe:
I feel like we might need to go deeper than `atomicapp/providers/lib/` because lib could be used for more than just the kube/shift library. i.e. we might want libraries for other providers in the future.

Should we add one level: `atomicapp/providers/lib/kubeshift/`

cdrage:
kubeshift sounds like a good name, I'll make that change @dustymabe

atomicapp/providers/lib/kubebase.py
```diff
@@ -0,0 +1,309 @@
...
+                logger.debug("Verification will be required for all API calls")
+
+        # If we're using a token, use it, otherwise it's assumed the user uses
+        # client-certificate and client-key
+        if "token" in self.user:
+            connection.headers["Authorization"] = "Bearer %s" % self.user["token"]
+
+        # Lastly, if we have client-certificate and client-key in the .kube/config
+        # we add them to the connection as a cert
+        if ["client-certificate", "client-key"] in self.user.keys():
```
kadel: We also need support for `client-certificate-data` and `client-key-data` keys.
If specified those options contain "inlined"  base64 encoded certificate/key.

https://github.com/kubernetes/kubernetes/blob/v1.2.2/pkg/client/unversioned/clientcmd/api/types.go#L83,L84

atomicapp/providers/lib/kubeconfig.py
```diff
@@ -14,6 +14,66 @@
...
+        Params:
+            filename (str): File location
+        '''
+
+        return anymarkup.parse_file(filename)
+
+    @staticmethod
+    def from_params(api=None, auth=None, ca=None, verify=True):
+        '''
+        Creates a .kube/config based upon the arguments given.
```
dustymabe: `Creates a .kube/config` confused me a bit. This doesn't actually create a file, but rather a data structure with the information in it. How about we use something like: `Creates a dict of .kube/config data based upon the arguments given` ?

atomicapp/providers/lib/kubernetes.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def __init__(self, config):
+        '''
+
+        Args:
+            config (obj): Object of the configuration data
+
+        '''
+
+        # The configuration data passed in will be .kube/config data, so process is accordingly.
```
dustymabe: `process is` ?

atomicapp/providers/lib/kubernetes.py
```diff
@@ -0,0 +1,170 @@
...
+        '''
+        Gathers a list of namespaces on the Kubernetes cluster
+        '''
+        url = urljoin(self.k8s_api, "namespaces")
+        ns = self.api.request("get", url)
+        return ns['items']
+
+    def _generate_kurl(self, obj, namespace, name=None, params=None):
+        '''
+        Generate the required URL by extracting the 'kind' from the object as well as the namespace.
```
dustymabe: maybe try to keep things under 80 chars

tests/units/providers/test_kubernetes_provider.py
```diff
@@ -53,7 +53,7 @@ def prepare_provider(self, data):
         return provider
 
     # Check that the provider configuration file exists
-    @mock.patch.object(KubernetesProvider, '_call', mock_provider_call)
+    # @mock.patch.object(KubernetesProvider, '_call', mock_provider_call)
```
dustymabe: No longer needed? Why?

dustymabe:
I'm confused right now by `atomicapp/providers/lib/openshift.py`. 
- We don't use this yet do we? Should we add it in a separate PR? 
- It seems to have very similar code to what is in `lib/kubernetes.py`. I see there is a #TODO to refactor. Is the plan to use inheritance here in the future? would seem logical to me since openshift is "theoretically" a superset of openshift.

atomicapp/providers/lib/kubernetes.py
```diff
@@ -0,0 +1,170 @@
...
+
+    def namespaces(self):
+        '''
+        Gathers a list of namespaces on the Kubernetes cluster
+        '''
+        url = urljoin(self.k8s_api, "namespaces")
+        ns = self.api.request("get", url)
+        return ns['items']
+
+    def _generate_kurl(self, obj, namespace, name=None, params=None):
```
dustymabe: kurl = kind url? 

I wonder if we should make that more explicit

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,234 @@ class KubernetesProvider(Provider):
...
-        1) Check the config for a provider_cli in the general section
-           remember to add /host prefix
-        2) Search /usr/bin:/usr/local/bin
+            Params:
+                If any provider specific parameters have been provided,
+                load the configuration through the answers.conf file
 
-        Use the first valid value found
+            Default configuration:
+                If no paramaters have been provider, as well as --provider-config,
```
dustymabe: s/provider/provided
s/as well as /as well as no /

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,234 @@ class KubernetesProvider(Provider):
...
+                  PROVIDER_CA_KEY: self.provider_ca}
+
+        # Get values from the loaded answers.conf / passed CLI params
+        for k in result.keys():
+            result[k] = self.config.get(k)
+
+        # Check that the required parameter is passed. If not, error out.
+        for k in [PROVIDER_API_KEY]:
+            if result[k] is None:
+                msg = "Kubernetes API access: You need to set %s in %s or pass it as a CLI param" % (k, ANSWERS_FILE)
```
dustymabe: very long line.. would be nice if we go back through at some point and try to keep things under 80 chars

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,234 @@ class KubernetesProvider(Provider):
...
+                If no paramaters have been provider, as well as --provider-config,
+                use the default .kube/config location and load the associated configuration.
+            '''
+            if self.config_file:
+                self.api = Client(KubeConfig.from_file(self.config_file), "kubernetes")
+            elif self._check_required_params():
+                self.api = Client(self._from_required_params(), "kubernetes")
+            else:
+                config_location = os.path.join(Utils.getRoot(), Utils.getUserHome().strip('/'), '.kube/config')
+                self.api = Client(KubeConfig.from_file(config_location), "kubernetes")
```
dustymabe: Before this PR (when using kubectl), kubectl would work even if no ~/.kube/config existed. Is there a sensible default we can use to achieve this same level of convenience?

If none of the previous things were existed then we could assume an endpoint (`http://localhost:8080/api`) and then fire off a test connection and if passed then use that as our config. Thoughts?

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,234 @@ class KubernetesProvider(Provider):
...
+
+            # Check if the namespace that the app is being deployed to is available
+            self._check_namespaces()
+
+    def _check_required_params(self):
+        '''
+        This checks to see if any parameters associated to the Kubernetes provider are passed
```
dustymabe: are we checking to make sure required paramters are passed and then erroring if not? 

What are required parameters for kubernetes? I don't think there are the same requirements as there are for openshift so we might have to change this up a bit. I tried to make a providerconfig and get it to work but there was too much information that was required.

atomicapp/providers/lib/kubebase.py
```diff
@@ -0,0 +1,309 @@
...
+        # Gather the "current-context" from .kube/config which lists what the
+        # associated cluster, user, token, etc. is being used.
+        if "current-context" not in config:
+            raise KubeBaseError("'current-context' needs to be set within .kube/config")
+        else:
+            self.current_context = config["current-context"]
+
+        # Gather the user, context and cluster details of .kube/config based upon the current_context
+        kubeconfig_context = self._contexts()[self.current_context]
+        kubeconfig_user = kubeconfig_context['user']
```
dustymabe: user isn't required.. this config works for me using kubectl:

```
apiVersion: v1
clusters:
- cluster:
    server: "http://localhost:8080"
  name: foo
contexts:
- context:
    cluster: "foo"
  name: foocont
current-context: foocont
kind: Config
preferences: {}
users: []
```

atomicapp/providers/lib/kubebase.py
cdrage: ACK. I'll work on this

atomicapp/providers/lib/kubeconfig.py
cdrage: Updated ^^

atomicapp/providers/lib/kubernetes.py
cdrage: Yup. I added that to the comments. We could do "gen_kind_url" instead, would this be better?

atomicapp/providers/kubernetes.py
cdrage: ACK

atomicapp/providers/kubernetes.py
cdrage: Thanks for pointing this out! Yeah, I'll add this here too. If no .kube/config exists we should try that end-point anyways.

atomicapp/providers/kubernetes.py
cdrage: I'll add a doc for this. Only the API end-point is required for the kubernetes provider.

atomicapp/providers/lib/kubebase.py
cdrage: ACK. I added 'user' due to picking up the "token" information. You are correct that it is not needed. I will update the PR for this edge-case.

cdrage:
#dotests

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,234 @@ class KubernetesProvider(Provider):
...
+        if 'labels' not in data['metadata']:
+            data['metadata']['labels'] = {'namespace': self.namespace}
+        else:
+            data['metadata']['labels']['namespace'] = self.namespace
+
+        self.k8s_artifacts[kind].append(data)
 
-        :return: str -- Resource name / identity
+    '''
+    This is DEPRECATED and not needed anymore as we check the /resource URL of the kubernetes api against the artifact
```
kadel: Can we remove this?

cdrage:
@dustymabe @kadel @rtnpro 

Hey all,

I've updated it based upon @dustymabe 's comments.. 

In regards to your comments, I've fixed the edge-case with the 'user' not being defined in the .kube/config class.

Same goes for when there is no .kube/config. Atomic App will generate the required api credentials.

I've moved it to `kubeshift` as well.

@kadel I haven't added the client-cert yet, but I will soon as I continue to work on this.

I'll also add a doc @dustymabe on what the required params are (just like

atomicapp/providers/lib/kubernetes.py
dustymabe: yes- i think so

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,243 @@ class KubernetesProvider(Provider):
...
+
+            Default configuration:
+                If no paramaters have been provided, as well as no --provider-config,
+                use the default .kube/config location and load the associated configuration.
+            '''
 
-        if self.dryrun:
-            # Testing env does not have kubectl in it
-            return "/usr/bin/kubectl"
+            config_location = os.path.join(Utils.getRoot(), Utils.getUserHome().strip('/'), '.kube/config')
```
dustymabe: what do you think about calling this `default_config` or `default_config_location` or something like that?

cdrage: default_config_loc maybe? I'll change it.

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,243 @@ class KubernetesProvider(Provider):
...
 
-        Use the first valid value found
-        """
+            .kube/config doesn't exist:
+                If the configuration file does not exist, generate one using
+                the default localhost:8080/api end-point.
+
+            Default configuration:
+                If no paramaters have been provided, as well as no --provider-config,
+                use the default .kube/config location and load the associated configuration.
```
dustymabe: I think these last two could use some more clarity. How about: 

```
            .kube/config:
                If no config file or params are provided by user then try to find and
                use a config file at the default location.

            no config at all:
                If no .kube/config file can be found then try to connect to the default
                unauthenticated http://localhost:8080/api end-point.
```

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,243 @@ class KubernetesProvider(Provider):
...
 
-        test_paths = ['/usr/bin/kubectl', '/usr/local/bin/kubectl']
-        if self.config.get("provider_cli"):
-            logger.info("caller gave provider_cli: " + self.config.get("provider_cli"))
-            test_paths.insert(0, self.config.get("provider_cli"))
+            if self.config_file:
+                self.api = Client(KubeConfig.from_file(self.config_file), "kubernetes")
+            elif self._check_required_params():
+                self.api = Client(self._from_required_params(), "kubernetes")
+            elif not os.path.isfile(config_location):
```
dustymabe: I think it would be more intuitive if we make this one `elif os.path.isfile()` and move the localhost:8080 stuff to the `else:`

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,243 @@ class KubernetesProvider(Provider):
...
+    k8s_artifacts = {}
+
+    # From the provider configuration
     config_file = None
-    kubectl = None
 
-    def init(self):
-        self.namespace = "default"
+    # Essential provider parameters
+    provider_api = "http://localhost:8080"
```
dustymabe: the code below that uses this "default" value would actually make more sense if we set `localhost:8080` there instead of here. I would just set this to `None` unless there is a technical reason why we shouldn't. 

also, one could make an argument that this should be a constant??

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,243 @@ class KubernetesProvider(Provider):
...
+            elif self._check_required_params():
+                self.api = Client(self._from_required_params(), "kubernetes")
+            elif not os.path.isfile(config_location):
+                self.config["provider-api"] = self.provider_api
+                self.api = Client(self._from_required_params(), "kubernetes")
+            else:
+                self.api = Client(KubeConfig.from_file(config_location), "kubernetes")
+
+            # Check if the namespace that the app is being deployed to is available
+            self._check_namespaces()
```
dustymabe: is this supposed to raise an exception? It currently does not.

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,243 @@ class KubernetesProvider(Provider):
...
-                if self.dryrun:
-                    logger.info("DRY-RUN: link %s from %s" % (kube_conf_path, host_kube_conf_path))
-                else:
-                    os.symlink(host_kube_conf_path, kube_conf_path)
-        else:
-            self.kubectl = self._find_kubectl()
+
+        self._process_artifacts()
 
         if not self.dryrun:
```
dustymabe: maybe we should just make this: `return if self.dryrun` and then we can unindent the rest of the function?

atomicapp/providers/kubernetes.py
dustymabe: I don't really see anything about this function that has been updated. I still don't understand why it is trying to verify the existence of 4 values when only 1 is required

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,243 @@ class KubernetesProvider(Provider):
...
+            elif not os.path.isfile(config_location):
+                self.config["provider-api"] = self.provider_api
+                self.api = Client(self._from_required_params(), "kubernetes")
+            else:
+                self.api = Client(KubeConfig.from_file(config_location), "kubernetes")
+
+            # Check if the namespace that the app is being deployed to is available
+            self._check_namespaces()
+
+    def _check_required_params(self):
```
dustymabe: btw this is a great example of a function that can be implemented for kubernetes and openshift - since they will probably have different requirements of what is required. openshift could inherit from a library and then "override" this function with one of its own.

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,243 @@ class KubernetesProvider(Provider):
...
+                  PROVIDER_AUTH_KEY: self.provider_auth,
+                  PROVIDER_TLS_VERIFY_KEY: self.provider_tls_verify,
+                  PROVIDER_CA_KEY: self.provider_ca}
+
+        # Get values from the loaded answers.conf / passed CLI params
+        for k in result.keys():
+            result[k] = self.config.get(k)
+
+        # Check that the required parameter is passed. If not, error out.
+        for k in [PROVIDER_API_KEY]:
```
dustymabe: i see now where we are checking that only provider api key is set.. and it is the only one required.. we need to have some coordination between `_from_required_params` and `_check_required_params` since they both "check required params"

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,243 @@ class KubernetesProvider(Provider):
...
-        :raises: ProviderFailedException
+        if 'labels' not in data['metadata']:
+            data['metadata']['labels'] = {'namespace': self.namespace}
+        else:
+            data['metadata']['labels']['namespace'] = self.namespace
+
+        self.k8s_artifacts[kind].append(data)
+
+    '''
+    This is DEPRECATED and not needed anymore as we check the /resource URL of the kubernetes api against the artifact
```
dustymabe: we'll eventually remove this - but thanks for leaving it in.. 

can you elaborate? what code are we doing now that we weren't in the past?  i.e. where do we "check the /resource URL" and why does doing that make this no longer needed. 

@tkral may be interested in this

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,243 @@ class KubernetesProvider(Provider):
...
-
-        for kind, artifact in self.k8s_manifests:
-            if not artifact:
-                continue
-
-            path = os.path.join(self.path, artifact)
-
-            if kind in ["ReplicationController", "rc", "replicationcontroller"]:
-                self._scale_replicas(path, replicas=0)
+        logger.info("Deploying to Kubernetes")
```
dustymabe: `Undeploying from`

atomicapp/providers/kubernetes.py
cdrage: @dustymabe this checks to see if any CLI params or answers have been passed. does that make sense? so when you pass `--provider-auth --provider-api --provider-cert, etc.` it will create a generate ~/.kube/config using that information. Similar to how the configuration is generated with the `openshift.py` provider.

dustymabe:
I notice we removed persistent storage code - with the 'rescoping' of nulecule does it make sense that we don't try to do persistent storage in the future?

atomicapp/providers/kubernetes.py
cdrage: ACK

atomicapp/providers/kubernetes.py
cdrage: ACK

atomicapp/providers/lib/kubeshift/client.py
```diff
@@ -0,0 +1,57 @@
...
+        self.provider = provider
+
+        # Choose the type of provider that is being used. Error out if it is not available
+        if provider is "kubernetes":
+            self.connection = KubeKubernetesClient(config)
+            logger.debug("Using Kubernetes Provider KubeClient library")
+        else:
+            raise KubeClientError("No provider by that name.")
+
+    # Create an object using it's respective API
```
dustymabe: s/it's/its

atomicapp/providers/lib/kubeshift/client.py
```diff
@@ -0,0 +1,57 @@
...
+    def create(self, obj, namespace="default"):
+        self.connection.create(obj, namespace)
+
+    # Delete an object using it's respective API
```
dustymabe: s/it's/its

atomicapp/providers/lib/kubeshift/kubebase.py
```diff
@@ -0,0 +1,310 @@
...
+        # Check to see if verification has been disabled, if it has
+        # disable tls-verification
+        if "insecure-skip-tls-verify" in self.cluster:
+            # If it's 'true' within .kube/config
+            if self.cluster["insecure-skip-tls-verify"]:
+                connection.verify = False
+                # Disable the 'InsecureRequestWarning' notifications.
+                # As per: https://github.com/kennethreitz/requests/issues/2214
+                # Instead make a large one-time noticable warning instead
+                requests.packages.urllib3.disable_warnings()
```
dustymabe: +1

atomicapp/providers/lib/kubeshift/kubebase.py
```diff
@@ -0,0 +1,310 @@
...
+                logger.warning("CAUTION: TLS verification has been DISABLED")
```
dustymabe: ++1

dustymabe:
With @rtnpro's Config PR I wonder if we shouldn't try to disambiguate a bit with our use of the `config` word within this PR. I wonder if we should use `kconfig` instead. Or something similar.

atomicapp/providers/kubernetes.py
```diff
@@ -38,227 +44,243 @@ class KubernetesProvider(Provider):
...
+        result = {PROVIDER_API_KEY: self.provider_api,
+                  PROVIDER_AUTH_KEY: self.provider_auth,
+                  PROVIDER_TLS_VERIFY_KEY: self.provider_tls_verify,
+                  PROVIDER_CA_KEY: self.provider_ca}
+
+        # Get values from the loaded answers.conf / passed CLI params
+        for k in result.keys():
+            result[k] = self.config.get(k)
+
+        if bool([a for a in result.values() if a is None]):
```
dustymabe: Too fancy.. could this be an alternative:

```
>>> values = [False, False, "foo"]
>>> if None in values:
...     print "noneinvalues"
... 
>>> values = [False, False, None]
>>> if None in values:
...     print "noneinvalues"
... 
noneinvalues
```

tests/units/providers/test_kubernetes_provider.py
```diff
@@ -53,7 +53,9 @@ def prepare_provider(self, data):
         return provider
 
     # Check that the provider configuration file exists
-    @mock.patch.object(KubernetesProvider, '_call', mock_provider_call)
+    # We no longer use the mock call as _call no longer exists within the
+    # kubernetes provider. Thus we test the configuration exists
+    # @mock.patch.object(KubernetesProvider, '_call', mock_provider_call)
```
dustymabe: then we don't need mock_provider_call function either.

dustymabe:
We moved `kubeconfig.py` so let's update the openshift provider:

```
  File "/opt/atomicapp/atomicapp/providers/openshift.py", line 42, in <module>
    from atomicapp.providers.lib.kubeconfig import KubeConfig
ImportError: No module named kubeconfig
```

atomicapp/providers/kubernetes.py
cdrage: ACK and updated

atomicapp/providers/kubernetes.py
cdrage: It should, ACK.

atomicapp/providers/kubernetes.py
cdrage: But I like fancy! ACK xD

atomicapp/providers/kubernetes.py
cdrage: This is taken from the code that was used in `providers/openshift.py` in regards to checking the CLI params. I can change this to an if statement instead.

What are your thoughts? Refactor it into another function?

cdrage:
Hey @dustymabe 
I updated the code again with your comments.

Could we keep the 'fancy' code? Seems a lot cleaner.

Yes in regards to the persistent storage. Since the scope has changed and the fact that persistent storage was not used (as much as we hoped) I purposely left it out.

atomicapp/providers/kubernetes.py
dustymabe: I just don't like it when the alternative (much easier to read) version right there for the taking :) 

I was only making a suggestion so we can leave it like it is, but do you agree that one is easier to read than the other?

atomicapp/providers/kubernetes.py
dustymabe: yeah somehow consolidate the two.. here is one way it could be done: 
[foofile.txt](https://github.com/projectatomic/atomicapp/files/285318/foofile.txt)

dustymabe:
> Could we keep the 'fancy' code? Seems a lot cleaner.
> 
I commented on that in  the other comment. The short answer is yes. 

> Yes in regards to the persistent storage. Since the scope has changed and the fact that persistent storage was not used (as much as we hoped) I purposely left it out.

We need to spell this out more explicitly - i.e. separate PR to remove the storage stuff and an announcement. Let's make sure we think through this first as well. Leave the PR as it is right now with respect to the storage stuff though.

dustymabe:
> With @rtnpro's Config PR I wonder if we shouldn't try to disambiguate a bit with our use of the config word within this PR. I wonder if we should use kconfig instead. Or something similar.

What did you think of this comment ^^ ?

cdrage:
I agree in regards to the config terminology. It'll be nice to separate the two. Have it use `kconfig` here instead.

I'll also update this with the combine of the two functions as you pointed out in foo.txt

dustymabe:
> I'll also update this with the combine of the two functions as you pointed out in foo.txt

you like the name ?? :)

cdrage:
Hey @dustymabe thank you for being so up-to-date with the reviewing.

So that's it for now in terms of code! I've updated based on your comments and added a few entries for logger.debug output in the `kubernetes.py` provider.

I'm going to do some last-minute testing (in terms of ssl gen + using certs) to make sure it's all 100%.

cdrage:
#dotests

dustymabe:
There is a `atomicapp/providers/.kubernetes.py.swn` file in here - not sure where it came from but needs to be removed.

atomicapp/providers/kubernetes.py
dustymabe: Can you remove this function in a 2nd commit?

dustymabe: Would still like @tkral's opinion.

cdrage:
@dustymabe ewwwww. WHY VIM WHY. Removed.

atomicapp/providers/lib/kubernetes.py
dustymabe: This has not yet been done.

dustymabe:
> I agree in regards to the config terminology. It'll be nice to separate the two. Have it use kconfig here instead.

I don't think you updated this

cdrage:
@dustymabe @kadel @rtnpro 

I just tested against https://github.com/pires/kubernetes-vagrant-coreos-cluster and it seems to work correctly :+1: could you guys test against your environments for any errors / issues?

cdrage:
@dustymabe sorry, was a bit confused on what you meant. is this regarding any of the references to `self.config`?  Since that inherits from the `config` of the Provider class?

So the best way to separate the two would be to simply do `self.kconfig = self.config` on startup and use kconfig?

dustymabe:
> @dustymabe sorry, was a bit confused on what you meant. is this regarding any of the references to self.config? Since that inherits from the config of the Provider class?
> 
> So the best way to separate the two would be to simply do self.kconfig = self.config on startup and use kconfig?

```
config (object): An object of the .kube/config configuration
```

I think anything that is referencing an object of the `.kube/config` configuration should be known as kconfig. This would be an attempt to disambiguate between kconfig and config (the class that rtnpro is writing).

cdrage:
@dustymabe ahhhh, that's what you meant! I was getting confused between @rtnpro 's refactor and `.kube/config`, I'll change the code to reflect that :)

kadel:
I tested this on [Kubernetes running on AWS](http://kubernetes.io/docs/getting-started-guides/aws/)

When authenticating using token, everything works fine.

When using `client-certificate`, `client-key` it fails
```
▶ atomicapp-dev stop . --provider=kubernetes --provider-config=/home/tomas/.kube/config --namespace=default --verbose                           
atomicapp-dev:1: command not found:  
INFO   :: - cli/main.py ::    _  _             _      _
INFO   :: - cli/main.py ::   /_\| |_ ___ _ __ (_)__  /_\  _ __ _ __   Version:  0.5.2
INFO   :: - cli/main.py ::  / _ \  _/ _ \ '  \| / _|/ _ \| '_ \ '_ \  Nulecule: 0.0.2
INFO   :: - cli/main.py :: /_/ \_\__\___/_|_|_|_\__/_/ \_\ .__/ .__/  Mode:     Stop
INFO   :: - cli/main.py ::                                |_|  |_|
DEBUG  :: - cli/main.py :: Final parsed cmdline: stop . --provider=kubernetes --provider-config=/home/tomas/.kube/config --namespace=default --verbose
DEBUG  :: - nulecule/main.py :: NuleculeManager init app_path: .
DEBUG  :: - nulecule/main.py :: NuleculeManager init image: None
DEBUG  :: - utils.py :: Loading answers from file: ./answers.conf
DEBUG  :: - utils.py :: Loading answers from file: ./answers.conf.gen
DEBUG  :: - providers/kubernetes.py :: Given config: {u'provider-tlsverify': False, u'provider-config': u'/home/tomas/.kube/config', u'provider': u'kubernetes', u'image': u'centos/httpd', u'hostport': 80, u'namespace': u'default'}
INFO   :: - providers/kubernetes.py :: Using namespace default
DEBUG  :: - providers/kubernetes.py :: Processing artifact: artifacts/kubernetes/.hello-apache-pod.json
DEBUG  :: - utils.py :: Finding the users home directory
DEBUG  :: - utils.py :: Running as user tomas. Using home directory /home/tomas for configuration data
DEBUG  :: - providers/kubernetes.py :: Provider configuration provided
ERROR  :: - cli/main.py :: 'token'
  File "/home/tomas/dev/github/kadel/atomicapp/atomicapp/cli/main.py", line 130, in cli_func_exec
    cli_func(cli_func_args)
  File "/home/tomas/dev/github/kadel/atomicapp/atomicapp/cli/main.py", line 97, in cli_stop
    nm.stop(**argdict)
  File "/home/tomas/dev/github/kadel/atomicapp/atomicapp/nulecule/main.py", line 349, in stop
    self.nulecule.stop(cli_provider, dryrun)
  File "/home/tomas/dev/github/kadel/atomicapp/atomicapp/nulecule/base.py", line 223, in stop
    component.stop(provider_key, dryrun)
  File "/home/tomas/dev/github/kadel/atomicapp/atomicapp/nulecule/base.py", line 351, in stop
    provider.init()
  File "/home/tomas/dev/github/kadel/atomicapp/atomicapp/providers/kubernetes.py", line 101, in init
    self.api = Client(KubeConfig.from_file(self.config_file), "kubernetes")
  File "/home/tomas/dev/github/kadel/atomicapp/atomicapp/providers/lib/kubeshift/client.py", line 42, in __init__
    self.connection = KubeKubernetesClient(config)
  File "/home/tomas/dev/github/kadel/atomicapp/atomicapp/providers/lib/kubeshift/kubernetes.py", line 41, in __init__
    self.api = KubeBase(config)
  File "/home/tomas/dev/github/kadel/atomicapp/atomicapp/providers/lib/kubeshift/kubebase.py", line 69, in __init__
    self.token = self.user['token']
KeyError: 'token'
```

cdrage:
@kadel 
Could you post your .kube/config (with your API credentials balnked out and whatever) I think this is an edge-case with user being set with no token available.

kadel:
@cdrage I don't think that this is edge case, this is how authentication for `system:admin` user in OpenShift is setup by default (using  `client-certificate-data` and `client-key-data`).

If you have config with both client-certificate/key and token for one user, you have to decide which one to use, it doesn't make sense to use both, even if in this case it is technically possible.

Examples of `.kube/config` with  different auth methods:
- client-{certificate,key} - https://gist.github.com/kadel/b2d3df8962a2d2feabc1c5b29de82b55
- client-{certificate,key}-data - https://gist.github.com/kadel/29df1fee245627ed90e53b44cdb7c2e7
- token - https://gist.github.com/kadel/1089d0906a997da017f5a4bc4f39bb0e
- username:password (BasicAuth) - https://gist.github.com/kadel/2a5c4bb74ab733688ab1eaf4976080b7

cdrage:
@kadel thank you for posting them! I'm going to go through each-one and figure out how we can work this in.

cdrage:
@kadel how did we do `certificate-authority-data` in the openshift.py provider? or was this intended to be added in the cert commit for openshift?

kadel:
@cdrage yes it was part of old PR with cert auth. Here is example how we handled that https://github.com/kadel/atomicapp/blob/bdc3a2bd3cc953ea1436392367a3f92a1b532b8d/atomicapp/providers/lib/kubeconfig.py#L116,L119
Temporray files was becuase requests library accepts certs only as files, or I didn't figure out how to do this differently

atomicapp/providers/lib/kubeshift/kubebase.py
```diff
@@ -0,0 +1,331 @@
...
+    def cert_file(data):
+        '''
+        Some certificate .kube/config components are required to be a filename.
+
+        Returns either the filename or a tmp location of said data in a file.
+        All certificates used with Kubernetes are base64 encoded and thus need to be decoded
+        '''
+
+        # If it starts with /, we assume it's a filename so we just return that.
+        if data.startswith('/'):
```
kadel: it can be related path to `.kube/config` and than it is not starting with `/`

atomicapp/providers/lib/kubeshift/kubebase.py
```diff
@@ -0,0 +1,331 @@
...
+            return data
+
+        # If it's data, we assume it's a base64 certificate and we decode and write to a tmp file
+        else:
+            with tempfile.NamedTemporaryFile(delete=False) as f:
+                f.write(base64.b64decode(data))
```
kadel: certificates and keys in files are not base64 encoded

cdrage: Thanks for this @kadel :) I've updated the PR to reflect what parts should be decoded and what shouldn't.

I'll have to add the other `-data` keys in a separate PR

cdrage:
@kadel updated the PR with your feedback on the `-data` files. In particular, it's in `kubebase.py` in relation to the `cert_file` declarations around the requests library.

kadel:
Sorry, but It is not working like this :-(
You are still expecting that under one key in `.kube/config` can be either base64 encoded string or filename. But this is not right. 
For example CA certificate can be in `certificate-authority` (than it is path to file) or in `certificate-authority-data` than it is base64 encoded string.

kadel:
There are still some missing features to properly handle Kubernetes config file.
But we can add those in another PR.
 - support for relative paths to certificates and keys
 - support for *-data keys (`certificate-authority-data`, `client-certificate-data`, `key-certificate-data`)

:+1:  LGTM
Awesome work @cdrage :sparkles:

cdrage:
@kadel 

Thanks man! :dancers: 

I'll add the `-data` as a separate PR as well as the support for relative paths (will have to use the `utils.py` functions we had before to find the proper path, due to /host being passed, etc.)

cdrage:
@dustymabe I think this may be good to go! I've updated kubebase to reflect the wording change on self.config to self.kubeconfig

This may be ready for mergin' :+1:

cdrage:
#dotests

dustymabe:
@cdrage - I haven't looked at this but from what I reviewed a week ago everything looked good. Since Tomas gave +1 I'm +1. Merge if you can get all tests passing.

cdrage:
WOOO! Kubernetes tests pass! :tada: 

Merging away :)

cdrage:
Thanks everyone for helping me out with this! It's been exciting to finally get this in.

dustymabe:
haha openshift tests failed - thoughts about that?

cdrage:
@dustymabe false alarm, tests passed locally for OpenShift on my system :)

Seems the previous issue is occuring again with the latest update to OpenShift:
```

ERROR  :: - cli/main.py :: 403 {u'status': u'Failure', u'kind': u'Status', u'code': 403, u'apiVersion': u'v1', u'reason': u'Forbidden', u'details': {u'kind': u'pods'}, u'message': u'User "system:anonymous" cannot create pods in project "foo"', u'metadata': {}}
```

cdrage:
Let me test one more thing ^^ updated my OpenShift image and it failed.

cdrage:
@dustymabe should be fine, no need to revert or anything. i knew I didn't touch the openshift integration (didn't change any underlying code)

I've opened up https://github.com/projectatomic/adb-tests/pull/33 to fix this.

cdrage:
#dotests

atomicapp/providers/lib/kubeshift/kubernetes.py
```diff
@@ -0,0 +1,171 @@
...
+        # Pass in the configuration data (.kube/config object) to the KubeBase
+        self.api = KubeBase(config)
+
+        # Check the API url
+        url = self.api.cluster['server']
+        if not re.match('(?:http|https)://', url):
+            raise KubeKubernetesError("Kubernetes API URL does not include HTTP or HTTPS")
+
+        # Gather what end-points we will be using
+        self.k8s_api = urljoin(url, "api/v1/")
```
khrisrichardson: Are there plans to support the other groups of APIS accessible via the /apis endpoint (e.g. - apps, autoscaling, batch, extensions, policy, etc.)?

cdrage: @khrisrichardson We hope so. Any changes to https://github.com/projectatomic/nulecule will reflect in atomic app in terms of providing support for autoscaling, batch commands, extensions, policies, etc. If the spec changes we'll reflect that as features in Atomic App :)

khrisrichardson: Thanks, @cdrage 

Is there a timeline in mind, because by restricting to api/v1, the behavior may be a regression for some?

```
ERROR  :: No kind by that name: Deployment
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.5.2-py2.7.egg/atomicapp/cli/main.py", line 130, in cli_func_exec
    cli_func(cli_func_args)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.5.2-py2.7.egg/atomicapp/cli/main.py", line 84, in cli_run
    nm.run(**argdict)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.5.2-py2.7.egg/atomicapp/nulecule/main.py", line 320, in run
    nulecule.run(cli_provider, dryrun)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.5.2-py2.7.egg/atomicapp/nulecule/base.py", line 224, in run
    provider.run()
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.5.2-py2.7.egg/atomicapp/providers/kubernetes.py", line 268, in run
    self.api.create(artifact, self.namespace)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.5.2-py2.7.egg/atomicapp/providers/lib/kubeshift/client.py", line 49, in create
    self.connection.create(obj, namespace)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.5.2-py2.7.egg/atomicapp/providers/lib/kubeshift/kubernetes.py", line 62, in create
    kind, url = self._generate_kurl(obj, namespace)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.5.2-py2.7.egg/atomicapp/providers/lib/kubeshift/kubernetes.py", line 142, in _generate_kurl
    raise KubeKubernetesError("No kind by that name: %s" % kind)
KubeKubernetesError: No kind by that name: Deployment
```

I was testing kubernetes v1.3.0-alpha.5, and prior to pulling this commit, I was able to create objects belonging to the extensions/v1beta1 API, such as DaemonSet and Deployment.

cdrage: @khrisrichardson Ah! That's what you meant. I'm going to open up an issue for this.

khrisrichardson: I have a patch to support API groups. Will fork and create PR.

PR #715, opened 2016-04-28
Updated the inContainer function doc string

surajssd:
Since the last few changes in the code now `.dockerenv` and `.dockerinit` are not checked in the code, but the same thing was not updated in the doc, so updated the stale doc string.


cdrage:
Hahah woops! Forgot to update it! Thanks! :)

PR #717, opened 2016-04-29
Removed unused function update from utils.py

surajssd:
This function is not used anywhere in the code.


cdrage:
#dotests

cdrage:
tests pass and from grep'ing this isn't used anymore.

Code LGTM

kadel:
LGTM

PR #719, opened 2016-05-04
Change provider-config warning to debug

cdrage:
This should be within the debug log rather than warn the user.
Basically, if the user has not provided --provider-config Atomic App
will warn. This should instead be put in the debugging output to let the
developer know that --provider-config hasn't been passed via
answers.conf or the CLI.


cdrage:
ping @rtnpro @surajssd @kadel for review

kadel:
I agree, that warning was sometimes confusing.
LGTM

cdrage:
Going to merge this :) If anyone has any objections please feel free to comment on here / revert the merge!

PR #720, opened 2016-05-05
Refactor Nulecule config.

### LINKED ISSUE #524: Create a config class to manage config data
As mentioned here: https://github.com/projectatomic/atomicapp/pull/522#issuecomment-172894076, management of config data is becoming convoluted using a dictionary. It'd be best if we create a ``Config`` class which abstracts all operations (loading, resolving of config data) on config data and provides a clean interface to interact with the config data.

rtnpro:
Fixes #524

### DIFF
--- a/atomicapp/cli/main.py
+++ b/atomicapp/cli/main.py
@@ -64,7 +64,8 @@ def cli_fetch(args):
     nm = NuleculeManager(app_spec=argdict['app_spec'],
                          destination=destination,
                          cli_answers=argdict['cli_answers'],
-                         answers_file=argdict['answers'])
+                         answers_file=argdict['answers'],
+                         answers_format=argdict.get('answers_format'))
     nm.fetch(**argdict)
     # Clean up the files if the user asked us to. Otherwise
     # notify the user where they can manage the application
@@ -81,7 +82,8 @@ def cli_run(args):
     nm = NuleculeManager(app_spec=argdict['app_spec'],
                          destination=destination,
                          cli_answers=argdict['cli_answers'],
-                         answers_file=argdict['answers'])
+                         answers_file=argdict['answers'],
+                         answers_format=argdict.get('answers_format'))
     nm.run(**argdict)
     # Clean up the files if the user asked us to. Otherwise
     # notify the user where they can manage the application
@@ -306,7 +308,7 @@ class CLI():
             help="A file which will contain anwsers provided in interactive mode")
         run_subparser.add_argument(
             "--provider",
-            dest="cli_provider",
+            dest="provider",
             choices=PROVIDERS,
             help="The provider to use. Overrides provider value in answerfile.")
         run_subparser.add_argument(
@@ -511,7 +513,8 @@ class CLI():
         # and make a dictionary of it to pass along in args.
         setattr(args, 'cli_answers', {})
         for item in ['provider-api', 'provider-cafile', 'provider-auth',
-                     'provider-config', 'provider-tlsverify', 'namespace']:
+                     'provider-config', 'provider-tlsverify', 'namespace',
+                     'provider']:
             if hasattr(args, item) and getattr(args, item) is not None:
                 args.cli_answers[item] = getattr(args, item)
 
--- a/atomicapp/nulecule/base.py
+++ b/atomicapp/nulecule/base.py
@@ -18,7 +18,6 @@
  along with Atomic App. If not, see <http://www.gnu.org/licenses/>.
 """
 import anymarkup
-import copy
 import logging
 import os
 import yaml
@@ -38,7 +37,7 @@ from atomicapp.constants import (APP_ENT_PATH,
                                  NAME_KEY,
                                  INHERIT_KEY,
                                  ARTIFACTS_KEY,
-                                 DEFAULT_PROVIDER)
+                                 NAMESPACE_SEPARATOR)
 from atomicapp.utils import Utils
 from atomicapp.requirements import Requirements
 from atomicapp.nulecule.lib import NuleculeBase
@@ -76,7 +75,7 @@ class Nulecule(NuleculeBase):
             metadata (dict): Nulecule metadata
             requirements (dict): Requirements for the Nulecule application
             params (list): List of params for the Nulecule application
-            config (dict): Config data for the Nulecule application
+            config (atomicapp.nulecule.config.Config): Config data
             namespace (str): Namespace of the current Nulecule application
 
         Returns:
@@ -88,7 +87,7 @@ class Nulecule(NuleculeBase):
         self.metadata = metadata or {}
         self.graph = graph
         self.requirements = requirements
-        self.config = config or {}
+        self.config = config
 
     @classmethod
     def unpack(cls, image, dest, config=None, namespace=GLOBAL_CONF,
@@ -101,7 +100,7 @@ class Nulecule(NuleculeBase):
             image (str): A Docker image name.
             dest (str): Destination path where Nulecule data from Docker
                         image should be extracted.
-            config (dict): Dictionary, config data for Nulecule application.
+            config: An instance of atomicapp.nulecule.config.Config
             namespace (str): Namespace for Nulecule application.
             nodeps (bool): Don't pull external Nulecule dependencies when
                            True.
@@ -115,7 +114,7 @@ class Nulecule(NuleculeBase):
         if Utils.running_on_openshift():
             # pass general config data containing provider specific data
             # to Openshift provider
-            op = OpenshiftProvider(config.get('general', {}), './', False)
+            op = OpenshiftProvider(config.globals, './', False)
             op.artifacts = []
             op.init()
             op.extract(image, APP_ENT_PATH, dest, update)
@@ -138,7 +137,8 @@ class Nulecule(NuleculeBase):
 
         Args:
             src (str): Path to load Nulecule application from.
-            config (dict): Config data for Nulecule application.
+            config (atomicapp.nulecule.config.Config): Config data for
+                Nulecule application.
             namespace (str): Namespace for Nulecule application.
             nodeps (bool): Do not pull external applications if True.
             dryrun (bool): Do not make any change to underlying host.
@@ -176,8 +176,9 @@ class Nulecule(NuleculeBase):
             raise NuleculeException("Failure parsing %s file. Validation error on line %s, column %s:\n%s"
                                     % (nulecule_path, line, column, output))
 
-        nulecule = Nulecule(config=config, basepath=src,
-                            namespace=namespace, **nulecule_data)
+        nulecule = Nulecule(config=config,
+                            basepath=src, namespace=namespace,
+                            **nulecule_data)
         nulecule.load_components(nodeps, dryrun)
         return nulecule
 
@@ -231,25 +232,23 @@ class Nulecule(NuleculeBase):
         It updates self.config.
 
         Args:
-            config (dict): Existing config data, may be from ANSWERS
-                           file or any other source.
+            config (atomicapp.nulecule.config.Config): Existing config data,
+                may be from ANSWERS file or any other source.
 
         Returns:
             None
         """
+        if config is None:
+            config = self.config
         super(Nulecule, self).load_config(
             config=config, ask=ask, skip_asking=skip_asking)
-        if self.namespace == GLOBAL_CONF and self.config[GLOBAL_CONF].get('provider') is None:
-            self.config[GLOBAL_CONF]['provider'] = DEFAULT_PROVIDER
-            logger.info("Provider not specified, using default provider - {}".
-                        format(DEFAULT_PROVIDER))
+
         for component in self.components:
             # FIXME: Find a better way to expose config data to components.
             #        A component should not get access to all the variables,
             #        but only to variables it needs.
-            component.load_config(config=copy.deepcopy(self.config),
+            component.load_config(config=config.clone(component.namespace),
                                   ask=ask, skip_asking=skip_asking)
-            self.merge_config(self.config, component.config)
 
     def load_components(self, nodeps=False, dryrun=False):
         """
@@ -270,8 +269,8 @@ class Nulecule(NuleculeBase):
             node_name = node[NAME_KEY]
             source = Utils.getSourceImage(node)
             component = NuleculeComponent(
-                node_name, self.basepath, source,
-                node.get(PARAMS_KEY), node.get(ARTIFACTS_KEY),
+                self._get_component_namespace(node_name), self.basepath,
+                source, node.get(PARAMS_KEY), node.get(ARTIFACTS_KEY),
                 self.config)
             component.load(nodeps, dryrun)
             components.append(component)
@@ -293,6 +292,12 @@ class Nulecule(NuleculeBase):
         """
         for component in self.components:
             component.render(provider_key=provider_key, dryrun=dryrun)
+
+    def _get_component_namespace(self, component_name):
+        current_namespace = '' if self.namespace == GLOBAL_CONF else self.namespace
+        return (
+            '%s%s%s' % (current_namespace, NAMESPACE_SEPARATOR, component_name)
+            if current_namespace else component_name)
 
 
 class NuleculeComponent(NuleculeBase):
@@ -356,12 +361,13 @@ class NuleculeComponent(NuleculeBase):
         """
         Load config for the Nulecule component.
         """
+        if config is None:
+            config = self.config
         super(NuleculeComponent, self).load_config(
             config, ask=ask, skip_asking=skip_asking)
         if isinstance(self._app, Nulecule):
-            self._app.load_config(config=copy.deepcopy(self.config),
+            self._app.load_config(config=self.config.clone(self.namespace),
                                   ask=ask, skip_asking=skip_asking)
-            self.merge_config(self.config, self._app.config)
 
     def load_external_application(self, dryrun=False, update=False):
         """
@@ -384,7 +390,8 @@ class NuleculeComponent(NuleculeBase):
                 'Found existing external application: %s '
                 'Loading: ' % self.name)
             nulecule = Nulecule.load_from_path(
-                external_app_path, dryrun=dryrun, update=update)
+                external_app_path, dryrun=dryrun, update=update,
+                namespace=self.namespace)
         elif not dryrun:
             logger.info('Pulling external application: %s' % self.name)
             nulecule = Nulecule.unpack(
--- a/atomicapp/nulecule/base.py
+++ b/atomicapp/nulecule/base.py
@@ -176,9 +176,8 @@ class Nulecule(NuleculeBase):
             raise NuleculeException("Failure parsing %s file. Validation error on line %s, column %s:\n%s"
                                     % (nulecule_path, line, column, output))
 
-        nulecule = Nulecule(config=config,
-                            basepath=src, namespace=namespace,
-                            **nulecule_data)
+        nulecule = Nulecule(config=config, basepath=src,
+                            namespace=namespace, **nulecule_data)
         nulecule.load_components(nodeps, dryrun)
         return nulecule
 
@@ -247,7 +246,7 @@ class Nulecule(NuleculeBase):
             # FIXME: Find a better way to expose config data to components.
             #        A component should not get access to all the variables,
             #        but only to variables it needs.
-            component.load_config(config=config.clone(component.namespace),
+            component.load_config(config=config,
                                   ask=ask, skip_asking=skip_asking)
 
     def load_components(self, nodeps=False, dryrun=False):
@@ -294,6 +293,18 @@ class Nulecule(NuleculeBase):
             component.render(provider_key=provider_key, dryrun=dryrun)
 
     def _get_component_namespace(self, component_name):
+        """
+        Get a unique namespace for a Nulecule graph item, by concatinating
+        the namespace of the current Nulecule (which could be the root Nulecule
+        app or a child or external Nulecule app) and name of the Nulecule
+        graph item.
+
+        Args:
+            component_name (str): Name of the Nulecule graph item
+
+        Returns:
+            A string
+        """
         current_namespace = '' if self.namespace == GLOBAL_CONF else self.namespace
         return (
             '%s%s%s' % (current_namespace, NAMESPACE_SEPARATOR, component_name)
@@ -366,7 +377,7 @@ class NuleculeComponent(NuleculeBase):
         super(NuleculeComponent, self).load_config(
             config, ask=ask, skip_asking=skip_asking)
         if isinstance(self._app, Nulecule):
-            self._app.load_config(config=self.config.clone(self.namespace),
+            self._app.load_config(config=self.config,
                                   ask=ask, skip_asking=skip_asking)
 
     def load_external_application(self, dryrun=False, update=False):
@@ -443,7 +454,7 @@ class NuleculeComponent(NuleculeBase):
             raise NuleculeException(
                 "Data for provider \"%s\" are not part of this app"
                 % provider_key)
-        context = self.get_context()
+        context = self.config.context(self.namespace)
         for provider in self.artifacts:
             if provider_key and provider != provider_key:
                 continue
--- a/atomicapp/nulecule/config.py
+++ b/atomicapp/nulecule/config.py
@@ -4,8 +4,7 @@ import logging
 from atomicapp.constants import (GLOBAL_CONF,
                                  LOGGER_COCKPIT,
                                  DEFAULT_PROVIDER,
-                                 DEFAULT_ANSWERS,
-                                 NAMESPACE_SEPARATOR)
+                                 DEFAULT_ANSWERS)
 from collections import defaultdict
 
 cockpit_logger = logging.getLogger(LOGGER_COCKPIT)
@@ -13,121 +12,125 @@ cockpit_logger = logging.getLogger(LOGGER_COCKPIT)
 
 class Config(object):
     """
-    Store config data for a Nulecule or Nulecule component.
+    This class allows to store config data in different scopes along with
+    source info for the data. When fetching the value for a key in a scope,
+    the source info and the PRIORITY order of sources is taken into account.
 
-    It stores config data from different sources (answers, cli, user data)
-    separately, and exposes high level interfaces to read, write data
-    from it, with enforced read/write policies.
+    Data sources:
+        cli: Config data coming from the CLI
+        runtime: Config data resolved during atomic app runtime. For example,
+            when the value for a parameter in a Nulecule or Nulecule graph
+            item is missing in answers data, we first try to load the default
+            value for the parameter. When there's no default value, or when
+            the user has specified to forcefully ask the user for values, we
+            ask the user for data. These data collected/resolved during runtime
+            form the runtime data.
+        answers: Config data coming from answers file
+        defaults: Default config data specified in atomicapp/constants.py
+
+    The priority order of the data sources is:
+    cli > runtime > answers > defaults
     """
 
-    def __init__(self, namespace='', answers=None, cli=None,
-                 data=None, is_nulecule=False):
-        self._namespace = namespace
-        self._is_nulecule = is_nulecule or False
-        self._parent_ns, self._current_ns = self._split_namespace(self._namespace)
-        # Store answers data
-        self._answers = defaultdict(dict)
-        self._answers.update(answers or {})
-        # Store CLI data
-        self._cli = cli or {}
-        # Store data collected during runtime
-        self._data = data or defaultdict(dict)
-        self._context = None
-        self._provider = None
+    PRIORITY = (
+        'cli',
+        'runtime',
+        'answers',
+        'defaults'
+    )
 
-    @property
-    def globals(self):
+    def __init__(self, answers=None, cli=None):
         """
-        Get global config params dict for a Nulecule.
-        """
-        d = self._answers.get(GLOBAL_CONF, {})
-        d.update(self._data.get(GLOBAL_CONF, {}))
-        d.update(self._cli.get(GLOBAL_CONF, {}))
-        return d
+        Initialize a Config instance.
 
-    @property
-    def provider(self):
+        Args:
+            answers (dict): Answers data
+            cli (dict): CLI data
         """
-        Get provider name.
+        answers = answers or {}
+        cli = cli or {}
+        # We use a defaultdict of defaultdicts so that we can avoid doing
+        # redundant checks in a nested dictionary if the value of the keys
+        # are dictionaries or None.
+        self._data = defaultdict(defaultdict)
+        # Initialize default data dict
+        self._data['defaults'] = defaultdict(defaultdict)
+        # Initialize answers data dict
+        self._data['answers'] = defaultdict(defaultdict)
+        # Initialize cli data dict
+        self._data['cli'] = defaultdict(defaultdict)
+        # Initialize runtime data dict
+        self._data['runtime'] = defaultdict(defaultdict)
 
-        Returns:
-            Provider name (str)
-        """
-        if self._provider is None:
-            self._provider = self._data[GLOBAL_CONF].get('provider') or \
-                self._answers[GLOBAL_CONF].get('provider')
-            if self._provider is None:
-                self._data[GLOBAL_CONF]['provider'] = DEFAULT_PROVIDER
-                self._provider = DEFAULT_PROVIDER
+        # Load default answers
+        for scope, data in DEFAULT_ANSWERS.items():
+            for key, value in data.items():
+                self.set(key, value, scope=scope, source='defaults')
+        self.set('provider', DEFAULT_PROVIDER, scope=GLOBAL_CONF, source='defaults')
 
-        return self._provider
+        # Load answers data
+        for scope, data in answers.items():
+            for key, value in data.items():
+                self.set(key, value, scope=scope, source='answers')
 
-    @property
-    def providerconfig(self):
-        """
-        Get provider config info taking into account answers and cli data.
-        """
-        pass
+        # Load cli data
+        for key, value in cli.items():
+            self.set(key, value, scope=GLOBAL_CONF, source='cli')
 
-    @property
-    def namespace(self):
+    def get(self, key, scope=GLOBAL_CONF, ignore_sources=[]):
         """
-        Get normalized namespace for this instance.
-
-        Returns:
-            Current namespace (str).
-        """
-        return self._namespace or GLOBAL_CONF
-
-    def set(self, key, value):
-        """
-        Set value for a key in the current namespace.
+        Get the value of a key in a scope. This takes care of resolving
+        the value by going through the PRIORITY order of the various
+        sources of data.
 
         Args:
             key (str): Key
-            value (str): Value.
-        """
-        self._data[self.namespace][key] = value
+            scope (str): Scope from which to fetch the value for the key
 
-    def get(self, key):
+        Returns:
+            Value for the key.
         """
-        Get value for a key from data accessible from the current namespace.
+        for source in self.PRIORITY:
+            if source in ignore_sources:
+                continue
+            value = self._data[source][scope].get(key) or self._data[source][
+                GLOBAL_CONF].get(key)
+            if value:
+                return value
+        return None
 
-        TODO: Improved data inheritance model. It makes sense for a component
-        to be able to access data from it's sibling namespaces and children
-        namespaces.
+    def set(self, key, value, source, scope=GLOBAL_CONF):
+        """
+        Set the value for a key within a scope along with specifying the
+        source of the value.
 
         Args:
             key (str): Key
+            value: Value
+            scope (str): Scope in which to store the value
+            source (str): Source of the value
+        """
+        self._data[source][scope][key] = value
 
+    def context(self, scope=GLOBAL_CONF):
+        """
+        Get context data for the scope of Nulecule graph item by aggregating
+        the data from various sources taking their priority order into
+        account. This context data, which is a flat dictionary, is used to
+        render the variables in the artifacts of Nulecule graph item.
+
+        Args:
+            scope (str): Scope (or namespace) for the Nulecule graph item.
         Returns:
-            Value for the key, else None.
+            A dictionary
         """
-        return (
-            self._data[self.namespace].get(key) or
-            (self._data[self._parent_ns].get(key) if self._parent_ns else None) or
-            self._data[GLOBAL_CONF].get(key) or
-            self._answers[self.namespace].get(key) or
-            (self._answers[self._parent_ns].get(key) if self._parent_ns else None) or
-            self._answers[GLOBAL_CONF].get(key)
-        )
-
-    def context(self):
-        """
-        Get context to render artifact files in a Nulecule component.
-
-        TODO: Improved data inheritance model. Data from siblings and children
-        namespaces should be available in the context to render an artifact
-        file in the current namespace.
-        """
-        if self._context is None:
-            self._context = {}
-            self._context.update(copy.copy(self._data[GLOBAL_CONF]))
-            self._context.update(copy.copy(self._data[self.namespace]))
-
-            self._context.update(copy.copy(self._answers[GLOBAL_CONF]))
-            self._context.update(copy.copy(self._answers[self.namespace]))
-        return self._context
+        result = {}
+        for source in reversed(self.PRIORITY):
+            source_data = self._data[source]
+            result.update(copy.deepcopy(source_data.get(GLOBAL_CONF) or {}))
+            if scope != GLOBAL_CONF:
+                result.update(copy.deepcopy(source_data.get(scope) or {}))
+        return result
 
     def runtime_answers(self):
         """
@@ -137,56 +140,34 @@ class Config(object):
             A defaultdict containing runtime answers data.
         """
         answers = defaultdict(dict)
-        answers.update(copy.deepcopy(DEFAULT_ANSWERS))
-        answers['general']['provider'] = self.provider
 
-        for key, value in self._answers.items():
-            answers[key].update(value)
-
-        for key, value in self._data.items():
-            answers[key].update(value)
+        for source in reversed(self.PRIORITY):
+            for scope, data in (self._data.get(source) or {}).items():
+                answers[scope].update(copy.deepcopy(data))
 
         # Remove empty sections for answers
         for key, value in answers.items():
-            if value is None:
-                answers.pop(key, None)
+            if not value:
+                answers.pop(key)
 
         return answers
 
-    def clone(self, namespace):
+    def update_source(self, source, data):
         """
-        Create a new config instance in the specified namespace.
+        Update answers data for a source.
 
         Args:
-            name (str): Name of the child component
+            source (str): Source name
+            data (dict): Answers data
+        """
+        data = data or {}
+        if source not in self._data:
+            raise
 
-        Returns:
-            A Config instance.
-        """
-        config = Config(namespace=namespace,
-                        answers=self._answers,
-                        cli=self._cli,
-                        data=self._data)
-        return config
+        # clean up source data
+        for k in self._data[source]:
+            self._data[source].pop(k)
 
-    def _split_namespace(self, namespace):
-        """
-        Split namespace to get parent and current namespace in a Nulecule.
-        """
-        if self._is_nulecule:
-            return '', namespace
-        words = namespace.rsplit(NAMESPACE_SEPARATOR, 1)
-        parent, current = '', ''
-        if len(words) == 2:
-            parent, current = words[0], words[1]
-        else:
-            parent, current = '', words[0]
-        return parent, current
-
-    def __eq__(self, obj):
-        """
-        Check equality of config instances.
-        """
-        if self._namespace == obj._namespace or self._answers == obj._answers or self._data == obj._data or self._cli == obj._cli:
-            return True
-        return False
+        for scope, data in data.items():
+            for key, value in data.items():
+                self.set(key, value, scope=scope, source=source)
--- a/atomicapp/nulecule/config.py
+++ b/atomicapp/nulecule/config.py
@@ -1,0 +1,192 @@
+import copy
+import logging
+
+from atomicapp.constants import (GLOBAL_CONF,
+                                 LOGGER_COCKPIT,
+                                 DEFAULT_PROVIDER,
+                                 DEFAULT_ANSWERS,
+                                 NAMESPACE_SEPARATOR)
+from collections import defaultdict
+
+cockpit_logger = logging.getLogger(LOGGER_COCKPIT)
+
+
+class Config(object):
+    """
+    Store config data for a Nulecule or Nulecule component.
+
+    It stores config data from different sources (answers, cli, user data)
+    separately, and exposes high level interfaces to read, write data
+    from it, with enforced read/write policies.
+    """
+
+    def __init__(self, namespace='', answers=None, cli=None,
+                 data=None, is_nulecule=False):
+        self._namespace = namespace
+        self._is_nulecule = is_nulecule or False
+        self._parent_ns, self._current_ns = self._split_namespace(self._namespace)
+        # Store answers data
+        self._answers = defaultdict(dict)
+        self._answers.update(answers or {})
+        # Store CLI data
+        self._cli = cli or {}
+        # Store data collected during runtime
+        self._data = data or defaultdict(dict)
+        self._context = None
+        self._provider = None
+
+    @property
+    def globals(self):
+        """
+        Get global config params dict for a Nulecule.
+        """
+        d = self._answers.get(GLOBAL_CONF, {})
+        d.update(self._data.get(GLOBAL_CONF, {}))
+        d.update(self._cli.get(GLOBAL_CONF, {}))
+        return d
+
+    @property
+    def provider(self):
+        """
+        Get provider name.
+
+        Returns:
+            Provider name (str)
+        """
+        if self._provider is None:
+            self._provider = self._data[GLOBAL_CONF].get('provider') or \
+                self._answers[GLOBAL_CONF].get('provider')
+            if self._provider is None:
+                self._data[GLOBAL_CONF]['provider'] = DEFAULT_PROVIDER
+                self._provider = DEFAULT_PROVIDER
+
+        return self._provider
+
+    @property
+    def providerconfig(self):
+        """
+        Get provider config info taking into account answers and cli data.
+        """
+        pass
+
+    @property
+    def namespace(self):
+        """
+        Get normalized namespace for this instance.
+
+        Returns:
+            Current namespace (str).
+        """
+        return self._namespace or GLOBAL_CONF
+
+    def set(self, key, value):
+        """
+        Set value for a key in the current namespace.
+
+        Args:
+            key (str): Key
+            value (str): Value.
+        """
+        self._data[self.namespace][key] = value
+
+    def get(self, key):
+        """
+        Get value for a key from data accessible from the current namespace.
+
+        TODO: Improved data inheritance model. It makes sense for a component
+        to be able to access data from it's sibling namespaces and children
+        namespaces.
+
+        Args:
+            key (str): Key
+
+        Returns:
+            Value for the key, else None.
+        """
+        return (
+            self._data[self.namespace].get(key) or
+            (self._data[self._parent_ns].get(key) if self._parent_ns else None) or
+            self._data[GLOBAL_CONF].get(key) or
+            self._answers[self.namespace].get(key) or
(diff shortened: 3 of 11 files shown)


atomicapp/nulecule/config.py
```diff
@@ -0,0 +1,71 @@
...
+            parent, current = '', words[0]
+        return parent, current
+
+    def context(self):
+        """
+        Get context to render artifact files in a Nulecule component.
+        """
+        context = {}
+
+        context.update(copy.copy(self._data[GLOBAL_CONF]))
```
surajssd: can we use `copy.deepcopy`, just to be on safe side?

rtnpro: It's only one level, so I did not do ``copy.deepcopy``

tests/units/nulecule/test_nulecule_component.py
```diff
@@ -136,9 +136,12 @@ def test_load_config_local_app(self):
             'general': {'a': 'b', 'key2': 'val2'},
             'some-app': {'key1': 'val1'}
         }
+        from atomicapp.nulecule.config import Config
+        config = Config(answers=initial_config)
 
         nc = NuleculeComponent('some-app', 'some/path', params=params)
-        nc.load_config(config=copy.deepcopy(initial_config))
+        nc.load_config(config=config)
+        from ipdb import set_trace; set_trace()
```
surajssd: `set_trace()`

surajssd:
@rtnpro there are some issues while using this PR

```bash
(atomic) ✔ ~/git/nulecule-library/etherpad-centos7-atomicapp [master ↑·39|✔] 
00:06 $ sudo atomicapp run . -v
1463457985 - [INFO] - cli/main.py - Action/Mode Selected is: run
1463457985 - [DEBUG] - cli/main.py - Final parsed cmdline: run . -v
1463457985 - [DEBUG] - nulecule/main.py - NuleculeManager init app_path: .
1463457985 - [DEBUG] - nulecule/main.py - NuleculeManager init image: None
1463457985 - [DEBUG] - nulecule/main.py - Request to unpack to None to .
1463457985 - [DEBUG] - atomicapp/plugin.py - Loading providers from /home/vagrant/atomicapp/atomicapp/providers
1463457985 - [DEBUG] - atomicapp/plugin.py - Loading providers from /home/vagrant/atomicapp/atomicapp/providers
1463457985 - [INFO] - nulecule/base.py - Pulling external application for mariadb-centos7-atomicapp.
1463457985 - [INFO] - nulecule/base.py - Unpacking image: projectatomic/mariadb-centos7-atomicapp to ./external/mariadb-centos7-atomicapp
[SNIP]
1463457989 - [DEBUG] - nulecule/main.py - FILE: ./answers.conf.gen
1463457989 - [DEBUG] - nulecule/main.py - ANSWERS: defaultdict(<type 'dict'>, {u'mariadb-centos7-atomicapp.mariadb-atomicapp': {u'db_pass': 'q', u'db_name': 'q', u'root_pass': u'MySQLPass', u'db_user': 'q'}, u'mariadb-centos7-atomicapp': {u'provider': u'kubernetes'}, u'etherpad-app': {u'db_pass': 'q', u'db_name': 'q', u'db_host': u'mariadb', u'db_user': 'q', u'image': u'centos/etherpad', u'hostport': 9001}, 'general': {'namespace': 'default', u'provider': 'kubernetes'}})

Your application resides in .
Please use this directory for managing your application

(atomic) ✔ ~/git/nulecule-library/etherpad-centos7-atomicapp [master ↑·39|…14] 
00:06 $ sudo atomicapp stop . -v
1463458001 - [INFO] - cli/main.py - Action/Mode Selected is: stop
1463458001 - [DEBUG] - cli/main.py - Final parsed cmdline: stop . -v
1463458001 - [DEBUG] - nulecule/main.py - NuleculeManager init app_path: .
1463458001 - [DEBUG] - nulecule/main.py - NuleculeManager init image: None
1463458001 - [DEBUG] - atomicapp/utils.py - Loading answers from file: ./answers.conf.gen
1463458001 - [ERROR] - cli/main.py - AnyMarkupError: caught <class 'yaml.parser.ParserError'>: expected '<document start>', but found '<scalar>'
  in "<file>", line 2, column 1
Original traceback:
  File "build/bdist.linux-x86_64/egg/anymarkup_core/__init__.py", line 110, in parse
    res = _do_parse(proper_inp, fmt, encoding, force_types)
  File "build/bdist.linux-x86_64/egg/anymarkup_core/__init__.py", line 252, in _do_parse
    res = yaml.safe_load(inp)
  File "build/bdist.linux-x86_64/egg/yaml/__init__.py", line 93, in safe_load
    return load(stream, SafeLoader)
  File "build/bdist.linux-x86_64/egg/yaml/__init__.py", line 71, in load
    return loader.get_single_data()
  File "build/bdist.linux-x86_64/egg/yaml/constructor.py", line 37, in get_single_data
    node = self.get_single_node()
  File "build/bdist.linux-x86_64/egg/yaml/composer.py", line 39, in get_single_node
    if not self.check_event(StreamEndEvent):
  File "build/bdist.linux-x86_64/egg/yaml/parser.py", line 98, in check_event
    self.current_event = self.state()
  File "build/bdist.linux-x86_64/egg/yaml/parser.py", line 174, in parse_document_start
    self.peek_token().start_mark)
ParserError: expected '<document start>', but found '<scalar>'
  in "<file>", line 2, column 1
  File "/home/vagrant/atomicapp/atomicapp/cli/main.py", line 130, in cli_func_exec
    cli_func(cli_func_args)
  File "/home/vagrant/atomicapp/atomicapp/cli/main.py", line 97, in cli_stop
    nm.stop(**argdict)
  File "/home/vagrant/atomicapp/atomicapp/nulecule/main.py", line 341, in stop
    self._process_answers()
  File "/home/vagrant/atomicapp/atomicapp/nulecule/main.py", line 396, in _process_answers
    self.answers = Utils.loadAnswers(self.answers_file)
  File "/home/vagrant/atomicapp/atomicapp/utils.py", line 367, in loadAnswers
    return anymarkup.parse_file(answers_file)
  File "build/bdist.linux-x86_64/egg/anymarkup_core/__init__.py", line 139, in parse_file
    return parse(f, format, encoding, force_types)
  File "build/bdist.linux-x86_64/egg/anymarkup_core/__init__.py", line 113, in parse
    raise AnyMarkupError(e, traceback.format_exc())
AnyMarkupError: AnyMarkupError: caught <class 'yaml.parser.ParserError'>: expected '<document start>', but found '<scalar>'
  in "<file>", line 2, column 1
Original traceback:
  File "build/bdist.linux-x86_64/egg/anymarkup_core/__init__.py", line 110, in parse
    res = _do_parse(proper_inp, fmt, encoding, force_types)
  File "build/bdist.linux-x86_64/egg/anymarkup_core/__init__.py", line 252, in _do_parse
    res = yaml.safe_load(inp)
  File "build/bdist.linux-x86_64/egg/yaml/__init__.py", line 93, in safe_load
    return load(stream, SafeLoader)
  File "build/bdist.linux-x86_64/egg/yaml/__init__.py", line 71, in load
    return loader.get_single_data()
  File "build/bdist.linux-x86_64/egg/yaml/constructor.py", line 37, in get_single_data
    node = self.get_single_node()
  File "build/bdist.linux-x86_64/egg/yaml/composer.py", line 39, in get_single_node
    if not self.check_event(StreamEndEvent):
  File "build/bdist.linux-x86_64/egg/yaml/parser.py", line 98, in check_event
    self.current_event = self.state()
  File "build/bdist.linux-x86_64/egg/yaml/parser.py", line 174, in parse_document_start
    self.peek_token().start_mark)
ParserError: expected '<document start>', but found '<scalar>'
  in "<file>", line 2, column 1
```
another one
```bash
(atomic) ✘-1 ~/git/nulecule-library/apache-centos7-atomicapp [master ↑·39|…17] 
00:03 $ sudo atomicapp run . -v
1463457814 - [INFO] - cli/main.py - Action/Mode Selected is: run
1463457814 - [DEBUG] - cli/main.py - Final parsed cmdline: run . -v
[SNIP]
1463457814 - [INFO] - providers/kubernetes.py - found kubectl at /usr/bin/kubectl
1463457814 - [INFO] - providers/kubernetes.py - Deploying to Kubernetes
1463457814 - [DEBUG] - providers/kubernetes.py - ./artifacts/kubernetes/.apache-centos7-atomicapp-pod.json
1463457814 - [DEBUG] - atomicapp/utils.py - 
<<< stdout >>>
pod "apache-centos7-atomicapp" created
<<< end >>>

1463457814 - [DEBUG] - atomicapp/utils.py - 
<<< stderr >>>
<<< end >>>

1463457814 - [DEBUG] - nulecule/main.py - Writing answers to file.
1463457814 - [DEBUG] - nulecule/main.py - FILE: ./answers.conf.gen
1463457814 - [DEBUG] - nulecule/main.py - ANSWERS: defaultdict(<type 'dict'>, {u'apache-centos7-atomicapp': {u'image': u'centos/httpd', u'hostport': 80}, 'general': {'namespace': 'default', u'provider': 'kubernetes'}})

Your application resides in .
Please use this directory for managing your application

(atomic) ✔ ~/git/nulecule-library/apache-centos7-atomicapp [master ↑·39|…18] 
00:03 $ sudo atomicapp stop .
[INFO] - main.py - Action/Mode Selected is: stop
[ERROR] - main.py - 'dict' object has no attribute 'set'
  File "/home/vagrant/atomicapp/atomicapp/cli/main.py", line 130, in cli_func_exec
    cli_func(cli_func_args)
  File "/home/vagrant/atomicapp/atomicapp/cli/main.py", line 97, in cli_stop
    nm.stop(**argdict)
  File "/home/vagrant/atomicapp/atomicapp/nulecule/main.py", line 346, in stop
    self.nulecule.load_config(config=self.answers)
  File "/home/vagrant/atomicapp/atomicapp/nulecule/base.py", line 242, in load_config
    config=config, ask=ask, skip_asking=skip_asking)
  File "/home/vagrant/atomicapp/atomicapp/nulecule/lib.py", line 73, in load_config
    config.set(param[NAME_KEY], value)
AttributeError: 'dict' object has no attribute 'set'
```

surajssd:
@rtnpro one more thing, I am not able to retrieve the command line parameters from the config object

```bash
(atomic) ✔ ~/git/nulecule-library/etherpad-centos7-atomicapp [master ↑·39|…11] 
01:21 $ sudo atomicapp run . -v --provider docker
1463462467 - [INFO] - cli/main.py - Action/Mode Selected is: run
[SNIP]
1463462467 - [DEBUG] - atomicapp/plugin.py - Loading providers from /home/vagrant/atomicapp/atomicapp/providers
==> db_user (Database User): q
==> db_pass (Database Password): a
==> db_name (Database Name): z
==> db_user (Database User): x
==> db_pass (Database Password): s
==> db_name (Database Name): w
[SNIP]
1463462476 - [DEBUG] - providers/docker.py - Given config: {u'db_pass': 'a', u'db_name': 'z', u'root_pass': u'MySQLPass', u'db_user': 'q', u'provider': u'kubernetes', 'namespace': 'default'}
1463462476 - [DEBUG] - providers/docker.py - Namespace: default
1463462476 - [WARNING] - providers/docker.py - The artifact name has not been provided within Nulecule, using a UUID instead
1463462476 - [DEBUG] - providers/docker.py - No image name found for artifact, using UUID aa7ea1b10b75 in container name
> /home/vagrant/atomicapp/atomicapp/nulecule/base.py(345)run()
    344         import ipdb; ipdb.set_trace()
--> 345         provider.run()
    346 

ipdb> import pprint
ipdb> pprint.pprint(self.config.__dict__)
{'_answers': defaultdict(<type 'dict'>, {u'mariadb-centos7-atomicapp.mariadb-atomicapp': {}, u'mariadb-centos7-atomicapp': {}, 'general': {'namespace': 'default'}}),
 '_cli': {},
 '_context': {u'db_name': 'z',
              u'db_pass': 'a',
              u'db_user': 'q',
              'namespace': 'default',
              u'provider': u'kubernetes',
              u'root_pass': u'MySQLPass'},
 '_current_ns': u'mariadb-atomicapp',
 '_data': defaultdict(<type 'dict'>, {u'mariadb-centos7-atomicapp.mariadb-atomicapp': {u'db_pass': 'a', u'db_name': 'z', u'db_user': 'q', u'root_pass': u'MySQLPass'}, u'mariadb-centos7-atomicapp': {u'provider': u'kubernetes'}, u'etherpad-app': {u'image': u'centos/etherpad', u'db_pass': 's', u'db_name': 'w', u'db_user': 'x', u'db_host': u'mariadb', u'hostport': 9001}, 'general': {u'provider': u'kubernetes'}}),
 '_is_nulecule': False,
 '_namespace': u'mariadb-centos7-atomicapp.mariadb-atomicapp',
 '_parent_ns': u'mariadb-centos7-atomicapp',
 '_provider': None}
ipdb> c
1463462527 - [INFO] - providers/docker.py - Deploying to provider: Docker
1463462527 - [INFO] - providers/docker.py - WARNING: Using --name provided within artifact file.
04d6972f9585820750e3e18a22593f66a769c6a5fc7da4640bff21737568390b
1463462527 - [DEBUG] - atomicapp/plugin.py - Found provider <class 'docker.DockerProvider'>
1463462527 - [WARNING] - atomicapp/plugin.py - Configuration option 'provider-config' not found
1463462527 - [DEBUG] - providers/docker.py - Given config: {u'db_pass': 's', u'db_name': 'w', u'db_host': u'mariadb', u'db_user': 'x', u'provider': u'kubernetes', u'image': u'centos/etherpad', u'hostport': 9001, 'namespace': 'default'}
1463462527 - [DEBUG] - providers/docker.py - Namespace: default
> /home/vagrant/atomicapp/atomicapp/nulecule/base.py(345)run()
    344         import ipdb; ipdb.set_trace()
--> 345         provider.run()
    346 

ipdb> import pprint
ipdb> pprint.pprint(self.config.__dict__)
{'_answers': defaultdict(<type 'dict'>, {u'etherpad-app': {}, 'general': {'namespace': 'default'}}),
 '_cli': {},
 '_context': {u'db_host': u'mariadb',
              u'db_name': 'w',
              u'db_pass': 's',
              u'db_user': 'x',
              u'hostport': 9001,
              u'image': u'centos/etherpad',
              'namespace': 'default',
              u'provider': u'kubernetes'},
 '_current_ns': u'etherpad-app',
 '_data': defaultdict(<type 'dict'>, {u'mariadb-centos7-atomicapp.mariadb-atomicapp': {u'db_pass': 'a', u'db_name': 'z', u'db_user': 'q', u'root_pass': u'MySQLPass'}, u'mariadb-centos7-atomicapp': {u'provider': u'kubernetes'}, u'etherpad-app': {u'image': u'centos/etherpad', u'db_pass': 's', u'db_name': 'w', u'db_user': 'x', u'db_host': u'mariadb', u'hostport': 9001}, 'general': {u'provider': u'kubernetes'}}),
 '_is_nulecule': False,
 '_namespace': u'etherpad-app',
 '_parent_ns': '',
 '_provider': None}
ipdb> c
1463462556 - [INFO] - providers/docker.py - Deploying to provider: Docker
1463462556 - [INFO] - providers/docker.py - WARNING: Using --name provided within artifact file.
181242ae722bf65e2d0c94b38a717058a3d56dc3c4bc711a0fcade238ca1dc51
1463462557 - [DEBUG] - nulecule/main.py - Writing answers to file.
1463462557 - [DEBUG] - nulecule/main.py - FILE: ./answers.conf.gen
1463462557 - [DEBUG] - nulecule/main.py - ANSWERS: defaultdict(<type 'dict'>, {u'mariadb-centos7-atomicapp.mariadb-atomicapp': {u'db_pass': 'a', u'db_name': 'z', u'root_pass': u'MySQLPass', u'db_user': 'q'}, u'mariadb-centos7-atomicapp': {u'provider': u'kubernetes'}, u'etherpad-app': {u'db_pass': 's', u'db_name': 'w', u'db_host': u'mariadb', u'db_user': 'x', u'image': u'centos/etherpad', u'hostport': 9001}, 'general': {'namespace': 'default', u'provider': u'kubernetes'}})

Your application resides in .
Please use this directory for managing your application
```

atomicapp/nulecule/config.py
```diff
@@ -0,0 +1,135 @@
...
+
+    @property
+    def namespace(self):
+        """
+        Get normalized namespace for this instance.
+
+        Returns:
+            Current namespace (str).
+        """
+        return self._namespace or GLOBAL_CONF
```
surajssd: We have a constant named `DEFAULT_NAMESPACE` why not use that, instead of `GLOBAL_CONF`?

dustymabe:
ping @cdrage @tkral - do you guys mind giving this a good look?

atomicapp/nulecule/config.py
rtnpro: @surajssd  ``DEFAULT_NAMESPACE`` and ``GLOBAL_CONF`` are different. ``GLOBAL_CONF`` stands for the ``general`` section in ``answers`` file.

surajssd:
- Since values like `provider` and `namespace` are set in `Config` class(making them property) so does that mean all the over-riding actions and decisions will be taken there?
- If overriding will be done, what is the precedence? (DEFAULT_PROVIDER < ANSWERS_PROVIDER < CLI_PROVIDER)
- Why some important values/attributes in `config` class start with `_`, which also means they are not reliable to use, so I have only `get` as a way of getting answers? What should I use there?
- How about the `config.get` method also returning the source of the value?
- Why `get` method not reading from `_cli` ? https://github.com/projectatomic/atomicapp/pull/720/files#diff-eb23c6d07d7b2b3e4fdf752145ce9dd4R80

atomicapp/nulecule/main.py
```diff
@@ -317,7 +319,9 @@ def run(self, cli_provider, answers_output, ask,
         if not self.answers_file:
             self._process_answers()
 
-        self.nulecule.load_config(config=self.nulecule.config, ask=ask)
+        self.nulecule.load_config(ask=ask)
+        if cli_provider:
+            self.nulecule.config.set('provider', cli_provider)
         self.nulecule.render(cli_provider, dryrun)
         self.nulecule.run(cli_provider, dryrun)
```
surajssd: changing above to something like this can help -> `self.nulecule.run(self.nulecule.config.set('provider', cli_provider), dryrun)`
If we do it here it removes the requirement of making checks at all here: https://github.com/projectatomic/atomicapp/pull/720/files#diff-093821a6e6f49d13d49fdcb4f911695dR93

rtnpro:
On Wed, May 25, 2016 at 2:19 PM, Suraj Deshmukh
<notifications@github.com> wrote:
> Since values like provider and namespace are set in Config class(making them
> property) so does that mean all the over-riding actions and decisions will
> be taken there?
Yes, in one place. IMO, we should have a single source of Truth.

> If overriding will be done, what is the precedence? (DEFAULT_PROVIDER <
> ANSWERS_PROVIDER < CLI_PROVIDER)
Yes

> Why some important values/attributes in config class start with _, which
> also means they are not reliable to use, so I have only get as a way of
> getting answers? What should I use there?
Yes, we should not access (read/write) the variables starting '_' from
outside the Config class.
The ``get`` is a way to retrieve data from the current namespace from
accessible namespaces.
This, however, creates room for some discrepancy, though. One
instance, will be, if you want to,
you can override a ``provider`` in a particular namespace, and
retrieve it using a ``get()`` call. But,
attributes like ``provider`` are global, and not namespace specific.
So, for such kind of global
attributes, I am using ``@property`` methods in Config class.

> How about the config.get method also returning the source of the value?
> Why get method not reading from _cli ?
> https://github.com/projectatomic/atomicapp/pull/720/files#diff-eb23c6d07d7b2b3e4fdf752145ce9dd4R80

We can take that into account. But, we shouldn't make the CLI too complex.



-- 
Ratnadeep Debnath,
https://www.waartaa.com
GPG Fingerprint: 033C 8041 A0E9 CDBA 2E02  B785 2119 5486 F245 DFD6

surajssd:
>We can take that into account. But, we shouldn't make the CLI too complex.

We don't need to make CLI or any thing any complex, all that can be done is put one more return somehow in `get`, the return will be like `value, "cli"` or `value, "answers"`

dustymabe:
Hey @rtnpro - Should we make the new Config class as generic as possible and then add features to it to be able to do what we want?

Right now we have arguments to `__init__()` like `answers=`, `cli=`, etc.. Shouldn't we make the Config class something that doesn't know much about the data it is storing? All the config class needs to know is: 
- I store key/value data
- The data can come from many sources
- I need a way for users to register sources that data comes from
- I need a way for users to specify priority of each source of data

So we could have something like: 

```
config = Config
config.registerSource(name='defaults', priority='1')
config.registerSource(name='answers', priority='10')
config.registerSource(name='cli', priority='100')
config.setval(key='provider', val='kubernetes', source='defaults')
config.setval(key='provider', val='openshift', source='answers')
config.setval(key='provider', val='docker', source='cli')
whatisprovider = config.getval(key='provider')
print whatisprovider
docker
```
we could also have a config.getvalallsources(key='provider') that would return something like:

```
{ cli => { source => cli, key => provider, val => docker, priority => 100 },
  answers => { source => answers, key => provider, val => docker, priority => 10 },
  defaults => { source => defaults, key => provider, val => docker, priority => 1 },
}
```

This function would be used when we needed to do some complex logic based on what value was provided from what source. The point is that the config class should be dumb and not know too much about the data it is handling, but just handle the data well.

A natural question is: where do we store that logic then? We can easily have a Config Class and then an AtomicAppConfig class that takes advantage of it and does higher level things like setting up the sources, etc.. 

Does that make sense? Am I crazy? Don't answer that last one :)

atomicapp/nulecule/base.py
```diff
@@ -78,6 +78,7 @@ def __init__(self, id, specversion, graph, basepath, metadata=None,
             params (list): List of params for the Nulecule application
             config (dict): Config data for the Nulecule application
             namespace (str): Namespace of the current Nulecule application
+            cli (dict): CLI data
```
dustymabe: I think this is the wrong way to go about it. We are already processing what we call `cli_answers` in the [cli]( https://github.com/projectatomic/atomicapp/blob/8c0805751570ee66d3f5d620679c9f9c88f8c68b/atomicapp/cli/main.py#L479-L485)

We do need to process data from the CLI, but I don't know if this is the right place. We should unify around this approach and what we are already doing.

rtnpro:
On Thu, May 26, 2016 at 2:15 AM, Dusty Mabe <notifications@github.com> wrote:
> Hey @rtnpro - Should we make the new Config class as generic as possible and
> then add features to it to be able to do what we want?
>
> Right now we have arguments to __init__() like answers=, cli=, etc..
> Shouldn't we make the Config class something that doesn't know much about
> the data it is storing? All the config class needs to know is:
>
> I store key/value data
> The data can come from many sources
> I need a way for users to register sources that data comes from
> I need a way for users to specify priority of each source of data
>
> So we could have something like:
>
> config = Config
> config.registerSource(name='defaults', priority='1')
> config.registerSource(name='answers', priority='10')
> config.registerSource(name='cli', priority='100')
> config.setval(key='provider', val='kubernetes', source='defaults')
> config.setval(key='provider', val='openshift', source='answers')
> config.setval(key='provider', val='docker', source='cli')
> whatisprovider = config.getval(key='provider')
> print whatisprovider
> docker

> we could also have a config.getvalallsources(key='provider') that would
> return something like:
>
> { cli => { source => cli, key => provider, val => docker, priority => 100 },
>   answers => { source => answers, key => provider, val => docker, priority
> => 10 },
>   defaults => { source => defaults, key => provider, val => docker, priority
> => 1 },
> }
>
> This function would be used when we needed to do some complex logic based on
> what value was provided from what source. The point is that the config class
> should be dumb and not know too much about the data it is handling, but just
> handle the data well.

I buy into this idea :+1:

> A natural question is: where do we store that logic then? We can easily have
> a Config Class and then an AtomicAppConfig class that takes advantage of it
> and does higher level things like setting up the sources, etc..

What about the ``namespace`` implementation? I did that keeping the
future in mind.

dustymabe:
> What about the ``namespace`` implementation? I did that keeping the future in mind.

TBH I think the "namespace" term is completely overloaded. We have "namespace" in the providers (kubernetes/openshift), we have "namespace" in the INI config files, etc.. Can we choose a new term that would help to disambiguate? It would also help if we had a short document that explained the strategy with the handling of config data. i.e. whatever new term we come up with for how we are going to handle "namespaces" in the config, let's write a short design document for how we are going to handle this stuff so people can read it and get the big picture before they dive into the code.

rtnpro:
@cdrage @kadel could you review this PR? I am reluctant to put too much effort to make the ``config`` class too generic for universal usage. I don't want to unnecessary add layers of abstraction. I see no way to do away with ``namespaces`` and neither do I find any other name apt to replace ``namespace``, may be a choice could be ``scope``.

As per @dustymabe's proposal, it will not be enough to just have sources, we need to have another layer for ``sections`` or ``namespaces`` in answers file.

@cdrage @kadel if you are OK with the code, I can fix the tests and get it merge ready.

kadel:
#dotests

atomicapp/cli/main.py
```diff
@@ -51,7 +51,7 @@ def print_app_location(app_path):
 def cli_genanswers(args):
     argdict = args.__dict__
     nm = NuleculeManager(app_spec=argdict['app_spec'],
-                         destination='none')
+                         destination='none', cli=argdict)
```
cdrage: Errrr...

argdict is already passed in `nm.genanswers(**argdict)`

Why do we need to pass it again? Could you change it so we simply pass it via NuleculeManager and remove the need to pass it within `nm.genanswers(**argdict)`?

atomicapp/nulecule/base.py
cdrage: Ah woops, my recent comment is the same as what @dustymabe just said about this PR. I agree with him too. Should only be passing it in once.

atomicapp/nulecule/base.py
```diff
@@ -88,11 +89,16 @@ def __init__(self, id, specversion, graph, basepath, metadata=None,
         self.metadata = metadata or {}
         self.graph = graph
         self.requirements = requirements
-        self.config = config or {}
+        if isinstance(config, Config):
+            self.config = config
+        else:
+            self.config = Config(namespace=namespace, answers=config,
+                                 cli=cli)
```
cdrage: Bit confused as to why there is an if statement here, if we are refactoring config, should we only pass it in once?

atomicapp/nulecule/config.py
```diff
@@ -0,0 +1,176 @@
...
+        self._parent_ns, self._current_ns = self._split_namespace(self._namespace)
+        # Store answers data
+        self._answers = defaultdict(dict)
+        self._answers.update(answers or {})
+        # Store CLI data
+        self._cli = cli or {}
+        # Store data collected during runtime
+        self._data = data or defaultdict(dict)
+        self._context = None
+        self._provider = None
```
cdrage: Why so many _?

atomicapp/nulecule/config.py
```diff
@@ -0,0 +1,176 @@
...
+        """
+        if self._is_nulecule:
+            return '', namespace
+        words = namespace.rsplit(NAMESPACE_SEPARATOR, 1)
+        parent, current = '', ''
+        if len(words) == 2:
+            parent, current = words[0], words[1]
+        else:
+            parent, current = '', words[0]
+        return parent, current
```
cdrage: Bit confused with this function. What's up with the separating?

What happens if there is a space in the namespace?

What are we checking here?

atomicapp/nulecule/base.py
```diff
@@ -236,19 +244,17 @@ def load_config(self, config=None, ask=False, skip_asking=False):
...
+            config = self.config
         super(Nulecule, self).load_config(
             config=config, ask=ask, skip_asking=skip_asking)
-        if self.namespace == GLOBAL_CONF and self.config[GLOBAL_CONF].get('provider') is None:
-            self.config[GLOBAL_CONF]['provider'] = DEFAULT_PROVIDER
-            logger.info("Provider not specified, using default provider - {}".
-                        format(DEFAULT_PROVIDER))
+
         for component in self.components:
             # FIXME: Find a better way to expose config data to components.
```
cdrage: I think this FIXME can be removed since you're fixing it :)

atomicapp/nulecule/base.py
```diff
@@ -269,8 +275,8 @@ def load_components(self, nodeps=False, dryrun=False):
             node_name = node[NAME_KEY]
             source = Utils.getSourceImage(node)
             component = NuleculeComponent(
-                node_name, self.basepath, source,
-                node.get(PARAMS_KEY), node.get(ARTIFACTS_KEY),
+                self._get_component_namespace(node_name), self.basepath,
+                source, node.get(PARAMS_KEY), node.get(ARTIFACTS_KEY),
                 self.config)
```
cdrage: PEP8? Fix spacing?

Ex:
```
self._get_component_namespace(node_name),
self.basepath,
source
...
```

atomicapp/nulecule/base.py
```diff
@@ -383,7 +396,8 @@ def load_external_application(self, dryrun=False, update=False):
                 'Found existing external application: %s '
                 'Loading: ' % self.name)
             nulecule = Nulecule.load_from_path(
-                external_app_path, dryrun=dryrun, update=update)
+                external_app_path, dryrun=dryrun, update=update,
+                namespace=self.namespace)
```
cdrage: PEP8

atomicapp/cli/main.py
rtnpro: :+1:

atomicapp/nulecule/base.py
rtnpro: I agree :+1:

atomicapp/nulecule/config.py
rtnpro: Because we are not encouraging people to access these variables directly.

cdrage: These are used in init anyways, they won't be able to access them unless you create it as a python read/write attribute

atomicapp/nulecule/config.py
rtnpro: We don't need to worry about ``space``s in the namespace, as we are splitting with ``.`` as a separator.

So, if a namespace is like:
``nulecule component1.nulecule 2.nulcule component 2``

then parent will namespace will be ``nulecule component1.nulecule 2`` and current namespace will be ``nulcule component 2``

atomicapp/nulecule/base.py
rtnpro: yay! :+1:

atomicapp/nulecule/base.py
rtnpro: ok

atomicapp/nulecule/base.py
rtnpro: ok

tests/units/nulecule/test_nulecule.py
```diff
@@ -50,70 +54,59 @@ def test_load_config_without_specified_provider(self):
...
+                    {
+                        "name": "key2",
+                        "default": "val2"
+                    }
+                ],
+                "artifacts": []
+            }
+        ]
 
-    def test_load_config_with_defaultprovider(self):
```
dustymabe: this function no longer exists: `test_load_config_with_defaultprovider`

tests/units/nulecule/test_nulecule.py
```diff
@@ -50,70 +54,59 @@ def test_load_config_without_specified_provider(self):
...
+                'key1': 'val1',
+                'key3': 'val3'
+            },
+            'component1': {
+                'key2': 'val2',
+                'key1': 'val1'
+            }
         })
 
-    def test_load_config_with_defaultprovider_overridden_by_provider_in_answers(self):
```
dustymabe: this function no longer exists: `test_load_config_with_defaultprovider_overridden_by_provider_in_answers`

dustymabe:
> @cdrage @kadel could you review this PR? I am reluctant to put too much effort to make the config class too generic for universal usage. I don't want to unnecessary add layers of abstraction. I see no way to do away with namespaces and neither do I find any other name apt to replace namespace, may be a choice could be scope.

"scope" could work..

I really think making the config generic and then building on top of it would enable us to understand things more. Having it too "intertwined" is why we are here doing this refactor in the first place. We could implement a class as described in https://github.com/projectatomic/atomicapp/pull/720#issuecomment-221701862 to help with this. 

> 
> As per @dustymabe's proposal, it will not be enough to just have sources, we need to have another layer for sections or namespaces in answers file.

We could extend the class that I described and just add another element. Configparser has "sections" https://docs.python.org/2/library/configparser.html#ConfigParser.ConfigParser.get. We could easily have the same thing called "scopes" or "namespaces". Which would modify my class described to something like this: 

```
config.setval(key='provider', val='kubernetes', scope='general', source='defaults')
whatisprovider = config.getval(key='provider', scope='general')
print whatisprovider
config.setval(key='dbname', val='defaultdb'', scope='mariadb', source='defaults')
config.setval(key='dbname', val='answersdb'', scope='mariadb', source='answers')
dbname = config.getval(key='dbname'', scope='mariadb'')
print dbname
```

rtnpro:
@cdrage tests are all passing now. Can you have a final look and merge it?

dustymabe:
> @cdrage tests are all passing now. Can you have a final look and merge it?

@rtnpro you are ignoring comment https://github.com/projectatomic/atomicapp/pull/720/files#r69980437 and comment https://github.com/projectatomic/atomicapp/pull/720#issuecomment-231207183

tests/units/nulecule/test_nulecule.py
rtnpro: I can bring it back. :+1:

rtnpro:
@dustymabe regarding https://github.com/projectatomic/atomicapp/pull/720#issuecomment-231207183, I find it a lot of over engineering ATM. We just wanted to abstract a dict based config in this PR.

atomicapp/nulecule/main.py
```diff
@@ -272,29 +272,26 @@ def fetch(self, nodeps=False, update=False, dryrun=False,
...
+        self.nulecule.load_config(skip_asking=True)
         runtime_answers = self._get_runtime_answers(
             self.nulecule.config, None)
         # write sample answers file
         self._write_answers(
             os.path.join(self.app_path, ANSWERS_FILE_SAMPLE),
             runtime_answers, answers_format)
 
-    cockpit_logger.info("Install Successful.")
+        cockpit_logger.info("Install Successful.")
```
cdrage: fixing spacing

cdrage:
Good work!

I don't see anything major that needs to be changed in the code. Much cleaner than the previous commit that was in.

This commit should be squashed into one instead of the many commits which are here. Would be much more difficult to understand from a history perspective going back through the code.

Tests are still failing.. Once the functional tests are green + unit tests we can do one final review + manual testing and we can merge.

cdrage:
#dotests

rtnpro:
On Fri, Jul 22, 2016 at 5:49 PM, Charlie Drage <notifications@github.com> wrote:
>
> Good work!
>
> I don't see anything major that needs to be changed in the code. Much cleaner than the previous commit that was in.
>
> This commit should be squashed into one instead of the many commits which are here. Would be much more difficult to understand from a history perspective going back through the code.

:+1: for squashing the commits.

> Tests are still failing.. Once the functional tests are green + unit tests we can do one final review + manual testing and we can merge.

I will fix the tests now. I was waiting for your feedback on the code :)

cdrage:
@rtnpro Each unit test is stuck on removing the docker container. Looks like a bug

atomicapp/nulecule/config.py
```diff
@@ -0,0 +1,123 @@
...
+
+        Returns:
+            Value for the key.
+        """
+        value = None
+        for source in self.PRIORITY:
+            value = self._data[source][scope].get(key)
+            if value:
+                break
+        return value
```
dustymabe: so why don't we do:
```
+        for source in self.PRIORITY:
+            value = self._data[source][scope].get(key)
+            if value:
+                return value
+        return None
```

atomicapp/nulecule/config.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def set(self, key, value, scope=GLOBAL_CONF, source='defaults'):
```
dustymabe: I would say we shouldn't have a default value for the source argument. We should force the caller to set that value IMO.

atomicapp/nulecule/config.py
```diff
@@ -0,0 +1,123 @@
...
+        data is used to render the artifacts of Nulecule graph item.
+
+        Args:
+            scope (str): Scope (or namespace) for the Nulecule graph item.
+        Returns:
+            A dictionary
+        """
+        result = {}
+        for source in reversed(self.PRIORITY):
+            data = self._data[source]
```
dustymabe: can we call the new var something like sourcedata vs just data.. starting to get confusing.

atomicapp/nulecule/config.py
```diff
@@ -0,0 +1,123 @@
...
+            key (str): Key
+            value: Value
+            scope (str): Scope in which to store the value
+            source (str): Source of the value
+        """
+        self._data[source][scope][key] = value
+
+    def context(self, scope=GLOBAL_CONF):
+        """
+        Get context data for the scope of Nulecule graph item. This context
```
dustymabe: can you give me more context here? :) what is the 6 lines of code below really doing and why is it useful?

atomicapp/nulecule/base.py
```diff
@@ -175,8 +175,9 @@ def load_from_path(cls, src, config=None, namespace=GLOBAL_CONF,
             raise NuleculeException("Failure parsing %s file. Validation error on line %s, column %s:\n%s"
                                     % (nulecule_path, line, column, output))
 
-        nulecule = Nulecule(config=config, basepath=src,
-                            namespace=namespace, **nulecule_data)
+        nulecule = Nulecule(config=config,
+                            basepath=src, namespace=namespace,
+                            **nulecule_data)
```
dustymabe: What changed here? If nothing can we put it back how it was so it doesn't show up in the diff?

atomicapp/nulecule/base.py
dustymabe: the FIXME comment is still there in the code.

atomicapp/nulecule/base.py
```diff
@@ -293,6 +292,12 @@ def render(self, provider_key=None, dryrun=False):
         for component in self.components:
             component.render(provider_key=provider_key, dryrun=dryrun)
 
+    def _get_component_namespace(self, component_name):
```
dustymabe: Could use a description somewhere for this function

atomicapp/nulecule/config.py
```diff
@@ -0,0 +1,123 @@
...
+            cli (dict): CLI data
+        """
+        answers = answers or {}
+        cli = cli or {}
+        self._data = defaultdict(defaultdict)
+        self._data['defaults'] = defaultdict(defaultdict)
+        self._data['answers'] = defaultdict(defaultdict)
+        self._data['cli'] = defaultdict(defaultdict)
+        self._data['runtime'] = defaultdict(defaultdict)
+
```
dustymabe: would be nice to have a 1 line comment above each one of these for loops below

atomicapp/nulecule/config.py
```diff
@@ -0,0 +1,123 @@
...
+        Args:
+            scope (str): Scope (or namespace) for the Nulecule graph item.
+        Returns:
+            A dictionary
+        """
+        result = {}
+        for source in reversed(self.PRIORITY):
+            data = self._data[source]
+            result.update(copy.deepcopy(data.get(GLOBAL_CONF) or {}))
+            result.update(copy.deepcopy(data.get(scope) or {}))
```
dustymabe: why do we have two result.update() lines here. One is from `GLOBAL_CONF` and one is for `scope`, but `scope` defaults to `GLOBAL_CONF`. The relationship could probably use a comment with some explanation.

atomicapp/nulecule/config.py
```diff
@@ -0,0 +1,123 @@
...
+        answers = defaultdict(dict)
+
+        for source in reversed(self.PRIORITY):
+            for scope, data in (self._data.get(source) or {}).items():
+                answers[scope].update(copy.deepcopy(data))
+
+        # Remove empty sections for answers
+        for key, value in answers.items():
+            if value is None:
+                answers.pop(key, None)
```
dustymabe: why `answers.pop(key, None)` vs `answers.pop(key)`. we already know `key` is in the dict right? so there is no risk of `KeyError` being raised

atomicapp/nulecule/lib.py
```diff
@@ -64,46 +63,21 @@ def load_config(self, config, ask=False, skip_asking=False):
             None
         """
         for param in self.params:
-            value = config.get(self.namespace, {}).get(param[NAME_KEY]) or \
-                config.get(GLOBAL_CONF, {}).get(param[NAME_KEY])
+            value = config.get(param[NAME_KEY], scope=self.namespace)
```
dustymabe: @rtnpro - would it be reasonable in the future (not for the PR) for us to maybe rename `namespace` from this class to `scope`?

atomicapp/nulecule/config.py
```diff
@@ -0,0 +1,123 @@
...
+cockpit_logger = logging.getLogger(LOGGER_COCKPIT)
+
+
+class Config(object):
+    """
+    This class allows to store config data in different scopes along with
+    source info for the data. When fetching the value for a key in a scope,
+    the source info and the PRIORITY order of sources is taken into account.
+    """
+
```
dustymabe: Can we get a summary of each `source` below in a comment. This would help future readers of the code.

atomicapp/nulecule/lib.py
```diff
@@ -64,46 +63,21 @@ def load_config(self, config, ask=False, skip_asking=False):
...
             if value is None and (ask or (
                     not skip_asking and param.get(DEFAULTNAME_KEY) is None)):
                 cockpit_logger.info("%s is missing in answers.conf." % param[NAME_KEY])
                 value = Utils.askFor(param[NAME_KEY], param)
             elif value is None:
                 value = param.get(DEFAULTNAME_KEY)
-            if config.get(self.namespace) is None:
-                config[self.namespace] = {}
-            config[self.namespace][param[NAME_KEY]] = value
+            config.set(param[NAME_KEY], value, scope=self.namespace, source='runtime')
```
dustymabe: we should only run `config.set` with `source='runtime'` if the value was `None` before.

atomicapp/nulecule/lib.py
```diff
@@ -64,46 +63,21 @@ def load_config(self, config, ask=False, skip_asking=False):
...
-        """
-        for group, group_vars in from_config.items():
-            to_config[group] = to_config.get(group) or {}
-            for key, value in (group_vars or {}).items():
-                if to_config[group].get(key) is None:
-                    to_config[group][key] = value
-
     def get_context(self):
         """
         Get context data from config data for rendering an artifact.
```
dustymabe: could still really use a better definition here.. "get context data" is pretty ambiguous

atomicapp/nulecule/main.py
```diff
@@ -64,6 +63,7 @@ def __init__(self, app_spec, destination=None,
             destination: where to unpack a nulecule to if it isn't local
             cli_answers: some answer file values provided from cli args
             answers_file: the location of the answers file
+            cli (dict): CLI data
```
dustymabe: I don't see any other reference to this.. copy/paste mistake?

atomicapp/nulecule/main.py
```diff
@@ -64,6 +63,7 @@ def __init__(self, app_spec, destination=None,
...
         """
         self.answers = copy.deepcopy(DEFAULT_ANSWERS)
         self.cli_answers = cli_answers
```
dustymabe: Ok so I'm going to go out on a limb and say the `self.answers` and `self.cli_answers` should no longer exist as instance/class variables. we should really be creating a `self.config = Config()` in this _init_ function and using that exclusively throughout the rest of the code, right? The point of the Config class is so that we aren't handling our data multiple times in various structures.

We should be able to run `_process_answers()` and update that function to return actual answer data as a return value that we can then use when calling `Config()`. or something along those lines.

atomicapp/nulecule/config.py
rtnpro: we don't need the ``return None`` as well, as ``None`` is returned by default.

atomicapp/nulecule/base.py
rtnpro: Yes, It's a mistake.

atomicapp/nulecule/config.py
rtnpro: the second ``result.update`` is mainly when ``scope != GLOBAL_CONF``. I will add that conditional.

atomicapp/nulecule/config.py
rtnpro: I was just being over cautious.

atomicapp/nulecule/lib.py
rtnpro: Created issue for this at: https://github.com/projectatomic/atomicapp/issues/780

atomicapp/nulecule/lib.py
rtnpro: It's an assignment by reference, so we can move ``self.config = config`` to the top of the function.

atomicapp/nulecule/lib.py
rtnpro: We can just get rid of this method with ``context = self.config.context(self.namespace)``

atomicapp/nulecule/main.py
rtnpro: I was just not touching this vars to keep the changes minimal, but we have already done a lot of changes, so, I will go ahead and clean up this bit as well.

atomicapp/nulecule/config.py
```diff
@@ -0,0 +1,149 @@
...
+            key (str): Key
+            scope (str): Scope from which to fetch the value for the key
+
+        Returns:
+            Value for the key.
+        """
+        for source in self.PRIORITY:
+            value = self._data[source][scope].get(key)
+            if value:
+                return value
```
dustymabe: I know `None` will get returned here if no call to return is given but should we explicitly do it anyway in order to stop people being confused about whether we meant to do it that way or not.

dustymabe:
Changes look good. I know we had a chat today about `_process_answers()` where we are going to refactor that a bit. Feel free to do that and also to squash your commits (we discussed that as well), but I would encourage you to not squash them all down to one commit. Maybe 2-4 commits that have meaningful messages would be best.

rtnpro:
#dotests

atomicapp/nulecule/main.py
```diff
@@ -54,7 +53,8 @@ class NuleculeManager(object):
     """
 
     def __init__(self, app_spec, destination=None,
-                 cli_answers=None, answers_file=None):
+                 cli_answers=None, answers_file=None,
+                 answers_format=ANSWERS_FILE_SAMPLE_FORMAT):
```
dustymabe: should probably be `answers_format=None` since we set this in the first line of the function anyway

atomicapp/nulecule/main.py
```diff
@@ -118,7 +117,7 @@ def __init__(self, app_spec, destination=None,
 
         # Process answers.
         self.answers_file = answers_file
-        self._process_answers()
+        self.config = self._process_answers(cli_answers)
```
dustymabe: I think the right approach is to not call process_answers here but only in run/stop.
Here we would call `self.config=Config(cli=cli_answers)` and then later in `_process_answers()` we would call a new function from the config class to update answers like: `self.config.update_source(source='answers', data=answers)` or something along those lines.

dustymabe: I believe this is the approach we discussed the other day, right?

rtnpro:
#dotests docker

rtnpro:
#dotests-docker

atomicapp/nulecule/main.py
rtnpro: Yes!

rtnpro:
#dotestdocker

cdrage:
#dotests

atomicapp/utils.py
```diff
@@ -360,13 +360,17 @@ def getUniqueUUID():
...
             raise AtomicAppUtilsException(
                 "Provided answers file does not exist: %s" % answers_file)
 
         logger.debug("Loading answers from file: %s", answers_file)
-        return anymarkup.parse_file(answers_file)
+        try:
+            result = anymarkup.parse_file(answers_file, format=format)
+        except anymarkup.AnyMarkupError:
+            result = anymarkup.parse_file(answers_file)
+        return result
```
dustymabe: what are we fixing here?

dustymabe:
hey - mostly looks good. when i run genanswers on our etherpad example I see this:

```
[vagrant@f24 etherpad-centos7-atomicapp (upstreammaster *%=)]$ myatomicapp genanswers ./ 
INFO   :: Atomic App: 0.6.2 - Mode: Genanswers
INFO   :: Pulling external application: mariadb-centos7-atomicapp
INFO   :: Unpacking image projectatomic/mariadb-centos7-atomicapp to /tmp/atomicappVb0gbX/external/mariadb-centos7-atomicapp
INFO   :: Skipping pulling docker image: projectatomic/mariadb-centos7-atomicapp
INFO   :: Extracting Nulecule data from image projectatomic/mariadb-centos7-atomicapp to /tmp/atomicappVb0gbX/external/mariadb-centos7-atomicapp
INFO   :: Copying files from image projectatomic/mariadb-centos7-atomicapp:application-entity to /tmp/nulecule-d8a7d8d6-5fce-11e6-b929-52540043f1b2
[vagrant@f24 etherpad-centos7-atomicapp (upstreammaster *%=)]$ cat answers.conf 
[mariadb-centos7-atomicapp]
[mariadb-centos7-atomicapp:mariadb-atomicapp]
db_pass = None
image = centos/mariadb
db_user = None
root_pass = MySQLPass
db_name = None
[etherpad-app]
db_pass = None
db_name = None
db_user = None
db_host = mariadb
hostport = 9001
image = centos/etherpad
db_port = 3306
[general]
namespace = default
provider = kubernetes
```

there is the extra [mariadb-centos7-atomicapp] in there

atomicapp/utils.py
rtnpro: so, when we have ``:`` or ``.`` in the section names in an ``INI`` file, anymarkup does not identify it as an INI file by default, so, we have to explicitly mention the fromat when parsing the file. This L370 also works well if the user specifies a non default answers format, as well.

L372 is to handle the scenario when the user points to a JSON/YAML answers file but does not set the answers format, which defaults to ``INI``. In such a case, L370 won't be able to load the file inspite of the fact that the file is a proper JSON, YAML file.

dustymabe: a couple of things:

- this would be good to add in as a comment because no one else would figure that out
- the fact that anymarkup does not detect an INI file if it has `:` - is that a bug in anymarkup or do INI files really not allow colons in the section headings. If that violates the INI file spec I'd be less happy about using it.

rtnpro: Actions:
- I will add some inline docs
- I will check the INI file format to see if it allows ``:`` in the section name

dustymabe:
rebase on latest master?

rtnpro:
@dustymabe @cdrage 

Another thing, we need to ensure that whatever delimiter we use for namespace, it should not appear in the names of the nulecule or graph item. So, chars like ``.`` or ``:`` are not that well suited for it. Should we use some characters like ``::``, which will not usually appear in the names?

Also, the Nulecule spec does not speak about content of ``answers`` file, so it makes it unnecessary to make changes to the spec.

So, I think we can proceed with this merge as:
- Decide on a better namespace delimiter, e.g., ``::``
-  Merge: https://github.com/projectatomic/adb-tests/pull/41
-  Merge this PR.

rtnpro:
Plan for merging this change:
- Merge https://github.com/projectatomic/adb-tests/pull/41
- Run ``dotests`` for this PR
- If functional tests are successful, merge this PR, and merge https://github.com/projectatomic/nulecule/pull/211/

@dustymabe @cdrage ^^

dustymabe:
hey @rtnpro I have reviewed this several times in the past. as far as this PR goes I am +1 if @cdrage approves.

as for https://github.com/projectatomic/nulecule/pull/211 shouldn't we also update https://github.com/projectatomic/atomicapp/blob/2f980660cb23ec3f5df53d06c4159d3af0390214/docs/spec/NULECULE_FILE.md within this repo?

cdrage:
#dotests

cdrage:
YAY! `atomicapp genanswers` works. I'm going to run through a bit more testing and then we're LGTM for the merge

cdrage:
#dotests

cdrage:
Local tests pass :+1:

cdrage:
k8s + docker pass, openshift timed out (tested it locally to make 100% it works)

LGTM!

thanks a lot @rtnpro !

PR #721, opened 2016-05-05
Remove README.MD from init

cdrage:
Init should be clean and thus no README.md should be provided. Some
people use pandoc, ascii, etc.

ping @rtnpro


cdrage:
Timeout... ping @rtnpro :)

rtnpro:
:+1: LGTM

PR #726, opened 2016-05-13
Hide container id output on container creation

cdrage:
This commit removes container ID output when talking to Docker either
for the Nulecule data extraction or when deploying a multi-container
application.

Replaces PR: https://github.com/projectatomic/atomicapp/pull/687

### DIFF
--- a/atomicapp/nulecule/container.py
+++ b/atomicapp/nulecule/container.py
@@ -85,7 +85,7 @@ class DockerHandler(object):
 
         if self.dryrun:
             logger.info("DRY-RUN: %s", pull_cmd)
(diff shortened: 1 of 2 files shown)


rtnpro:
LGTM :+1:

dustymabe:
For the changes where you went from `call()` to `check_output()` you changed the behavior. `call()` doesn't raise an exception when the command fails, `check_output()` does.

cdrage:
@dustymabe Ah crap. So for this, should still leave it be as it *should* fail on if it can't remove, etc? Or ignore the exception (if it ever occurs)

PR #727, opened 2016-05-13
Change dir of /external to /nulecule/external

cdrage:
This folder should be located within /nulecule/external instead of
/external.

Thus within the code there is no need to include '/../'.

You were right @rtnpro, it should be there instead :)


rtnpro:
LGTM :+1:

PR #734, opened 2016-05-25
Remove ASCII art

cdrage:
Remove ASCII art in favour of a more minimalistic logging

### DIFF
--- a/atomicapp/cli/main.py
+++ b/atomicapp/cli/main.py
@@ -448,18 +448,9 @@ class CLI():
         # Setup logging (now with arguments from cmdline) and log a few msgs
         Logging.setup_logging(args.verbose, args.quiet, args.logtype)
 
-        art = """
-   _  _             _      _
-  /_\| |_ ___ _ __ (_)__  /_\  _ __ _ __
- / _ \  _/ _ \ '  \| / _|/ _ \| '_ \ '_ \\
-/_/ \_\__\___/_|_|_|_\__/_/ \_\ .__/ .__/
-                               |_|  |_|
-        """.splitlines()
-        logger.info("%s" % art[1])
-        logger.info("%s   Version:  %s" % (art[2], __ATOMICAPPVERSION__))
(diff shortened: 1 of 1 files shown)


dustymabe:
So it might be good to include in the commit message why we decided to remove it. 

Also could we merge the "replacement" text down to just a single line of output? I think the ultimate goal is to reduce the number of lines - if we still had the same number of lines I'd rather just keep the ascii art.

@jberkus - want to comment?

cdrage:
@dustymabe @jberkus 
What type of formatting would you like for the versioning / nulecule versioning, name of the application, etc? Could you give me an example?

dustymabe:
@cdrage - a single line like this maybe:

```
INFO   :: - cli/main.py :: Atomic App: 0.5.2 - Nulecule: 0.0.2 - Mode: Run
```

Same Amount of Information - single line

That brings up a nother question - what exactly is the Nulecule version we are printing out? Is it the minimum supported by Atomic App? Is it the maximum?

cdrage:
@dustymabe Updated with your suggestions.

Hmmm. Recommended Nulecule version?

dustymabe:
> Hmmm. Recommended Nulecule version?

Yeah I don't really know what it is - we keep it as a constant in the code. Do we use it anywhere?

cdrage:
@dustymabe 
Other than `atomicapp init`, nope.

dustymabe:
@cdrage - hmm so I think what is is supposed to indicate is what version/versions we are supposed to support.

it should probably show the "minimum" supported version - or we could just not output that information for now since we can't even really determine why it's useful

cdrage:
@dustymabe we can just output Atomic App version and the run mode. No need for Nulecule spec version

dustymabe:
> @dustymabe we can just output Atomic App version and the run mode. No need for Nulecule spec version

+1

cdrage:
@dustymabe updated, feel free to have a look over and LGTM it :)

dustymabe:
+1

cdrage:
#dotests

PR #739, opened 2016-06-10
add support for api-groups

### LINKED ISSUE #738: Support for v1beta in Kubernetes API / provider
Because we use the `api/v1` end-point, certain things aren't deployable for example, from the `v1beta` extensions. Source: http://kubernetes.io/docs/api-reference/extensions/v1beta1/definitions/

Let's fix this :) One of the things we do is check `resources` for the list of API end-points that are allowed to be deployed. The same should be kept for checking the `v1/beta` endpoints, `v2` in the future, etc.

khrisrichardson:
Fixes #738

### DIFF
--- a/atomicapp/providers/lib/kubeshift/kubernetes.py
+++ b/atomicapp/providers/lib/kubeshift/kubernetes.py
@@ -19,11 +19,13 @@
 
 from urlparse import urljoin
 from urllib import urlencode
-from atomicapp.providers.lib.kubeshift.kubebase import KubeBase
-from atomicapp.providers.lib.kubeshift.exceptions import (KubeKubernetesError)
-from atomicapp.constants import LOGGER_DEFAULT
 import logging
 import re
+
+from atomicapp.constants import LOGGER_DEFAULT
+from atomicapp.providers.lib.kubeshift.kubebase import KubeBase
+from atomicapp.providers.lib.kubeshift.exceptions import (KubeKubernetesError)
+
 logger = logging.getLogger(LOGGER_DEFAULT)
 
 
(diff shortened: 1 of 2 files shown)


dustymabe:
@cdrage and @kadel - can I get you guys to review this patch?

dustymabe:
#dotests

kadel:
:+1: LGTM
Thank you  @khrisrichardson code looks nice, and works even better :wink:

rtnpro:
:+1: LGTM

atomicapp/providers/lib/kubeshift/kubernetes.py
```diff
@@ -52,7 +54,20 @@ def __init__(self, config):
...
+        self.k8s_apis = urljoin(url, "apis/")
+
+        # Gather the group names from which resource names will be derived
+        self.k8s_api_groups = self.api.get_groups(self.k8s_apis)
+
+        for (name, versions) in self.k8s_api_groups:
+            for version in versions:
+                api = "%s/%s" % (name, version)
+                url = urljoin(self.k8s_apis, api)
+                self.k8s_api_resources[api] = self.api.get_resources(url)
```
cdrage: Nothing wrong with this part, but in the future when we add the refactored OpenShift code (that uses KubeBase) we should add this above section to KubeBase as well since both OpenShift and Kubernetes share the similar end-points (v1beta, etc.).

Thanks for the contribution @khrisrichardson !

cdrage:
Thank you very much @khrisrichardson for the contribution to Atomic App! Code is done beautifully and tests pass with flying colors against Kubernetes. 

I've tested this against Kubernetes v1.3.0-alpha.5 and it works well!

Code LGTM, mergin'

PR #746, opened 2016-06-20
use determinate directed graph

### LINKED ISSUE #731: nulecule component graph rendering is non-Hamiltonian
From an examination of the code in [nulecule/base.py](https://github.com/projectatomic/atomicapp/blob/master/atomicapp/nulecule/base.py), it would appear that what is referred to as a graph is actually rendered as nested lists of lists.

The predicament this raises is the potential for the same "node" to appear multiple times in the data structure if it is a dependency of multiple other "nodes", which invariably results in an error and terminates execution when the "node" is deployed to a provider a second time.

To avoid such a predicament an acyclic hamiltonian graph is needed in place of the nested list of lists. networkx is one such library that could help with topological sorting.

khrisrichardson:
Fixes #731 

@rtnpro, this is a functioning implementation of a directed graph, which is deterministic, whereas as discussed previously, a recursive implementation is prone to indeterminism in the face of a reoccurring dependency.

Although at first I balked at the common abstract base class for the Nulecule and NuleculeComponent concrete classes, I'm no longer opposed to decomposing this back into two classes for an easier to comprehend mapping back to the spec: one to represent the Nulecule manifest object, and the other to represent the Nulecule graph component.

I think my objection was that even though the classes share some common attributes, their common inheritance felt somewhat convenient as opposed to representative. Not intended as a jab, but rather justifying my own thought processes. As separate classes, the "progenitor" attribute would be less clumsy.

FYI - I tested this against a majority of the examples in nulecule-library without a hitch. I'm not opposed to decomposing this PR into functional changes and pylint recommendations. Also let me know if there are any objection to squashing the nesting of external dependencies into a single directory, which does pose the potential for conflicts if unique components are named identically.

And finally the topological sort "should" catch cycles in the graph, and thus could benefit from exception handling to warn the user of circular dependencies in a friendlier manner than a stack trace. Also given that cycles are now catchable, it is probably worthwhile to add idempotency to the provider requests. I've done this already for Kubernetes, and will create an additional PR.

No really, one final thought: for Kubernetes, as opposed to depending on the existence of an answers.conf.gen file to "stop" components, it may be worthwhile to explore [annotations](http://kubernetes.io/docs/user-guide/annotations/) as a way to tag and introspect, since the answers.conf.gen could very easily be deleted mistakenly.


rtnpro:
@khrisrichardson I had a chance to test your changes and they worked fine. However, I was not able to review your code, as I am stuck with some CentOS container pipeline work. I will review your code and get back to you ASAP.

khrisrichardson:
Thanks @rtnpro. I've been trying to chip a way at the unit tests, of which there are still several failures, when I've managed to find some free time, so no huge rush.

atomicapp/nulecule/base.py
```diff
@@ -62,37 +62,49 @@ class Nulecule(NuleculeBase):
...
             specversion (str): Nulecule spec version
-            graph (list): Nulecule graph of components
             basepath (str): Basepath for Nulecule application
             metadata (dict): Nulecule metadata
             requirements (dict): Requirements for the Nulecule application
             params (list): List of params for the Nulecule application
             config (dict): Config data for the Nulecule application
             namespace (str): Namespace of the current Nulecule application
+            progenitor (bool): Whether or not Nulecule instance is attributable
+                to parent manifest as opposed to deployable graph component
```
rtnpro: doc for ``artifacts`` is missing.

atomicapp/nulecule/base.py
```diff
@@ -174,11 +190,25 @@ def load_from_path(cls, src, config=None, namespace=GLOBAL_CONF,
...
-
-        nulecule = Nulecule(config=config, basepath=src,
-                            namespace=namespace, **nulecule_data)
-        nulecule.load_components(nodeps, dryrun)
-        return nulecule
+        else:
+            kwargs = deepcopy(nulecule_data)
+            kwargs['progenitor'] = True
+            del kwargs[GRAPH_KEY]
+            progenitor = Nulecule(config=config, basepath=src,
```
rtnpro: Isn't ``progenitor`` always going to be ``None`` since ``__init__()`` of ``Nulecule`` does not return anything.

atomicapp/nulecule/main.py
```diff
@@ -185,14 +158,55 @@ def init(app_name, destination=None, app_version='1.0',
...
+
+        for external in externals:
+            app_spec = Utils.getSourceImage(external)
+            name = external[NAME_KEY]
+            destination = "%s/%s/%s" % (self.app_path, EXTERNAL_APP_DIR, name)
+            if app_spec is not None:
+                kwargs.update({'namespace': name})
+                kwargs.update(self._process_app_spec(app_spec, destination))
+                subgraph = self.digraph(**kwargs)
+                if isinstance(subgraph, DiGraph) and subgraph.order() > 0:
```
rtnpro: isn't ``subgraph`` always going to be an instance of ``DiGraph``?

khrisrichardson: Yeah, that's an artifact from a previous rendition of the code. I will clip that.

atomicapp/nulecule/main.py
```diff
@@ -185,14 +158,55 @@ def init(app_name, destination=None, app_version='1.0',
...
+        for external in externals:
+            app_spec = Utils.getSourceImage(external)
+            name = external[NAME_KEY]
+            destination = "%s/%s/%s" % (self.app_path, EXTERNAL_APP_DIR, name)
+            if app_spec is not None:
+                kwargs.update({'namespace': name})
+                kwargs.update(self._process_app_spec(app_spec, destination))
+                subgraph = self.digraph(**kwargs)
+                if subgraph.order() > 0:
+                    patronage = internals or [progenitor]
```
rtnpro: Could you please elaborate on this logic? Why ``internals or [progenitor]``? How can the ``internals`` graph items patron an external app?

khrisrichardson: Of all the lines of code, you found the one that I thought was most worthy of some commentary, yet hardest to explain. Essentially what's going on here is as follows, which probably makes it worthwhile to separate things back into two classes, assuming they don't inherit from the same base class.

```progenitor``` and ```externals``` each have the ```progenitor``` attribute set to ```true``` and align with what was previously the ```Nulecule``` class, whereas ```internals``` describe the actual units of deployment and align with what was previously the ```NuleculeComponent``` class. It seems reasonable to restore that basic class hierarchy.

Take for example two use cases.

```
---
specversion: 0.0.2
id: NuleculeA

metadata:
  name: NuleculeA
  appversion: 1.0
  description: App description

graph:
  - name: NuleculeB
    source: docker://NuleculeB

  - name: NuleculeC
    source: docker://NuleculeC
```

I'll revert to using the prior lingo to explain what's going on. In this first use case, we have a ```Nulecule``` with relationships only to other ```Nulecule``` type objects, whereas in the latter case the ```Nulecule``` has a relationship to both ```Nulecule``` and ```NuleculeComponent``` type objects.

In the first example, since there are no ```NuleculeComponent``` type objects, we need to create an edge from ```NuleculeA``` (i.e. ```or [progenitor]```) to ```NuleculeB``` and ```NuleculeC```.

```
     NuleculeA
        / \
NuleculeB NuleculeC
```

```
---
specversion: 0.0.2
id: NuleculeA

metadata:
  name: NuleculeA
  appversion: 1.0
  description: App description

graph:
  - name: NuleculeComponent1
    artifacts:
      kubernetes:
        - file://artifacts/kubernetes/template.yaml

  - name: NuleculeComponent2
    artifacts:
      kubernetes:
        - file://artifacts/kubernetes/template.yaml

  - name: NuleculeB
    source: docker://NuleculeB

  - name: NuleculeC
    source: docker://NuleculeC
```

In this second use case, ```NuleculeA``` has a first-order relationship with ```NuleculeComponent1``` and ```NuleculeComponent2``` and a second-order relationship with ```NuleculeB``` and ```NuleculeC```, since either or both of ```NuleculeComponent1``` and ```NuleculeComponent2``` (i.e. ```internals```) are what actually depend on ```NuleculeB``` or ```NuleculeC```.

Given the inability to map which of ```NuleculeComponent1``` or ```NuleculeComponent2``` actually depends on ```NuleculeB``` or ```NuleculeC``` with any precision, more sweeping assumptions must be made about their relationships. That is why I was originally allured by the assumption that there was a one-to-one relationship between ```Nulecule``` and ```NuleculeComponent```, as opposed to one-to-many, since some granularity is lost in the latter case.

```
              NuleculeA
                 / \
 NuleculeComponent1 NuleculeComponent2
        / \                 / \
NuleculeB NuleculeC NuleculeB NuleculeC
```

**Comment by rtnpro** [comment]
Apart from the failing tests, LGTM :+1:

File: atomicapp/nulecule/main.py (hunk 4 above)
Comment (rtnpro) [inline atomicapp/nulecule/main.py]: On Wed, Jun 29, 2016 at 8:28 PM, Khris Richardson
<notifications@github.com> wrote:
> In atomicapp/nulecule/main.py:
>
>> +
>> +        digraph.add_node(progenitor)
>> +
>> +        for internal in internals:
>> +            digraph.add_edge(progenitor, internal)
>> +
>> +        for external in externals:
>> +            app_spec = Utils.getSourceImage(external)
>> +            name = external[NAME_KEY]
>> +            destination = "%s/%s/%s" % (self.app_path, EXTERNAL_APP_DIR,
>> name)
>> +            if app_spec is not None:
>> +                kwargs.update({'namespace': name})
>> +                kwargs.update(self._process_app_spec(app_spec,
>> destination))
>> +                subgraph = self.digraph(**kwargs)
>> +                if subgraph.order() > 0:
>> +                    patronage = internals or [progenitor]
>
> Of all the lines of code, you found the one that I thought was most worthy
> of some commentary, yet hardest to explain. Essentially what's going on here
> is as follows, which probably makes it worthwhile to separate things back
> into two classes, assuming they don't inherit from the same base class.

:+1: for having two classes for Nulecule and NuleculeComponent and not
inheriting from a common base class.

> progenitor and externals each have the progenitor attribute set to true and
> align with what was previously the Nulecule class, whereas internals
> describe the actual units of deployment and align with what was previously
> the NuleculeComponent class. It seems reasonable to restore that basic class
> hierarchy.
>
> Take for example two use cases.
>
> ---
> specversion: 0.0.2
> id: NuleculeA
>
> metadata:
>   name: NuleculeA
>   appversion: 1.0
>   description: App description
>
> graph:
>   - name: NuleculeB
>     source: docker://NuleculeB
>
>   - name: NuleculeC
>     source: docker://NuleculeC
>
> I'll revert to using the prior lingo to explain what's going on. In this
> first use case, we have a Nulecule with relationships only to other Nulecule
> type objects, whereas in the latter case the Nulecule has a relationship to
> both Nulecule and NuleculeComponent type objects.
>
> In the first example, since there are no NuleculeComponent type objects, we
> need to create an edge from NuleculeA (i.e. or [progenitor]) to NuleculeB
> and NuleculeC.
>
>      NuleculeA
>         / \
> NuleculeB NuleculeC
>
> ---
> specversion: 0.0.2
> id: NuleculeA
>
> metadata:
>   name: NuleculeA
>   appversion: 1.0
>   description: App description
>
> graph:
>   - name: NuleculeComponent1
>     artifacts:
>       kubernetes:
>         - file://artifacts/kubernetes/template.yaml
>
>   - name: NuleculeComponent2
>     artifacts:
>       kubernetes:
>         - file://artifacts/kubernetes/template.yaml
>
>   - name: NuleculeB
>     source: docker://NuleculeB
>
>   - name: NuleculeC
>     source: docker://NuleculeC
>
> In this second use case, NuleculeA has a first-order relationship with
> NuleculeComponent1 and NuleculeComponent2 and a second-order relationship
> with NuleculeB and NuleculeC, since either or both of NuleculeComponent1 and
> NuleculeComponent2 (i.e. internals) are what actually depend on NuleculeB or
> NuleculeC.

This makes things clear :)

> Given the inability to map which of NuleculeComponent1 or NuleculeComponent2
> actually depends on NuleculeB or NuleculeC with any precision, more sweeping
> assumptions must be made about their relationships. That is why I was
> originally allured by the assumption that there was a one-to-one
> relationship between Nulecule and NuleculeComponent, as opposed to
> one-to-many, since some granularity is lost in the latter case.
>
>               NuleculeA
>                  / \
>  NuleculeComponent1 NuleculeComponent2
>         / \                 / \
> NuleculeB NuleculeC NuleculeB NuleculeC

This is a limitation of the spec. I think we should have a discussion
about this. It'd be great if you could open up an issue for this
limitation of Nulecule, and we can discuss if it's worthwhile to
include this feature in the Nulecule spec itself.

**Comment by dustymabe** [comment]
hey @khrisrichardson 

With regards to your commit "epel version of networkx lacks reverse topo sort" it would be really great if we can still pull from RPMs so that our downstream counterpart won't have issues when integrated the new work. I am sending an email to the maintainers of python-networkx and hopefully we'll get some good news back from them on updating the rpm in EPEL.

**Comment by khrisrichardson** [comment]
Thanks @dustymabe. I assumed that commit might be somewhat controversial.

In the meantime, things have been busy on my end, so I haven't been able to complete refactoring the PR to use Nulecule and NuleculeComponent classes once more, as opposed to the progenitor attribute. What I have done has made the mapping back to the spec much easier to reason about.

However, the refactoring has made it apparent that in order for the NuleculeComponents to inherit configs from the multiple Nulecules that may have them as graph members, a rethinking of the graph data structure may be in order.

I'm thinking that the graph's nodes should just be comprised of NuleculeComponents and a tuple of the Nulecules. For example...

```
                            (NuleculeComponentA, (Nulecule1, Nulecule2))
                                                 /\
     (NuleculeComponentB, (Nulecule1, Nulecule3))  (NuleculeComponentC (Nulecule2, Nulecule3))
```

Kind of a nightmare to reason about, so trying to see where there's room to simplify here.

dustymabe:
> Kind of a nightmare to reason about, so trying to see where there's room to simplify here.

well we certainly appreciate that! any refactoring that helps us conceptually break up this puzzle into smaller pieces would be very welcome. i know I certainly scratch my head quite a bit when I start to look at the code after being away for some time.

PR #747, opened 2016-06-21
more inclusive rules for adding 'es'

khrisrichardson:
Objects of kind 'Ingress' in Kubernetes were being rejected because of improper pluralization, and changing the pluralization rules is one manner to address that.

https://www.grammarly.com/handbook/mechanics/spelling/5/spelling-plurals-with-s-es-or-other-plurals/

### DIFF
--- a/atomicapp/providers/lib/kubeshift/kubebase.py
+++ b/atomicapp/providers/lib/kubeshift/kubebase.py
@@ -224,7 +224,7 @@ class KubeBase(object):
             Resource name (str) (kind in plural form)
         """
         singular = kind.lower()
-        if singular.endswith("status"):
+        if singular.endswith(("s", "x", "z", "ch", "sh")):
             plural = singular + "es"
         else:
             if singular[-1] == "s":


cdrage:
@kadel does this make sense for the openshift provider too?

kadel:
@cdrage Yes it this makes sense for both OpenShift and Kubernetes.

I looked at Kubernetes source and they also updated this logic.


:+1: LGTM

cdrage:
#dotests

cdrage:
Source for Go: https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/kubectl.go#L190

Anyways, running tests. But code does LGTM :+1: So once tests pass I'll merge.

kadel:
hmm, there is also this https://github.com/kubernetes/kubernetes/blob/master/pkg/api/meta/restmapper.go#L161

cdrage:
k8s tests pass :+1: so good for merging after @khrisrichardson looks at https://github.com/kubernetes/kubernetes/blob/master/pkg/api/meta/restmapper.go#L161 :)

khrisrichardson:
@kadel, is the suggestion to drop the resource existence checks from the client, and allow the k8s api to test for validity of the resource name, since it handles the plurality as well? If so I am in favor.

kadel:
@khrisrichardson I'm sorry, I'm not sure if I follow :disappointed: 

By the resource existence check you mean this [kubernetes.py#L165] (https://github.com/projectatomic/atomicapp/blob/f4b96d4d822278b4a5ad2821faf67fd760564dd4/atomicapp/providers/lib/kubeshift/kubernetes.py#L165)?

I guess we can get rid of the resource existence check, because k8s api is going to return 404 anyway.
But we still need to construct correct url. For that we need to know the resource name, so we still need to do kind to resource conversion (make plural form)

khrisrichardson:
@kadel I wasn't quite sure where the REST mapper came into play, and am still not certain after some further poking and prodding. I lazily assumed you were suggesting that making a request to a url with the singular form would be converted to the plural on the server.

RE: existence check, yes, you interpreted that correctly. It seems like a lot of round-trips to determine what kinds are valid, when a 404 is just as damning.

Looks like something is needed until [#18622](https://github.com/kubernetes/kubernetes/issues/18622) to do the kind to resource conversion. Can you please clarify what you are suggesting with the REST mapper remark, since my interpretation was clearly off the mark?

kadel:
@khrisrichardson with restmapper remark I just wanted to point out how it is implemented in k8s. 
 But @cdrage found slightly  different implementation in kubectl client from what is in restmapper, so it looks like even between different parts of kubernetes it is not consistent. 

So I think this is ok, lets merge this.

cdrage:
@kadel You're right. Let's merge this for now.

PR #750, opened 2016-06-28
Add "index" command to Atomic App

cdrage:
This adds an index command to list multiple Nulecule's from an external
library that contains an index.yaml file in it's root directory.

Built upon @vpavlin 's work :)

TODO:
- [x] `atomicapp index list --update` command
- [x] "official" index docker image (will need to add this to projectatomic/nulecule-library)
- [x] script to generate index.yaml
- [x] integration tests
- [x] docs going to update via https://github.com/projectatomic/atomicapp/issues/768

### DIFF
--- a/atomicapp/cli/main.py
+++ b/atomicapp/cli/main.py
@@ -37,6 +37,7 @@ from atomicapp.nulecule import NuleculeManager
 from atomicapp.nulecule.exceptions import NuleculeException, DockerException
 from atomicapp.plugin import ProviderFailedException
 from atomicapp.utils import Utils
+from atomicapp.index import Index
 
 logger = logging.getLogger(LOGGER_DEFAULT)
 
@@ -110,6 +111,18 @@ def cli_init(args):
     except Exception as e:
         logger.error(e, exc_info=True)
         sys.exit(1)
+
+
+def cli_index(args):
+    argdict = args.__dict__
+    i = Index()
+    if argdict["index_action"] == "list":
+        i.list()
+    elif argdict["index_action"] == "update":
+        i.update()
+    elif argdict["index_action"] == "generate":
+        i.generate(argdict["location"])
+    sys.exit(0)
 
 
 # Create a custom action parser. Need this because for some args we don't
@@ -381,6 +394,25 @@ class CLI():
             help='The name of a container image containing an Atomic App.')
         gena_subparser.set_defaults(func=cli_genanswers)
 
+        # === "index" SUBPARSER ===
+        index_subparser = toplevel_subparsers.add_parser(
+            "index", parents=[globals_parser])
+        index_action = index_subparser.add_subparsers(dest="index_action")
+
+        index_list = index_action.add_parser("list")
+        index_list.set_defaults(func=cli_index)
+
+        index_update = index_action.add_parser("update")
+        index_update.set_defaults(func=cli_index)
+
+        index_generate = index_action.add_parser("generate")
+        index_generate.add_argument(
+            "location",
+            help=(
+                "Path containing Nulecule applications "
+                "which will be part of the generated index"))
+        index_generate.set_defaults(func=cli_index)
+
         # === "init" SUBPARSER ===
         init_subparser = toplevel_subparsers.add_parser(
             "init", parents=[globals_parser])
@@ -466,7 +498,7 @@ class CLI():
         # a directory if they want to for "run". For that reason we won't
         # default the RUN label for Atomic App to provide an app_spec argument.
         # In this case pick up app_spec from $IMAGE env var (set by RUN label).
-        if args.action != 'init' and args.app_spec is None:
+        if args.action != 'init' and args.action != 'index' and args.app_spec is None:
             if os.environ.get('IMAGE') is not None:
                 logger.debug("Setting app_spec based on $IMAGE env var")
                 args.app_spec = os.environ['IMAGE']
--- a/atomicapp/constants.py
+++ b/atomicapp/constants.py
@@ -82,3 +82,10 @@ PERSISTENT_STORAGE_FORMAT = ["ReadWriteOnce", "ReadOnlyMany", "ReadWriteMany"]
 
 # If running in an openshift POD via `oc new-app`, the ca file is here
 OPENSHIFT_POD_CA_FILE = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+
+# Index
+INDEX_IMAGE = "projectatomic/nulecule-library"
+INDEX_DEFAULT_IMAGE_LOCATION = "localhost"
+INDEX_NAME = "index.yaml"
+INDEX_LOCATION = ".atomicapp/" + INDEX_NAME
+INDEX_GEN_DEFAULT_OUTPUT_LOC = "./" + INDEX_NAME
--- a/atomicapp/index.py
+++ b/atomicapp/index.py
@@ -1,0 +1,203 @@
+"""
+ Copyright 2014-2016 Red Hat, Inc.
+
+ This file is part of Atomic App.
+
+ Atomic App is free software: you can redistribute it and/or modify
+ it under the terms of the GNU Lesser General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ Atomic App is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU Lesser General Public License for more details.
+
+ You should have received a copy of the GNU Lesser General Public License
+ along with Atomic App. If not, see <http://www.gnu.org/licenses/>.
+"""
+
+from __future__ import print_function
+import os
+
+import logging
+import errno
+from constants import (INDEX_IMAGE,
+                       INDEX_LOCATION,
+                       INDEX_DEFAULT_IMAGE_LOCATION,
+                       INDEX_GEN_DEFAULT_OUTPUT_LOC,
+                       INDEX_NAME)
+from nulecule.container import DockerHandler
+from nulecule.base import Nulecule
+from atomicapp.nulecule.exceptions import NuleculeException
+
+from copy import deepcopy
+
+import anymarkup
+from atomicapp.utils import Utils
+
+logger = logging.getLogger(__name__)
+
+
+class IndexException(Exception):
+    pass
+
+
+class Index(object):
+
+    """
+    This class represents the 'index' command for Atomic App. This lists
+    all available packaged applications to use.
+    """
+
+    index_template = {"location": ".", "nulecules": []}
+
+    def __init__(self):
+
+        self.index = deepcopy(self.index_template)
+        self.index_location = os.path.join(Utils.getUserHome(), INDEX_LOCATION)
+        self._load_index_file(self.index_location)
+
+    def list(self):
+        """
+        This command lists all available Nulecule packaged applications in a
+        properly formatted way.
+        """
+
+        # In order to "format" it correctly, find the largest length of 'name', 'id', and 'appversion'
+        # Set a minimum length of '7' due to the length of each column name
+        id_length = 7
+        app_length = 7
+        location_length = 7
+
+        # Loop through each 'nulecule' and retrieve the largest string length
+        for entry in self.index["nulecules"]:
+            id = entry.get('id') or ""
+            version = entry['metadata'].get('appversion') or ""
+            location = entry['metadata'].get('location') or INDEX_DEFAULT_IMAGE_LOCATION
+
+            if len(id) > id_length:
+                id_length = len(id)
+            if len(version) > app_length:
+                app_length = len(version)
+            if len(location) > location_length:
+                location_length = len(location)
+
+        # Print out the "index bar" with the lengths
+        index_format = ("{0:%s}  {1:%s}  {2:10} {3:%s}" % (id_length, app_length, location_length))
+        print(index_format.format("ID", "VER", "PROVIDERS", "LOCATION"))
+
+        # Loop through each entry of the index and spit out the formatted line
+        for entry in self.index["nulecules"]:
+            # Get the list of providers (first letter)
+            providers = ""
+            for provider in entry["providers"]:
+                providers = "%s,%s" % (providers, provider[0].capitalize())
+
+            # Remove the first element, add brackets
+            providers = "{%s}" % providers[1:]
+
+            # Retrieve the entry information
+            id = entry.get('id') or ""
+            version = entry['metadata'].get('appversion') or ""
+            location = entry['metadata'].get('location') or INDEX_DEFAULT_IMAGE_LOCATION
+
+            # Print out the row
+            print(index_format.format(
+                id,
+                version,
+                providers,
+                location))
+
+    def update(self, index_image=INDEX_IMAGE):
+        """
+        Fetch the latest index image and update the file based upon
+        the INDEX_IMAGE attribute. By default, this should pull the
+        'official' Nulecule index.
+        """
+
+        logger.info("Updating the index list")
+        logger.info("Pulling latest index image...")
+        self._fetch_index_container()
+        logger.info("Index updated")
+
+    # TODO: Error out if the locaiton does not have a Nulecule file / dir
+    def generate(self, location, output_location=INDEX_GEN_DEFAULT_OUTPUT_LOC):
+        """
+        Generate an index.yaml with a provided directory location
+        """
+        logger.info("Generating index.yaml from %s" % location)
+        self.index = deepcopy(self.index_template)
+
+        if not os.path.isdir(location):
+            raise Exception("Location must be a directory")
+
+        for f in os.listdir(location):
+            nulecule_dir = os.path.join(location, f)
+            if f.startswith("."):
+                continue
+            if os.path.isdir(nulecule_dir):
+                try:
+                    index_info = self._nulecule_get_info(nulecule_dir)
+                except NuleculeException as e:
+                    logger.warning("SKIPPING %s. %s" %
+                                   (nulecule_dir, e))
+                    continue
+                index_info["path"] = f
+                self.index["nulecules"].append(index_info)
+
+        if len(index_info) > 0:
+            anymarkup.serialize_file(self.index, output_location, format="yaml")
+        logger.info("index.yaml generated")
+
+    def _fetch_index_container(self, index_image=INDEX_IMAGE):
+        """
+        Fetch the index container
+        """
+        # Create the ".atomicapp" dir if it does not exist
+        if not os.path.exists(os.path.dirname(self.index_location)):
+            try:
+                os.makedirs(os.path.dirname(self.index_location))
+            except OSError as exc:  # Guard against race condition
+                if exc.errno != errno.EEXIST:
+                    raise
+
+        dh = DockerHandler()
+        dh.pull(index_image)
+        dh.extract_files(index_image, "/" + INDEX_NAME, self.index_location)
+
+    def _load_index_file(self, index_file=INDEX_LOCATION):
+        """
+        Load the index file. If it does not exist, fetch it.
+        """
+        # If the file/path does not exist, retrieve the index yaml
+        if not os.path.exists(index_file):
+            logger.warning("Couldn't load index file: %s", index_file)
+            logger.info("Retrieving index...")
+            self._fetch_index_container()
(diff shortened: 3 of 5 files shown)


cdrage:
Here's what it looks like right now if you pull the PR and run `atomicapp index list`:

```sh
dropbox/dev/atomicapp  add-index ✔                                                                                                                                                                                                                       2m  ⍉
▶ atomicapp index list
ID                 VER      PROVIDERS  LOCATION                                
redis-atomicapp    0.0.1    {D,K}      docker://projectatomic/redis-atomicapp  
mariadb-atomicapp  1.0.0    {D,O,K}    docker://projectatomic/mariadb-atomicapp
helloapache-app    0.0.1    {D,K}      docker://projectatomic/helloapache   
```

cdrage:
https://github.com/projectatomic/nulecule-library/pull/96 location metadata has been added for updating the information (including docker.io urls)

cdrage:
Hey guys.

Pretty much done at this point in terms of Atomic App code.

Here are the updates

- Not going to include `atomicapp index info` for now until I research more on how the output format should look like, also whether or not we should include `atomicapp index edit`
- Started work on tests, have two more to go
- Need to add docs not only on how to use the command, but also that there will now be an `.atomicapp/index` on your system
- Haven't tested with `atomic cli` functionality yet

In order to try out this PR, you must:
```
# with https://github.com/projectatomic/nulecule-library/pull/97 PR patched
cd nulecule-library
docker build -t projectatomic/nulecule-library .
# then
atomicapp index list # should auto-update (put the index.yaml into your .atomicapp/index.yaml
```

Hope you guys like it :)
@kadel @rtnpro @dustymabe @vpavlin

cdrage:
What I'm waiting for/block with is:

- https://github.com/projectatomic/nulecule-library/pull/97 to be merged
- hub.docker.com to be updated with projectatomic/nulecule-library docker image for people to pull
- code reviews

atomicapp/index.py
```diff
@@ -0,0 +1,204 @@
...
+
+    index_template = {"location": ".", "nulecules": []}
+
+    def __init__(self):
+
+        # This "hack" moves the 'cursor' in the terminal up one,
+        # eliminating the INFO :: Atomic App: X.X.X - Mode: Index
+        # logging information. If we passed --debug, ignore.
+        # if not logging.getLogger().getEffectiveLevel() == logging.DEBUG:
+        #     sys.stdout.write("\033[F")
```
dustymabe: remove?

atomicapp/index.py
```diff
@@ -0,0 +1,204 @@
...
+            if os.path.isdir(nulecule_dir):
+                index_info = self._nulecule_get_info(nulecule_dir)
+                index_info["path"] = f
+                self.index["nulecules"].append(index_info)
+
+        if len(index_info) > 0:
+            anymarkup.serialize_file(self.index, output_location, format="yaml")
+        logger.info("index.yaml generated")
+
+    def _fetch(self, index_image=INDEX_IMAGE):
```
dustymabe: would prefer for this funcname to be more specific `fetch_index_container` or something along those lines.

atomicapp/index.py
```diff
@@ -0,0 +1,204 @@
...
+        if not os.path.exists(os.path.dirname(self.index_location)):
+            try:
+                os.makedirs(os.path.dirname(self.index_location))
+            except OSError as exc:  # Guard against race condition
+                if exc.errno != errno.EEXIST:
+                    raise
+
+        dh = DockerHandler()
+        dh.pull(index_image)
+        dh.extract(index_image, "/index.yaml", self.index_location, file=True)
```
dustymabe: shouldn't we be using `INDEX_NAME` here?

atomicapp/constants.py
```diff
@@ -82,3 +82,10 @@
 
 # If running in an openshift POD via `oc new-app`, the ca file is here
 OPENSHIFT_POD_CA_FILE = "/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+
+# Index
+INDEX_IMAGE = "projectatomic/nulecule-library"  # temp until we have an official image on projectatomic docker hub
+INDEX_LOCATION = ".atomicapp/index.yaml"
+INDEX_NAME = "index.yaml"
+INDEX_DEFAULT_IMAGE_LOCATION = "localhost"
+INDEX_GEN_DEFAULT_OUTPUT_LOC = "./index.yaml"
```
dustymabe: many of these variables essentially contain the same information. should we somehow have INDEX_NAME incorporated in the other ones that currently have `index.yaml` hardcoded?

atomicapp/index.py
dustymabe: I don't see any place where _fetch() is called with an argument. should we not enable someone to specify a different image to use as an index if they prefer?

atomicapp/index.py
```diff
@@ -0,0 +1,204 @@
...
+        index_info = {}
+        nulecule = Nulecule.load_from_path(
+            nulecule_dir, nodeps=True)
+        index_info["id"] = nulecule.id
+        index_info["metadata"] = nulecule.metadata
+        index_info["specversion"] = nulecule.specversion
+
+        providers_set = set()
+        for component in nulecule.components:
+            if component.artifacts:
```
dustymabe: Should there ever be a case where no artifacts exist?

atomicapp/index.py
```diff
@@ -0,0 +1,204 @@
...
+                if len(providers_set) == 0:
+                    providers_set = set(component.artifacts.keys())
```
dustymabe: does this properly handle the "inherit" case for kube/openshift?

atomicapp/index.py
```diff
@@ -0,0 +1,204 @@
...
+        logger.info("Generating index.yaml from %s" % location)
+        self.index = deepcopy(self.index_template)
+        if not os.path.exists(location):
+            location = self._get_repo(location)
+
+        if not os.path.isdir(location):
+            raise Exception("Location must be a git repository")
+
+        for f in os.listdir(location):
+            nulecule_dir = os.path.join(location, f)
```
dustymabe: so every dir in the `location` is a nulecule_dir? we just assume this?

atomicapp/index.py
```diff
@@ -0,0 +1,204 @@
...
+            if f.startswith("."):
+                continue
+            if os.path.isdir(nulecule_dir):
+                index_info = self._nulecule_get_info(nulecule_dir)
```
dustymabe: what if nulecule_dir doesn't actually have a Nulecule in it?

atomicapp/nulecule/container.py
```diff
@@ -155,9 +155,12 @@ def extract(self, image, source, dest, update=False):
 
         # Copy files
         logger.debug('Copying nulecule data from %s to %s' % (src, dest))
-        Utils.copy_dir(src, dest, update)
-        logger.debug('Removing tmp dir: %s' % tmpdir)
-        Utils.rm_dir(tmpdir)
+        if file:
+            Utils.copy_file(src, dest)
+        else:
+            Utils.copy_dir(src, dest, update)
```
dustymabe: the two statements below need to be unindented. the tmpdir needs to be deleted regardless of if we were doing just a file copy or an entire dir copy.

atomicapp/nulecule/container.py
```diff
@@ -90,7 +90,7 @@ def pull(self, image, update=False):
 
         cockpit_logger.info('Skipping pulling docker image: %s' % image)
 
-    def extract(self, image, source, dest, update=False):
+    def extract(self, image, source, dest, update=False, file=False):
```
dustymabe: ok after some thought I think we need to refactor this function. unfortunately this isn't a generic function for copying out data from a container. It appears that way but we have all of the `App exists locally. Performing update...` stuff in the middle. This should be broken out into two functions. 

- One. This function just copies a file or directory out of a container to a target file or directory (the specified directory could be a tmpdir that will be later cleaned up by the calling function). We may even choose to put this function in a different file as it may belong somewhere else.
- Two. This function creates a tmp dir, calls the first function, does the `App exists locally. Performing update...` and then cleans up the tmp dir. 

Does that make sense?

requirements.txt
```diff
@@ -2,3 +2,4 @@ anymarkup
 jsonpointer
 requests
 websocket-client
+GitPython
```
dustymabe: I'm not a big fan of adding a new dependency for something that is probably going to get used 1% of the time. hmmm

dustymabe: at least I think it is only used for the "generate" portion of the code right? so that means it probably won't get used that much compared to the other functionality

atomicapp/index.py
cdrage: I'll add this as a CLI command. But yes, by default it'll pull the atomicapp one.

atomicapp/index.py
cdrage: Hopefully not. But I'll add this is a race condition.

atomicapp/index.py
cdrage: Yes. For example (the redis example):
```
redis-atomicapp           0.0.1    {D,O,K}    docker.io/projectatomic/redis-centos7-atomicapp
```

atomicapp/index.py
cdrage: Yes. We assume each directory has a Nulecule component for the generation command.

From your comment down below (in regards to the Git dependency) we can instead create a generation script inside of the projectatomic/nulecule-library repo instead and further improve it.

atomicapp/index.py
cdrage: Good question :)

atomicapp/nulecule/container.py
cdrage: That makes sense! I'll implement it.

requirements.txt
cdrage: I agree @dustymabe in regards to the extra dependency. This call will be rarely used. Only used if the person wants to create their own "nulecule library" hosted on a different docker container, etc.

Do you agree that we should instead take the `generation` command and add it to `projectatomic/nulecule-library` repo?

atomicapp/nulecule/container.py
cdrage: Bit confused on the second function, since we don't detect explicit whether something needs to be updated, we just pass in True/False to the `update` variable.

So the second function is just going to call the first and override the previous information?

dustymabe: can you fpaste or pastebin what you have right now (the two new functions) and I'll try to understand your confusion and then try to help come up with a solution

requirements.txt
dustymabe: Another alternative is to just only support generation from a local directory and  have the user run `git clone` before doing a "generation".

cdrage: Ah.... that too xD we can just say that we support paths only.

atomicapp/nulecule/container.py
dustymabe: @cdrage - here is a 5 minute try at this: http://paste.fedoraproject.org/389012/46799478

I would suggest we do this in a separate PR and then rebase this PR on top.

cdrage:
#dotests

cdrage:
#dotests

cdrage:
#dotests

cdrage:
#dotests

cdrage:
#dotests

cdrage:
#dotests

cdrage:
#dotests

cdrage:
@dustymabe other than the TODO I added (going to work on adding a test + failure if there is no `Nulecule` in the directory when using generate, this is good to go.

dustymabe:
@cdrage - i'll try to look at this tomorrow - also you still have one check box for docs

atomicapp/index.py
cdrage: This actually fails already:
```
ERROR  :: No Nulecule file exists in directory: ./gocounter-scratch-atomicapp
```

:)

cdrage:
Okay this has been update and ready for code review + possible merging.

@dustymabe @rtnpro @containscafeine @kadel if ya'll want to try it out / review it :)

cdrage:
I'll be updating the docs in a separate PR since we need to update https://github.com/projectatomic/atomicapp/blob/master/docs/cli.md anyways.
Opened up issue #768 for this.

cdrage:
#dotests

atomicapp/cli/main.py
```diff
@@ -381,6 +395,28 @@ def create_parser(self):
...
         gena_subparser.set_defaults(func=cli_genanswers)
 
+        # === "index" SUBPARSER ===
+        index_subparser = toplevel_subparsers.add_parser(
+            "index", parents=[globals_parser])
+        index_action = index_subparser.add_subparsers(dest="index_action")
+
+        index_list = index_action.add_parser("list")
+        index_list.add_argument(
+            "--update",
```
dustymabe: so what do you think about having this as a separate subcommand rather than an option to `--list`? seems like `atomicapp index update` would be more natural than `atomicapp index list --update`.

atomicapp/index.py
dustymabe: did you add the option to specify a different container name to the CLI? I don't see it.

atomicapp/cli/main.py
vpavlin: How about `atomicapp index pull` ? (It's a docker pull anyway:) )

atomicapp/index.py
dustymabe: so the problem with assuming it is a nulecule dir is: will you get an exception if it's not ? we need to fail gracefully. can you test and make sure?

atomicapp/index.py
dustymabe: ahh ok - so does that mean if there is one directory in the whole repo that doesn't have a nulecule then the whole process will fail?

atomicapp/utils.py
```diff
@@ -372,6 +373,11 @@ def copy_dir(src, dest, update=False, dryrun=False):
             distutils.dir_util.copy_tree(src, dest, update)
 
     @staticmethod
+    def copy_file(src, dest, dryrun=False):
```
dustymabe: did you forget to add a file to the commit? I don't see where copy_file gets used anywhere.

atomicapp/index.py
cdrage: @dustymabe yes. Should we "warn" instead?

atomicapp/cli/main.py
cdrage: That'd be good, what do you think @dustymabe ?

atomicapp/utils.py
cdrage: You're right. Updated this.

atomicapp/cli/main.py
dustymabe: ehh i like "update" but don't really care.. potato potato - go with whatever

atomicapp/index.py
dustymabe: yeah I would just warn (or maybe even INFO) and say no nulecule file within dir.. skipping

cdrage:
@dustymabe 

updated based on your comments. added the `atomicapp index update` command (i liked your suggestion better). removed the random `utils.py` stuff I still had there as well.

cdrage:
#dotests

atomicapp/cli/main.py
```diff
@@ -381,6 +394,25 @@ def create_parser(self):
...
+        index_list.set_defaults(func=cli_index)
+
+        index_update = index_action.add_parser("update")
+        index_update.set_defaults(func=cli_index)
+
+        index_generate = index_action.add_parser("generate")
+        index_generate.add_argument(
+            "location",
+            help=(
+                "Path or Git repository link containing Nulecule applications "
```
dustymabe: 'or Git repository' - is that still true?

atomicapp/index.py
```diff
@@ -0,0 +1,203 @@
...
+
+        for f in os.listdir(location):
+            nulecule_dir = os.path.join(location, f)
+            if f.startswith("."):
+                continue
+            if os.path.isdir(nulecule_dir):
+                try:
+                    index_info = self._nulecule_get_info(nulecule_dir)
+                except NuleculeException as e:
+                    logger.warning("SKIPPING %s. %s" %
```
dustymabe: just a little more context would be nice.. "No Nulecule found in /path/to/dir, Skipping.." 

Doesn't have to be that, just a suggestion.

atomicapp/index.py
```diff
@@ -0,0 +1,203 @@
...
+        if not os.path.exists(os.path.dirname(self.index_location)):
+            try:
+                os.makedirs(os.path.dirname(self.index_location))
+            except OSError as exc:  # Guard against race condition
+                if exc.errno != errno.EEXIST:
+                    raise
+
+        dh = DockerHandler()
+        dh.pull(index_image)
+        dh.extract_files(index_image, "/" + INDEX_NAME, self.index_location)
```
dustymabe: hmm ok - so running this in the "in container" use case doesn't work. It doesn't copy files to ~/.atomicapp/ - it copies them to that dir in the container but not outside the container.

one thing to consider is do we need to store things on the machine or could we just leave the index in the container and cp it out each time the user runs the command.

atomicapp/index.py
cdrage: I left it out since the error out from `base.py` does that already: 
```
raise NuleculeException("No Nulecule file exists in directory: %s" % src)
```

dustymabe: ok

atomicapp/index.py
cdrage: @dustymabe 
Could we use the `utils.py` function `getUserHome()` to set the file location appropriately?

dustymabe: yes. i think that is what that function is for, but we should also make sure the file isn't owned by root after we get done creating it. right?

cdrage:
Hey @dustymabe

So the index_location has been updated in the `init` of the `index.py` file to the proper location via:
```
self.index_location = os.path.join(Utils.getUserHome(), INDEX_LOCATION)
```

```
Utils.setFileOwnerGroup(dest)
```
has also been added to the `extract_files` function so it properly sets the permission when pulling.

One last review? :)

cdrage:
```
dropbox/dev/atomicapp  add-index ✔                                                                                                                                                                                                                        1m  
▶ atomicapp index list
INFO   :: Atomic App: 0.6.1 - Mode: Index
ID                        VER      PROVIDERS  LOCATION                                             
postgresql-atomicapp      1.0.0    {D,O,K}    docker.io/projectatomic/postgresql-centos7-atomicapp 
flask_redis_nulecule      0.0.1    {D,K}      docker.io/projectatomic/flask-redis-centos7-atomicapp
redis-atomicapp           0.0.1    {D,O,K}    docker.io/projectatomic/redis-centos7-atomicapp      
gocounter                 0.0.1    {D,K}      docker.io/projectatomic/gocounter-scratch-atomicapp  
mariadb-atomicapp         1.0.0    {D,O,K}    docker.io/projectatomic/mariadb-centos7-atomicapp    
helloapache-app           0.0.1    {D,K,M}    docker.io/projectatomic/helloapache                  
mongodb-atomicapp         1.0.0    {D,O,K}    docker.io/projectatomic/mongodb-centos7-atomicapp    
etherpad-app              0.0.1    {D,O,K}    docker.io/projectatomic/etherpad-centos7-atomicapp   
apache-centos7-atomicapp  0.0.1    {D,K,M}    docker.io/projectatomic/apache-centos7-atomicapp     
wordpress-atomicapp       2.0.0    {D,O,K}    docker.io/projectatomic/wordpress-centos7-atomicapp  
skydns-atomicapp          0.0.1    {K}        docker.io/projectatomic/skydns-atomicapp             
guestbookgo-atomicapp     0.0.1    {O,K}      docker.io/projectatomic/guestbookgo-atomicapp        
mariadb-app               0.0.1    {D,K}      docker.io/projectatomic/mariadb-fedora-atomicapp     
gitlab-atomicapp          1.2.0    {D,K}      docker.io/projectatomic/gitlab-centos7-atomicapp     

dropbox/dev/atomicapp  add-index ✔                                                                                                                                                                                                                        1m  
▶ atomicapp index update
INFO   :: Atomic App: 0.6.1 - Mode: Index
INFO   :: Updating the index list
INFO   :: Pulling latest index image...
INFO   :: Skipping pulling docker image: projectatomic/nulecule-library
INFO   :: Copying files from image projectatomic/nulecule-library:/index.yaml to /home/wikus/.atomicapp/index.yaml
INFO   :: Index updated
```

And the permissions:
```
▶ tree ~/.atomicapp -ug 
/home/wikus/.atomicapp
└── [wikus    wikus   ]  index.yaml

0 directories, 1 file
```

dustymabe:
LGTM

cdrage:
#dotests

PR #751, opened 2016-06-28
use check_call instead of check_command

khrisrichardson:
To check whether the return code is 0, the check_call method should be used instead of check_command.

### DIFF
--- a/atomicapp/nulecule/container.py
+++ b/atomicapp/nulecule/container.py
@@ -85,7 +85,7 @@ class DockerHandler(object):
 
         if self.dryrun:
             logger.info("DRY-RUN: %s", pull_cmd)
-        elif subprocess.check_output(pull_cmd) != 0:
+        elif subprocess.check_call(pull_cmd) != 0:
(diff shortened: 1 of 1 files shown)


atomicapp/nulecule/container.py
```diff
@@ -85,7 +85,7 @@ def pull(self, image, update=False):
 
         if self.dryrun:
             logger.info("DRY-RUN: %s", pull_cmd)
-        elif subprocess.check_output(pull_cmd) != 0:
+        elif subprocess.check_call(pull_cmd) != 0:
```
rtnpro: I think neither the older or new version of this line are going to work. Both ``check_call`` and ``check_output`` raises ``CalledProcessError`` exception when a non-zero status is returned. So, I think we should handle the ``CalledProcessError`` exception and raise ``DockerException`` rather than inside the ``elif`` block.

cdrage:
Hey @rtnpro @khrisrichardson 

Please have a look at PR #753 

I've gone ahead and fixed the issue. Since for some reason it occured with #752

PR #757, opened 2016-07-08
Remove cockpit

cdrage:
Due to the deprecation of Nulecule in the Cockpit project as part of the
refactor of the Kubernetes functionality within Cockpit, we will be
removing cockpit logging from Atomic App until the Cockpit project
decides to re-integrate it into the project.

ping @sub-mod @dustymabe

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ install:
 test:
 	pip install -qr requirements.txt
 	pip install -qr test-requirements.txt
-	$(PYTHON) -m pytest -vv --cov atomicapp
+	$(PYTHON) -m pytest tests/ -vv --cov atomicapp
 
 .PHONY: image
 image:


cdrage:
#dotests

cdrage:
ping @kadel @rtnpro @dustymabe your thoughts?

dustymabe:
ehh. I guess this is ok. would be great to know if we are going to need to add it back. I guess we'll always have the code (thanks git) so it shouldn't be too hard to add back in the future. 

I'm +1 unless @sub-mod has any thoughts about the future.

dustymabe:
close?

cdrage:
@dustymabe looks like I deleted this branch by accident git-side, i'll re-open it.

dustymabe:
re-open?

PR #767, opened 2016-07-12
Fix coverage

cdrage:
This fixes the coverage issues (adds --cov to test cli command)


cdrage:
#dotests

cdrage:
#dotests

cdrage:
Yay, there we go!

Issue #745, opened 2016-06-20
When fetching via atomicapp, should set to current user uid/guid


cdrage:
Possibly use: http://stackoverflow.com/questions/5994840/how-to-change-the-user-and-group-permissions-for-a-directory-by-name to achieve this.

But at the moment, whenever you fetch an `atomicapp`. It is pulled via sudo and thus all files are chown'd to root.

Should we implement a function that sets the guid/uid of the file pulled to the current user?

abhi1004:
@cdrage Can I work on this, looked into it think I can fix it.

cdrage:
@abhi1004 sorry! funny enough I was *just* putting up a PR for this, should be up in the next hour :)

abhi1004:
Haha okay .

cdrage:
Closed via #772

PR #772, opened 2016-07-18
When fetching or extracting, set the correct uid + guid

cdrage:
Issue: Whenever we fetch a container and extract it to a location, it will set
the permission as "root".

Solution: By checking if we are in a container as well as running via
SUDO we are able to determine what user we are running as. This commit
sets the files which are extracting to the corresponding user permission that
ran the initial atomicapp whether through the atomicapp CLI or atomic CLI.

### DIFF
--- a/atomicapp/nulecule/container.py
+++ b/atomicapp/nulecule/container.py
@@ -182,6 +182,9 @@ class DockerHandler(object):
         logger.debug('Removing tmp dir: %s' % tmpdir)
         Utils.rm_dir(tmpdir)
 
+        # Set the proper permissions on the extracted folder
+        Utils.setFileOwnerGroup(dest)
+
     def is_image_present(self, image):
         """
         Check if a Docker image is present in the host.
--- a/atomicapp/utils.py
+++ b/atomicapp/utils.py
@@ -20,6 +20,7 @@
 from __future__ import print_function
 import distutils.dir_util
 import os
+import pwd
 import sys
 import tempfile
 import re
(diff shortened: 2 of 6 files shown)


cdrage:
Closes #745

cdrage:
#dotests

atomicapp/utils.py
```diff
@@ -377,6 +377,53 @@ def rm_dir(directory):
...
+
+        Returns:
+            (int): User UID
+        """
+
+        # If we're in a container we should be looking in the /host/ directory
+        if Utils.inContainer():
+            passwd = os.path.join(HOST_DIR, "etc/passwd")
+        else:
+            passwd = "/etc/passwd"
```
dustymabe: I think we aren't going to be able to use this. the user's mapping won't be in /etc/passwd if the user/grp information comes from ldap or some other identity management system. We'll have to use something like `getent passwd cdrage` to give us that information.

atomicapp/utils.py
```diff
@@ -377,6 +377,53 @@ def rm_dir(directory):
...
+
+        uidmap = {}
+        for line in file(passwd):
+            lst = line.split(':')
+            uidmap[lst[0]] = lst[2]
+
+        return int(uidmap[user])
+
+    @staticmethod
+    def setFilePermissions(src):
```
dustymabe: so this name is a bit of a misleading name. we are not actually changing any file perms, just changing who owns the file - so maybe we should call it setFileOwnerGroup()

atomicapp/utils.py
cdrage: The problem I found is getting the UID from within the container (hence why
I do it via /etc/passwd). How could we possibly do this from within the
container?

dustymabe: probably using `os.chroot()` the same way we do in `getUserHome()`

cdrage:
#dotests

dustymabe:
ok I think we are close. there might be a few more places that we need to add the code that does the owner fixups. here is what I get on my system:

```
[vagrant@f24 projectatomic-guestbookgo-atomicapp-204b4b9f525a]$ ls -lh ./
total 24K
-rw-r--r--. 1 vagrant vagrant  298 Jul 18 20:47 answers.conf.gen
drwxr-xr-x. 4 vagrant vagrant 4.0K Jul 18 20:47 artifacts
-rw-r--r--. 1 vagrant vagrant  313 Jul 18 19:03 Dockerfile
drwxr-xr-x. 3 root    root    4.0K Jul 18 20:47 external
-rw-r--r--. 1 vagrant vagrant  801 Jul 18 19:03 Nulecule
-rw-r--r--. 1 vagrant vagrant 2.1K Jul 18 19:03 README.md
```

so there is that one `root root` in there.

cdrage:
@dustymabe boo. i'll work on that

cdrage:
Updated, but fixing test error atm with `TestNuleculeComponentLoadExternalApplication.test_loading_app_by_unpacking`

atomicapp/nulecule/container.py
```diff
@@ -182,6 +182,10 @@ def extract_nulecule_data(self, image, source, dest, update=False):
         logger.debug('Removing tmp dir: %s' % tmpdir)
         Utils.rm_dir(tmpdir)
 
+        # Set the proper permissions on the extracted folder
+        logger.debug("Setting file ownership on %s" % dest)
```
dustymabe: would it be better to have this debug statement in the setFileOwnerGroup function?  Then we could have it say "Setting owner/group of %s to %s/%s"

atomicapp/utils.py
```diff
@@ -377,6 +378,50 @@ def rm_dir(directory):
...
+        """
+        This function sets the correct uid and gid bits to a source
+        file or directory given the current user that is running Atomic
+        App.
+        """
+        user = Utils.getUserName()
+
+        # Get the UID of the User
+        uid = Utils.getUid(user)
+        gid = uid
```
dustymabe: this is not a safe assumption to make. I think it would be safer if we actually did `pwd.getpwnam(user).pw_gid`. If `getUid()` is a new function then we could just change it to `getUidGid()` and have it return a tuple `(uid,gid)`.

dustymabe:
LGTM

dustymabe:
I found a corner case with the `getUserName()` function. I think we should still merge this PR. This case fails though:

If I am a normal user that has access to the docker daemon (i.e. it is unsecured like on the adb/cdk) I can execute docker run commands without sudo. If I execute `docker run` without sudo then my egid will be 0 but SUDO_USER env variable will not be set so we hit this case in the if statement: https://github.com/projectatomic/atomicapp/blob/b2bf5caec27e32298009f154253499e515a22de1/atomicapp/utils.py#L391

So we improperly get 'root' ownership on the files.

cdrage:
#dotests

cdrage:
@dustymabe added one simple test as well as updated the code. Good for mergin'

cdrage:
openshift failing = timing issue

cdrage:
#dotests

dustymabe:
Feel free to merge when you are happy.

PR #775, opened 2016-07-19
collect only atomicapp tests

containscafeine:
In case of external libraries, their tests are not run, but only the ones in tests directory are run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ install:
 test:
 	pip install -qr requirements.txt
 	pip install -qr test-requirements.txt
-	$(PYTHON) -m pytest -vv --cov atomicapp
+	$(PYTHON) -m pytest tests/ -vv --cov atomicapp
 
 .PHONY: image
 image:


cdrage:
add to whitelist

cdrage:
test this please

cdrage:
ok to test

cdrage:
#dotests

cdrage:
@containscafeine this looks okay to me, but how are we going to test docker-py? shouldn't this actually be added though as part of the test suite?

containscafeine:
@cdrage, I have written functional tests for shipy, and those require access to the network and certain docker images, and are very slow to run, I don't think it makes sense for atomicapp to run those. I run those locally, and shipy used here is always even with shipy upstream.

cdrage:
@containscafeine ah makes sense.

This LGTM then

cdrage:
@kadel @dustymabe @rtnpro does this LGTY (look good to you)? :)

Issue #768, opened 2016-07-14
Update CLI docs


cdrage:
This needs to be updated to correlate to our newest commands :)

https://github.com/projectatomic/atomicapp/blob/master/docs/cli.md

PR #781, opened 2016-07-26
Remove lifecycle for updated CLI doc

cdrage:
Removes the lifecycle doc and combined it into the CLI doc. This also adds a brief description on the `init` and `index` commands.

Closes https://github.com/projectatomic/atomicapp/issues/768


dustymabe:
LGTM

Issue #785, opened 2016-08-10
Genanswers does not work at the moment. Filename error.


cdrage:
Seems that `atomicapp genaswers .` doesn't work at the moment:

```
▶ atomicapp genanswers .
INFO   :: Atomic App: 0.6.2 - Mode: Genanswers
INFO   :: Pulling external application: mariadb-centos7-atomicapp
INFO   :: Unpacking image projectatomic/mariadb-centos7-atomicapp to /tmp/atomicappQXLrd0/external/mariadb-centos7-atomicapp
INFO   :: Skipping pulling docker image: projectatomic/mariadb-centos7-atomicapp
INFO   :: Extracting Nulecule data from image projectatomic/mariadb-centos7-atomicapp to /tmp/atomicappQXLrd0/external/mariadb-centos7-atomicapp
INFO   :: Copying files from image projectatomic/mariadb-centos7-atomicapp:application-entity to /tmp/nulecule-408e3324-5f0b-11e6-94ac-340286185b7d
ERROR  :: AnyMarkupError: Failed to guess markup format based on: filename
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.6.2-py2.7.egg/atomicapp/cli/main.py", line 143, in cli_func_exec
    cli_func(cli_func_args)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.6.2-py2.7.egg/atomicapp/cli/main.py", line 56, in cli_genanswers
    nm.genanswers(**argdict)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.6.2-py2.7.egg/atomicapp/nulecule/main.py", line 252, in genanswers
    self._write_answers(answers_file, answers, answers_format)
  File "/usr/local/lib/python2.7/dist-packages/atomicapp-0.6.2-py2.7.egg/atomicapp/nulecule/main.py", line 419, in _write_answers
    anymarkup.serialize_file(answers, path, format=answers_format)
  File "build/bdist.linux-x86_64/egg/anymarkup_core/__init__.py", line 197, in serialize_file
    return serialize(struct, format, f, encoding)
  File "build/bdist.linux-x86_64/egg/anymarkup_core/__init__.py", line 168, in serialize
    fmt = _get_format(format, fname)
  File "build/bdist.linux-x86_64/egg/anymarkup_core/__init__.py", line 420, in _get_format
    raise AnyMarkupError(err_string)
[...]

PR #786, opened 2016-08-10
Formatting error with anymarkup and genanswers

cdrage:
Seems that self.answers_format isn't passed in correctly to
_write_answers when using `atomicapp genanswers` correctly.

Resulting in an error where anymarkup believes the format is 'None'.
This commit explicitley passes self.answers_format which defaults to
'ini' if nothing is passed.

Closes https://github.com/projectatomic/atomicapp/issues/785


dustymabe:
LGTM

Issue #788, opened 2016-08-11
deploying to openshift gives message about kubernetes


dustymabe:
atomicapp at 0a3939d

I see this when deploying/undeploying to openshift:
```
$ myatomicapp run ./                                                                                                                                 
INFO   :: Atomic App: 0.6.2 - Mode: Run
INFO   :: Found existing external application: mariadb-centos7-atomicapp Loading: 
ANSWER => mariadb-centos7-atomicapp:mariadb-atomicapp | db_user (Database User): du
ANSWER => mariadb-centos7-atomicapp:mariadb-atomicapp | db_pass (Database Password): dp
ANSWER => mariadb-centos7-atomicapp:mariadb-atomicapp | db_name (Database Name): du
ANSWER => etherpad-app | db_user (Database User): du
ANSWER => etherpad-app | db_pass (Database Password): dp
ANSWER => etherpad-app | db_name (Database Name): du
INFO   :: Using namespace sample-project
WARN   :: CAUTION: TLS verification has been DISABLED
INFO   :: Deploying to Kubernetes
INFO   :: Pods 'mariadb' successfully created
INFO   :: Services 'mariadb' successfully created
INFO   :: Using namespace sample-project
WARN   :: CAUTION: TLS verification has been DISABLED
INFO   :: Deploying to Kubernetes
INFO   :: Services 'etherpad-svc' successfully created
INFO   :: Replicationcontrollers 'etherpad' successfully created

Your application resides in ./
Please use this directory for managing your application
```

```
$ myatomicapp stop ./                                                                                                                                
INFO   :: Atomic App: 0.6.2 - Mode: Stop
INFO   :: Found existing external application: mariadb-centos7-atomicapp Loading: 
INFO   :: Using namespace sample-project
WARN   :: CAUTION: TLS verification has been DISABLED
INFO   :: Undeploying from Kubernetes
INFO   :: Pods 'mariadb' successfully deleted
INFO   :: Services 'mariadb' successfully deleted
INFO   :: Using namespace sample-project
WARN   :: CAUTION: TLS verification has been DISABLED
INFO   :: Undeploying from Kubernetes
INFO   :: Services 'etherpad-svc' successfully deleted
INFO   :: 'etherpad' successfully scaled to 0
INFO   :: Replicationcontrollers 'etherpad' successfully deleted
```

The `INFO   :: Undeploying from Kubernetes` and `INFO   :: Deploying to Kubernetes` should say openshift.

PR #790, opened 2016-08-11
Fix typo in openshift.py provider

cdrage:
Should say "Deploying to OpenShift" rather than Kubernetes.

Same goes with "Undeploying from OpenShift"


cdrage:
Closes #788

dustymabe:
LGTM

PR #795, opened 2016-08-31
Use travis CI to run tests from tests/units dir.

rtnpro:
This is needed so that travis CI does not end up running functional tests from ``tests/`` dir.


cdrage:
centos unit tests are messing up for some reason, but this LGTM

PR #797, opened 2016-09-01
Fix CentOS CI Errors

cdrage:
Don't know *exactly* why it's failing, but let's try this. Could be
related to "foobar" actually going to an internal server.


cdrage:
#dotests

cdrage:
yay, looks like it fixed the tests, going to fix this "upstream" too in kubeshift.

dustymabe:
should we try to understand the situation here a bit more? would be nice to know the root cause.

cdrage:
No idea :( i can try pinging "foobar" from within the infrastructure and see what's up

PR #802, opened 2016-11-10
Add checking if resource is OpenShift specific to properly create OpenShift artifacts

dymurray:
It appears that this file was copy pasted from the Kubernetes Kubeshift file and wasn't tested. Currently deploying an atomicapp which has a route as an artifact will fail since a route isn't listed as a Kubernetes resource. I tested this deploying an Etherpad atomicapp (https://github.com/fusor/nulecule-library/tree/master/etherpad-atomicapp) on Openshift with an added route artifact with success.

┆Issue is synchronized with this [Asana task](https://app.asana.com/0/87361382444136/210898191031938)

### DIFF
--- a/atomicapp/cli/main.py
+++ b/atomicapp/cli/main.py
@@ -540,5 +540,6 @@ def main():
     cli = CLI()
     cli.run()
 
+
 if __name__ == '__main__':
     main()
--- a/atomicapp/providers/lib/kubeshift/openshift.py
+++ b/atomicapp/providers/lib/kubeshift/openshift.py
@@ -180,6 +180,8 @@ class KubeOpenshiftClient(object):
                 url = self.k8s_api
             else:
                 url = urljoin(self.k8s_apis, "%s/" % api_version)
+        elif resource in self.oc_api_resources:
+            url = self.oc_api
         else:
             raise KubeOpenshiftError("No kind by that name: %s" % kind)
 
--- a/tests/units/kubeshift/test_openshift.py
+++ b/tests/units/kubeshift/test_openshift.py
@@ -43,7 +43,7 @@ class FakeClient():
         pass
 
     def get_resources(self, *args):
-        return ['Pod', 'template']
+        return ['Pod', 'template', 'Route']
 
     def get_groups(self, *args):
         return {}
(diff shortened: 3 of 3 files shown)


cdrage:
#dotests

cdrage:
#dotests

atomicapp/cli/main.py
```diff
@@ -540,5 +540,6 @@ def main():
     cli = CLI()
     cli.run()
 
+
```
cdrage: random newline added?

atomicapp/providers/lib/kubeshift/openshift.py
```diff
@@ -180,6 +180,8 @@ def _generate_kurl(self, obj, namespace, name=None, params=None):
                 url = self.k8s_api
             else:
                 url = urljoin(self.k8s_apis, "%s/" % api_version)
+        elif resource in self.oc_api_resources:
+            url = self.oc_api
```
cdrage: This does LGTM, but I'm just going to make sure this *still* works when creating k8s resources via OpenShift.

tests/units/kubeshift/test_openshift.py
```diff
@@ -69,9 +69,29 @@ def test_create(mock_class):
...
+
+@mock.patch("atomicapp.providers.lib.kubeshift.openshift.KubeBase")
+def test_oc_delete(mock_class):
+    mock_class.return_value = FakeClient()
+    mock_class.kind_to_resource_name.return_value = 'Route'
+
+    oc_object = {"apiVersion": "v1", "kind": "Route", "metadata": {"labels": {"name": "helloapache-route"}, "name": "helloapache-route"}, "spec": {
+        "host": "$endpoint", "to": [{"kind": "Service", "name": "helloapache-svc"}]}}
+    a = KubeOpenshiftClient(config)
+    a.delete(oc_object, "foobar")
```
cdrage: :+1: for more tests. thank you!

atomicapp/cli/main.py
dymurray: @cdrage flake8 was updated last week and since the version is not locked down for CI testing it failed because the latest version requires a double newline between all def statements and the start of the program. Adding this newline passed the CI tests.

cdrage:
Seems that this works correctly on all fronts! :) I'm going to go ahead and merge this.

cdrage:
Thanks @dymurray

